PR #6, opened 2016-12-26
JProfiler support

ingokegel:
Requires the as of yet unreleased JProfiler version 9.2.2 because the gradle daemon limits VM parameters on the command line so that the JMX server port could not be set.

 To try it out, the current build can be downloaded at

 http://download.ej-technologies.com/jprofiler/jprofiler_macos_9_2_2.dmg

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
+classes
 /.gradle
 /.idea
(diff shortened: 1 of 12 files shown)


adammurdoch:
Thanks for the pull request.

PR #15, opened 2016-12-28
JProfiler support

ingokegel:
* Mark the heap before the build starts (available in JProfiler 10), so the heap dump can show new objects without the need for allocation recording
* Profile all packages that contain "gradle", profile "com.android.**"
* Added a new built-in config "sampling-all" without any call tree filters
* Added option --jprofiler-config-file for using an exported config file


adammurdoch:
Thanks for the pull request.

PR #16, opened 2017-01-03
Updated README to call out Gradle version requirement for chrome-trace

jmslau:
Specified Gradle 3.3+ is required to use chrome-trace as the profiler.


oehme:
Thanks @jmslau!

PR #39, opened 2017-06-15
No-daemon mode for JProfiler without MBean-polling

ingokegel:
Instead of polling whether the JProfiler MBean has been published,
provide all the profiling commands upfront by adding the appropriate
triggers to the config file with an XSL transform.

### DIFF
--- a/src/main/java/org/gradle/profiler/jprofiler/JProfilerController.java
+++ b/src/main/java/org/gradle/profiler/jprofiler/JProfilerController.java
@@ -28,25 +28,26 @@ public class JProfilerController implements ProfilerController {
 
     @Override
     public void start() throws IOException, InterruptedException {
-        if (!profileWholeLifeTime()) {
-            invoke("startCPURecording", true);
-            if (jProfilerConfig.isRecordAlloc()) {
-                invoke("startAllocRecording", true);
-            }
-            if (jProfilerConfig.isRecordMonitors()) {
-                invoke("startMonitorRecording");
-            }
-            for (String probeName : jProfilerConfig.getRecordedProbes()) {
-                boolean eventRecording = jProfilerConfig.getProbesWithEventRecording().contains(probeName);
-                boolean specialRecording = jProfilerConfig.getProbesWithSpecialRecording().contains(probeName);
-                invoke("startProbeRecording", probeName, eventRecording, specialRecording);
-            }
-            if (jProfilerConfig.isHeapDump() && hasOperation("markHeap")) { // available in JProfiler 10
-                invoke("markHeap");
-            }
-            if (profileWholeLifeTime()) {
-                invoke("saveSnapshotOnExit", getSnapshotPath());
-            }
+        if (profileWholeLifeTime()) {
+            return;
+        }
+        invoke("startCPURecording", true);
+        if (jProfilerConfig.isRecordAlloc()) {
+            invoke("startAllocRecording", true);
+        }
+        if (jProfilerConfig.isRecordMonitors()) {
+            invoke("startMonitorRecording");
+        }
+        for (String probeName : jProfilerConfig.getRecordedProbes()) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/gradle/profiler/jprofiler/JProfilerController.java
```diff
@@ -33,60 +28,25 @@ public JProfilerController(ScenarioSettings settings, JProfilerConfig jProfilerC
...
+            }
+            for (String probeName : jProfilerConfig.getRecordedProbes()) {
+                boolean eventRecording = jProfilerConfig.getProbesWithEventRecording().contains(probeName);
+                boolean specialRecording = jProfilerConfig.getProbesWithSpecialRecording().contains(probeName);
+                invoke("startProbeRecording", probeName, eventRecording, specialRecording);
+            }
+            if (jProfilerConfig.isHeapDump() && hasOperation("markHeap")) { // available in JProfiler 10
+                invoke("markHeap");
+            }
+            if (profileWholeLifeTime()) {
```
oehme: Unreachable statement

src/main/java/org/gradle/profiler/jprofiler/JProfilerController.java
```diff
@@ -33,60 +28,25 @@ public JProfilerController(ScenarioSettings settings, JProfilerConfig jProfilerC
...
-                startNow();
-                started = true;
-            } catch (Exception e) {
-                lastProblem = e;
-                try {
-                    Thread.sleep(10);
-                } catch (InterruptedException interrupt) {
-                    throw new RuntimeException(interrupt);
-                }
+        if (!profileWholeLifeTime()) {
```
oehme: Style: We prefer a quick return over a wrapping a long block inside the if.

src/main/java/org/gradle/profiler/jprofiler/JProfilerJvmArgsCalculator.java
```diff
@@ -67,19 +76,75 @@ private String getAgentPathParameter(int port) {
...
+        if (configFile != null) {
+            builder.append("id=").append(id).append(",config=").append(transformConfigFile(configFile, id));
         }
+
         builder.append(",sysprop=jprofiler.jmxServerPort=").append(port);
         return builder.toString();
     }
 
+    private File transformConfigFile(File configFile, String id) {
+        if (profileWholeLifeTime()) {
```
oehme: See stylistic comment above

src/main/java/org/gradle/profiler/jprofiler/JProfilerJvmArgsCalculator.java
```diff
@@ -67,19 +76,75 @@ private String getAgentPathParameter(int port) {
...
+        }
+        if (configFile != null) {
+            builder.append("id=").append(id).append(",config=").append(transformConfigFile(configFile, id));
         }
+
         builder.append(",sysprop=jprofiler.jmxServerPort=").append(port);
         return builder.toString();
     }
 
+    private File transformConfigFile(File configFile, String id) {
```
oehme: This code should be in its own class and have a unit test

ingokegel:
OK, done. Btw, can I update JProfiler.MAJOR_VERSION to 10 or would that break your test environment?

src/main/java/org/gradle/profiler/jprofiler/JProfilerJvmArgsCalculator.java
```diff
@@ -100,7 +120,7 @@ private File getConfigFile() {
                 try (InputStream inputStream = resource.openStream()) {
                     Files.copy(inputStream, tmpFile, StandardCopyOption.REPLACE_EXISTING);
                 }
-                return tmpFile.toFile();
+                return JProfiler.deleteOnExit(tmpFile.toFile());
```
oehme: This method doesn't seem like it belongs on `JProfiler`.

oehme:
Thanks a lot Ingo!

PR #40, opened 2017-07-06
Support for Linux perf profiling

DanielThomas:
@oehme as promised: `perf` support! Example flamegraph:

https://gist.github.com/DanielThomas/c6a4a152dd77e79144b206018c856f62

### DIFF
--- a/src/main/java/org/gradle/profiler/CommandExec.java
+++ b/src/main/java/org/gradle/profiler/CommandExec.java
@@ -1,6 +1,9 @@
 package org.gradle.profiler;
 
+import org.gradle.api.Nullable;
+
 import java.io.*;
+import java.lang.reflect.Field;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
@@ -39,7 +42,7 @@ public class CommandExec {
     public String runAndCollectOutput(String... commandLine) {
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         try {
-            run(new ProcessBuilder(commandLine), outputStream);
+            run(new ProcessBuilder(commandLine), outputStream, null, null).waitForSuccess();
         } catch (RuntimeException e) {
             System.out.print(new String(outputStream.toByteArray()));
             throw e;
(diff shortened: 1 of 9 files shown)


DanielThomas:
I've got one last commit coming to fix the package flamegraphs.

DanielThomas:
Looks like the current `perf script` output is different to the pkgsplit expectation. I'll circle back on that once I've talked to Brendan.

DanielThomas:
I did add icicle graphs though:

https://gist.github.com/DanielThomas/2295747d0c6ac72879dcd8bda6773fd4

DanielThomas:
Improved the flamegraph output by adding options for coloring and omitting smaller functions: https://gist.github.com/DanielThomas/11c4e5fd4090a206fe46563df542d2ab

src/main/java/org/gradle/profiler/CommandExec.java
```diff
@@ -39,57 +42,171 @@ public String runAndCollectOutput(List<String> commandLine) {
...
+            }
+            if (result != 0) {
+                throw new RuntimeException(commandErrorMessage(processBuilder) + ". Exited with result " + result);
+            }
+        }
+
+        public void interrupt() {
+            int pid;
+            try {
+                Field field = process.getClass().getDeclaredField("pid");
```
oehme: Will this work on windows?

melix:
> There's one change of which I'm not sure it will work on Windows. Apart from that, it would be great if @melix could give this a try from a user's perspective.

Yes it's on my TODO list, I have more reviews to complete before :)

src/main/java/org/gradle/profiler/CommandExec.java
DanielThomas: No, it's `UNIXProcess` specific. https://github.com/zeroturnaround/zt-process-killer looks to provide useful abstractions, perhaps should replace with that.

DanielThomas: Hmm, not very useful I'm afraid - it just abstracts `destroy` for `JavaProcess` and I couldn't get `destroy` to send a sigint and let `perf` flush it's write buffer safely.

DanielThomas:
Improvement made by 62bd609:

https://gist.github.com/DanielThomas/709cdfc313b97f7cb4f15eef48ecd78f

DanielThomas:
@melix I should mention - the flamegraphs are from profiling the profiler project itself w/ `clean installDist`.

src/main/java/org/gradle/profiler/CommandExec.java
oehme: We need the profiler to work on windows. Afaik this code is always called and new depends on unix. So we'll need to find a different way of doing this.

DanielThomas: `kill` is the same as always. `interrupt` is Unix-only.

oehme: Ah sorry, I missed that the call was only in the perf profiler.

oehme:
This is good to go from my side, as soon as Cédric has given it a spin.

DanielThomas:
Here's a concrete example for https://github.com/gradle/gradle/issues/1600:

https://gist.github.com/DanielThomas/3ea2ac2e745ca6c458422873f9cce625

melix:
@DanielThomas it works great! However I noticed that the preparation of support tools is done during iterations, not warmups. For stability of measurements, would it be possible to move the setup during warmups?

And I also noticed that the profiler is started/stopped for each iteration. Since starting and stopping has a significant cost (it takes several seconds on my laptop), would it make sense to start it before the first iteration and stop it after the last one completed? It would be consistent with what we do for the other profilers.

I will need some time to interpret the graphs, though. Examples I tested yet are hard to interpret, I see lots of "[perfxxx.map]" stacks, I wonder if it's expected...

DanielThomas:
@melix sure, we could move that somewhere else - I figured it didn't matter if it happened at profiler start time, given we're not measuring anything there.

I followed the pattern of the other controllers for start/stop; I assumed that would do the right thing for iterations etc. How would I achieve that if `ProfilerController` has no concept of iteration start/stop vs profiling session start/stop? 

FWIW, I'll come back and add support for both using the built-in stack folding in the later versions of perf and eBPF which will both significantly improve performance when we stop profiling.

The `[perfxxx.map]` frames are the Java frames. If you see nothing but these, then there's a problem gathering those on your machine, check the profiler log and see what went wrong (look for the `Fetching maps for all Java processes...` line). If you're seeing most of the Java frames, missing stack data means we saw a recompilation between the sample and the map file being output.

melix:
Indeed there's something fishy with start/stop, let me take a look. Thanks for the perf.map hint, I'll take a look.

melix:
I'm seeing this in the log:
```
Mapping PID 2078 (user root):
wc(1):   8177  24929 875968 /tmp/perf-2078.map

Mapping PID 2257 (user root):
wc(1):   164918   495570 27271818 /tmp/perf-2257.map

Mapping PID 4480 (user cchampeau):
wc(1):   61670  185879 8691660 /tmp/perf-4480.map

Mapping PID 6472 (user cchampeau):
wc(1):   15800   48104 2016391 /tmp/perf-6472.map

'trace' not valid for hardware events. Ignoring.
'trace' not valid for software events. Ignoring.
'trace' not valid for unknown events. Ignoring.
'trace' not valid for unknown events. Ignoring.
Failed to open /tmp/perf-3518.map, continuing without symbols
Failed to open /tmp/perf-5433.map, continuing without symbols
Failed to open /tmp/perf-6043.map, continuing without symbols
Failed to open /tmp/perf-5682.map, continuing without symbols
Failed to open /tmp/perf-5518.map, continuing without symbols
Failed to open /tmp/perf-5240.map, continuing without symbols
Failed to open /tmp/perf-5724.map, continuing without symbols
Failed to open /tmp/perf-6229.map, continuing without symbols
Failed to open /tmp/perf-5378.map, continuing without symbols
```
does it ring a bell?

oehme:
> Indeed there's something fishy with start/stop, let me take a look. 

It's just "start recording" and "stop recording". We have no concept of "start the profiler process"/"stop the profiler process", because no profiler so far needed this. E.g. JProfiler and YourKit just start up together with the JVM and die together with the JVM.

DanielThomas:
@melix that looks right, the first PID should be the profiler, the second the daemon and looks like they wrote out okay. Normal to see failures, if there were processes that went away by the time jmaps ran. If you still think the flamegraph doesn't look right, send me an example on Slack and I'll take a look.

melix:
@DanielThomas I'm still seeing strange things in profiling results. Here's an example:

selection_087

What I see is that the `perf` profiler is generating jmap files properly for the daemon process. However in the flame graph some stack trace elements seem to belong to a _different_ PID (the PID in this screenshot is **not** the PID of the daemon). I could also verify that the script is generating map files for every java process running, which would tend to indicate that either the java process was killed before the map file was created, or it's not a java process.

In any case I'm confused why a stack trace would contain elements that do _not_ belong to the daemon process.

DanielThomas:
@melix that's right, `perf` is not filtering by process, and includes everything. I then use `JavaProcessSanitizeFunction` to filter out the profiler process, but keep any Java processes in the results; the rationale being that a Gradle daemon spawns it's own workers, that you may want to profile. Limiting to one of those processes is just a matter of clicking the bottom-most PID frame, or searching.

melix:
Makes perfect sense. I kind of thought it could be the workers. Then it means it would be nice to find out the PIDs of the workers during the build so that we can map them too.

PR #45, opened 2017-08-21
Resolve placeholders in configuration

wolfs:
In this way, placeholders can be used in the scenario file.
E.g. one can define
```
defaults {
  run-using = tooling-api
  cleanup-tasks = ["clean"]
}

buildCache = ${defaults} {
  tasks = ["compileJava"]
  gradle-args = ["--build-cache"]
}
```


wolfs:
@oehme I added a test. Please have another look.

PR #48, opened 2017-10-24
Add support for benchmarking Bazel builds

eriwen:
Allow Bazel builds to be invoked by the --bazel flag and bazel-
specific configuration. Discover and execute from BAZEL_HOME as is
done for Maven. Added initial setup instructions to README.


eriwen:
Yes. Done with 8a31b04dade2231023a2c61cc048286c26069ad9

PR #49, opened 2017-10-30
Change chrome trace file rename to platform independent move

ghale:


ghale:
this was a mistake - closing.

PR #53, opened 2017-11-17
Allow profiling multiple iterations

oehme:
When using --profile together with --iterations,
the profiler used to overwrite the result for every
iteration, which was not very helpful.

Instead the profiler can now aggregate the data
from multiple iterations into one snapshot. It
pauses data collection when cleanup tasks are run.
Some profilers don't support pausing. These will
throw an exception if both cleanup tasks and multiple
iterations are requested.

This helps a lot when profiling fast-running scenarios
like configuration time or up-to-date tasks. A single
invocation usually doesn't contain enough samples to
trust it. Averaging over a dozen invocations however
gives results that show the real hotspots clearly.

Build scans and chrome traces remain unchanged and
create one snapshot per iteration.

### DIFF
--- a/src/main/java/org/gradle/profiler/hp/HonestProfilerControl.java
+++ b/src/main/java/org/gradle/profiler/hp/HonestProfilerControl.java
@@ -35,6 +35,7 @@ public class HonestProfilerControl implements ProfilerController {
 
     private final HonestProfilerArgs args;
     private final ScenarioSettings scenarioSettings;
+    private boolean recordedBefore = false;
 
     public HonestProfilerControl(final HonestProfilerArgs args, ScenarioSettings scenarioSettings) {
         this.args = args;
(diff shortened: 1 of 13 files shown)


src/main/java/org/gradle/profiler/hp/HonestProfilerControl.java
```diff
@@ -35,20 +35,34 @@
...
     @Override
-    public void start() throws IOException, InterruptedException {
+    public void startSession() throws IOException, InterruptedException {
+
+    }
+
+    @Override
+    public void startRecording() throws IOException, InterruptedException {
+        if (recordedBefore) {
+            throw new RuntimeException("Recording multiple iterations with cleanup runs in between is not supported by Honest Profiler");
```
adammurdoch: it would be good if this were somehow applied as a decoration over those implementations that can only start and stop once.

src/main/java/org/gradle/profiler/jprofiler/JProfilerController.java
```diff
@@ -27,7 +27,14 @@ public JProfilerController(ScenarioSettings settings, JProfilerConfig jProfilerC
     }
 
     @Override
-    public void start() throws IOException, InterruptedException {
+    public void startSession() throws IOException, InterruptedException {
+        if (!profileWholeLifeTime()) {
```
adammurdoch: this feels like something that could perhaps be taken care of at some higher level, as it's not really honouring the contract. The main loop might be aware in some way of profiling that is applied to the whole process lifetime and can avoid switching it on and off, possibly by simply using a no-op controller when using whole process profiling.

adammurdoch:
This is great, exactly what I need

src/main/java/org/gradle/profiler/hp/HonestProfilerControl.java
oehme: Extracted into a common superclass, so they only need to implement two methods instead of four.

src/main/java/org/gradle/profiler/jprofiler/JProfilerController.java
oehme: It's now using a no-op control in daemon-mode, just like the YourKitProfiler already did.

src/main/java/org/gradle/profiler/hp/HonestProfilerControl.java
melix: It would be nice to provide an option to _not_ fail if the profiler doesn't support pausing: we can still filter out some frames after recording, to remove the cleaning steps, for example.

oehme: You could change the implementation to actually support pausing instead of failing. I.e. you could remember the pause/resume times and filter frames that fall into the paused time. I just did the minimum work to port the existing implementation.

PR #56, opened 2018-02-05
Introduce Git revert and cache clearing mutators

lptr:
Now it is possible to declare mutators that mutate the state of the whole projectDir, and there are more hooks to tie into, too:

- `beforeScenario()`
- `beforeCleanup()`
- `afterCleanup(Throwable)`
- `beforeBuild()`
- `afterBuild(Throwable)`
- `afterScenario()`

This allows adding two new mutators:

- `GitRevertMutator` reverts the given Git commits before the actual build starts (but not before cleaup has finished), allowing the simulation of changes made by users.
- `ClearBuildCacheMutator` clears build caches at the given point in the scenario's execution.


lptr:
Build: https://builds.gradle.org/viewQueued.html?itemId=10817578

src/main/java/org/gradle/profiler/BuildMutator.java
```diff
@@ -1,9 +1,33 @@
...
+     */
+    default void beforeScenario() {};
+
+    /**
+     * Runs before each iteration if cleanup tasks are declared.
+     */
+    default void beforeCleanup() {};
+
+    /**
+     * Runs after each iteration of the build has finished.
```
oehme: ⭕️ cleanup

PR #57, opened 2018-02-14
Set system properties with profiling information

lptr:
The profiler now makes the following system properties available to Gradle builds:

- `org.gradle.profiler.scenario` -> name of scenario or `default` for ad-hoc scenarios
- `org.gradle.profiler.type` -> `WARM_UP` or `MEASUREMENT`
- `org.gradle.profiler.step` -> `CLEANUP` or `BUILD`
- `org.gradle.profiler.number` -> build counter starting from `1`


lptr:
@oehme I'm going to optimistically merge this, please do the review on the PR, and I can fix anything in a followup PR later. Thanks!

PR #59, opened 2018-04-03
Make chrome trace work with Gradle 4.7

wolfs:
This makes it possible to generate Chrome traces with Gradle 4.7.

The PR also breaks compatibility with older Gradle versions :(

Any suggestions how to improve the compatibility are welcome.

### DIFF
--- a/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
@@ -13,7 +13,7 @@ import spock.lang.Unroll
 class ProfilerIntegrationTest extends Specification {
 
     @Shared
-    List<String> supportedGradleVersions = ["3.3", "3.4.1", "3.5", "4.0", "4.1", "4.2.1"]
+    List<String> supportedGradleVersions = ["3.3", "3.4.1", "3.5", "4.0", "4.1", "4.2.1", "4.7-rc-1"]
     @Shared
     String minimalSupportedGradleVersion = supportedGradleVersions.first()
     @Shared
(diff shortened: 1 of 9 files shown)


oehme:
You can use reflection/proxies for the older versions. I think some of the adapters already did this.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -13,11 +13,11 @@ import spock.lang.Unroll
...
 
     @Shared
-    List<String> supportedGradleVersions = ["3.3", "3.4.1", "3.5", "4.0", "4.1", "4.2.1"]
+    String minimalSupportedGradleVersion = "3.3"
     @Shared
-    String minimalSupportedGradleVersion = supportedGradleVersions.first()
+    String latestSupportedGradleVersion = "4.7-rc-1"
     @Shared
-    String latestSupportedGradleVersion = supportedGradleVersions.last()
+    List<String> chromeTraceSupportedGradleVersions = [latestSupportedGradleVersion]
```
lptr: Shouldn't this be explicitly set to `4.7-rc-1`? Or maybe something like:

```groovy
 Set<String> chromeTraceSupportedGradleVersions = ['4.7-rc-1', latestSupportedGradleVersion]
```

lptr:
What happens now if I try to run with Gradle 4.4 + chrome trace?

wolfs:
I changed the code so now older Gradle version are still supported.

lptr:
Should we make running `gradle-profiler` a smoke test for Gradle to shake out problems like this?

wolfs:
@lptr I think that sounds like a good idea. So we would need to write a different set of tests for the `gradle-profiler` in `gradle/gradle`?

oehme:
I think we can just do a mockup of what the Gradle profiler does in gradle/gradle. I.e. have a simple build script that uses the most current build operation API and add a comment saying "If this fails, you need to upgrade the Gradle profiler.

PR #71, opened 2018-06-01
Add a mutator for Kotlin that adds a function in the end of the file

NikitaKozlov:
I created a very simple mutator for Kotlin, it just adds a function in the end of the file. Unfortunately I couldn't find a good parser for Kotlin, so nobody is calling this new function. Hacking my way through with a regex sounds like too much of a hack.


NikitaKozlov:
I'm not sure if I understood you correctly. 

Do you want me to change `ApplyAbiChangeToJavaSourceFileMutator` and `ApplyNonAbiChangeToJavaSourceFileMutator` to handle Kotlin? They look Java-specific with `JavaParser` and parent that handles parsing of the original file to `CompilationUnit`. 

Or you want me to write an alternative hierarchy of classes: `ApplyAbiChangeToKotlinSourceFileMutator`, `ApplyNonAbiChangeToKotlinSourceFileMutator` and potentially `AbstractKotlinSourceFileMutator`?

In any case it will be very problematic to make Non-ABI mutator without a good Kotlin parser. I can rename `ApplyChangeToKotlinSourceFileMutator` to `ApplyAbiChangeToKotlinSourceFileMutator`, but 
its contract will differ from `ApplyAbiChangeToJavaSourceFileMutator` - without the Kotlin parser I won't be able to ensure that some other methods call newly created ones.

As alternative I can propose to use [ANTLR4 grammar for Kotlin](https://github.com/antlr/grammars-v4/tree/master/kotlin) to parse Kotlin files, but this solution is not nearly as elegant and concise as the one used for Java files. 

What would you prefer?

oehme:
What I had in mind was this:

AbiChangeMutator delegates to either KotlinAbiChangeMutator or JavaAbiChangeMutator, depending on the file extension. Same for NonAbiChangeMutator.

Doing a non-abi change in Kotlin is just as easy as an abi change: Just make the method private.

NikitaKozlov:
I hope I implemented what you had in mind. I think there is a problem now that it is impossible to change both Java and Kotlin file in a single scenario. I think this functionality is valuable to measure incremental build of mixed projects. But it is not a part of this PR.

oehme:
Thank you Nikita!

PR #73, opened 2018-06-07
Simpler flamegraphs

oehme:
Improve the flame graph simplification so that stacks become much shorter and related stacks (e.g. task execution from multiple threads) are shown in the same "tower".


oehme:
And here's what `gradle assemble` on the gradle/gradle repo looks like with this change:
screen shot 2018-06-07 at 11 03 59

wolfs:
Nice! Will this also be ported to our performance test infrastructure in `gradle/gradle`?

PR #83, opened 2018-08-18
Fix crash starting HP

SUPERCILEX:


oehme:
Thank you for the fix!

SUPERCILEX:
Yeah, that's what I ended up doing in the end. Was looking for flame graphs though, and it wasn't until I read the source code that I realized I had to hop into to linux to get them.

PR #87, opened 2018-09-17
Add mutation to clear transform caches

wolfs:
We also enable the chrome trace for Gradle 5.x

### DIFF
--- a/src/main/java/org/gradle/profiler/mutations/ClearArtifactTransformCacheMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/ClearArtifactTransformCacheMutator.java
@@ -1,0 +1,41 @@
+package org.gradle.profiler.mutations;
+
+import org.apache.commons.io.FileUtils;
+import org.gradle.profiler.BuildMutator;
+
+import java.io.File;
+import java.io.FilenameFilter;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+public class ClearArtifactTransformCacheMutator extends AbstractCleanupMutator {
+
+	public ClearArtifactTransformCacheMutator(File gradleUserHome, CleanupSchedule schedule) {
+		super(gradleUserHome, schedule, "transforms-");
+	}
+
+	@Override
+	protected void cleanupCacheDir(File cacheDir) {
+		filesAsStream(
+				cacheDir,
+				(dir, name) -> name.startsWith("files-"))
+				.flatMap(file -> filesAsStream(file, (dir, name) -> !name.endsWith(".lock")))
+				.forEach(FileUtils::deleteQuietly);
+	}
+
+	private Stream<File> filesAsStream(File dir, FilenameFilter filter) {
+		return Arrays.stream(Objects.requireNonNull(dir.listFiles(filter)));
+	}
+
+	public static class Configurator extends AbstractCleanupMutator.Configurator {
+		public Configurator(File gradleUserHome) {
+			super(gradleUserHome);
+		}
+
+		@Override
+		protected BuildMutator newInstance(File gradleUserHome, AbstractCleanupMutator.CleanupSchedule schedule) {
+			return new ClearArtifactTransformCacheMutator(gradleUserHome, schedule);
+		}
+	}
+}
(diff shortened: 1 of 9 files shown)


src/main/java/org/gradle/profiler/mutations/ClearArtifactTransformCacheMutator.java
```diff
@@ -0,0 +1,41 @@
...
+		public Configurator(File gradleUserHome) {
+			super(gradleUserHome);
+		}
+
+		@Override
+		protected BuildMutator newInstance(File gradleUserHome, AbstractCleanupMutator.CleanupSchedule schedule) {
+			return new ClearArtifactTransformCacheMutator(gradleUserHome, schedule);
+		}
+	}
+}
```
lptr: Let's reformat to use spaces. :/ Can you please add an `.editorconfig` to ensure we don't add more code with tabs?

subprojects/chrome-trace/src/main/java/org/gradle/trace/listener/BuildOperationListenerAdapter.java
```diff
@@ -18,7 +18,7 @@ static BuildOperationListenerAdapter create(Gradle gradle, TraceResult traceResu
         if (gradleVersion.matches("4\\.[0-6](\\D.*)?")) {
             return new Gradle40BuildOperationListenerAdapter(gradle, new Gradle40BuildOperationListenerInvocationHandler(traceResult));
         }
-        if (gradleVersion.startsWith("4.")) {
+        if (gradleVersion.startsWith("4.") || gradleVersion.startsWith("5.")) {
```
lptr: This sounds flimsy. Shouldn't this be `> 4.0`? I.e. add a `VersionNumber` type that can parse and compare version numbers and use that?

subprojects/chrome-trace/src/main/java/org/gradle/trace/listener/BuildOperationListenerAdapter.java
```diff
@@ -1,28 +1,35 @@
...
-        if (gradleVersion.equals("4.0-milestone-2")) {
+        if (gradleVersion.equals(version("4.0-milestone-2"))) {
             return new Gradle35BuildOperationListenerAdapter(gradle, new Gradle40BuildOperationListenerInvocationHandler(traceResult));
         }
-        if (gradleVersion.matches("4\\.[0-6](\\D.*)?")) {
+        if (inVersionRange(gradleVersion, "4.0", "4.8")) {
             return new Gradle40BuildOperationListenerAdapter(gradle, new Gradle40BuildOperationListenerInvocationHandler(traceResult));
         }
-        if (gradleVersion.startsWith("4.") || gradleVersion.startsWith("5.")) {
+        if (gradleVersion.compareTo(version("4.8")) >= 0) {
```
lptr: What happened to 4.7? Shouldn't `Gradle47BuildOperationListenerInvocationHandler` be used on that version, too?

wolfs: Yes, I failed to parse the regex correctly... Fixed.

subprojects/chrome-trace/src/main/java/org/gradle/trace/listener/BuildOperationListenerAdapter.java
```diff
@@ -1,28 +1,35 @@
...
             return new Gradle47BuildOperationListenerAdapter(gradle, new Gradle47BuildOperationListenerInvocationHandler(traceResult));
         }
         throw new IllegalStateException("Gradle version " + gradleVersion + " not supported, 3.3+ required");
     }
 
+    static boolean inVersionRange(GradleVersion gradleVersion, String lower, String upper) {
+        return gradleVersion.compareTo(version(lower)) >= 0 && gradleVersion.compareTo(version(upper)) < 0;
+    }
```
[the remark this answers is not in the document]
wolfs: If we make the upper limit inclusive, we need to know which Gradle versions had patch releases. That's why I made the bound exclusive.

I wrote out the comparisons, but the verbose Java `compare` calls made it worse IMO.

lptr:
LGTM

wolfs:
@lptr I added an integration test. PTAL!

Issue #85, opened 2018-09-07
Allow to mutate multiple files with one mutation 


NikitaKozlov:
It might be very useful to change multiple files with a single mutation, for example change source files in different modules. Ideally I would love to write a scenario with something like this:
```
apply-abi-change-to = ["filePath1", "filePath2"]
```
Please, let me know if you are interested in the PR for this change. I tried it locally and I believe it is a fairly easy change.

oehme:
Makes perfect sense, a PR is welcome :)

oehme:
Fixed by #98

PR #98, opened 2018-11-06
Added support for mutation multiple files with a single mutation

NikitaKozlov:
Issue https://github.com/gradle/gradle-profiler/issues/85
Main change was in the way config's values are parsed, but lucky `ConfigUtil.strings()` can parse both array of strings and a single string and return list of strings. Then this corresponding `Mutator` is created for each of the strings. Later all mutators are passed to `CompositeBuildMutator`, that is public now.

Unfortunately I couldn't find an elegant way to test all the added functionality, so I just made sure that `ScenarioLoader` can load a scenario without an error.

I like how concise current README is, so I simply added an example for mutation of multiple files with a single mutation. I also added to the README that ABI and non-ABI changes can apply to Kotlin files as well.


Nohands253:
https://www.easyhits4u.com/account/stats/

oehme:
Thank you Nikita!

PR #99, opened 2018-11-06
Add error message for missing warm-up for profile

sunyal:
- a pid is required for the profiler to start recording but without a warm-up a pid won't be supplied. This provides a better error message than an NPE.


oehme:
Thank you Alex!

PR #106, opened 2019-01-04
Add buildscan-additional-repo option for running profiler in Intranet

patliu85:
When users run gradle-profiler to profile buildscan on Intranet environment that does not have access to Internet/default repo (https://plugins.gradle.org/m2), it gets 'Network is unreachable' error.

Adding --buildscan-additional-repo option would allow the users to provide their internal repository/artifacts (their internal repo that mirrors github), hence the profiler could get all required files from internal repo.

E.g.: --buildscan-additional-repo https://artifactory.mycompany.com/java-virtual

### DIFF
--- a/src/main/java/org/gradle/profiler/bs/BuildScanInitScript.java
+++ b/src/main/java/org/gradle/profiler/bs/BuildScanInitScript.java
@@ -22,15 +22,28 @@ import java.io.PrintWriter;
 public class BuildScanInitScript extends GeneratedInitScript {
 
     private final String version;
+    private final String additionalRepo;
 
-    public BuildScanInitScript(String version) {
+    public BuildScanInitScript(String version, String additionalRepo) {
         this.version = version;
+        this.additionalRepo = additionalRepo;
     }
 
     @Override
     public void writeContents(final PrintWriter writer) {
         writer.write("initscript {\n");
         writer.write("    repositories {\n");
+        if (additionalRepo != null) {
+            /*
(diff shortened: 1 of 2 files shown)


patliu85:
Hi Stefan,
Thanks for your reply.
I already added my company's repo in build.gradle 

repositories {
    maven { url "https://repo.mycompany.com" }...}
to run "./gradlew installDist" to generate those files & dirs in build/ directory

But when I run the profiler (build/install/gradle-profiler/bin/gradle-profiler)since you have "https://plugins.gradle.org/m2" hardcoded in "src/main/java/org/gradle/profiler/bs/BuildScanInitScript.java"the profiler would always try to access this repo (https://plugins.gradle.org/m2)
That's why I was trying to find a way to overwrite this part. Please suggest how to overwrite this part without modifying those Java files?
Thank you.
Best,Patrick


    On Sunday, January 6, 2019, 2:26:37 AM PST, Stefan Oehme <notifications@github.com> wrote:  
 
 
Thank you for the suggestion, but I don't think we should go this way. In fact I think we should remove the --build-scan-version option. If you want to customize which build scan plugin version you want to use and where you want to get it from, add the scan plugin to your build (either in your build.gradle or via an init script).

I've just fixed the profiler so it will no longer interfere with an already applied build scan plugin in your build.

oehme:
That init script is no longer used if you already have the plugin applied in your build. Please upgrade the profiler.

PR #110, opened 2019-01-21
Add support for profiling/benchmarking a simulated Android Studio sync

adammurdoch:
This PR adds support for measuring a mock up of the Android Studio sync.

The logic is a copy of the code from the `gradle/gradle` performance tests. At some point it would be good to reuse this logic instead, ideally from the Android Studio source. This PR does not cover that issue.

### DIFF
--- a/src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
+++ b/src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
@@ -1,0 +1,56 @@
+package org.gradle.profiler;
+
+import com.android.builder.model.AndroidProject;
+import org.gradle.tooling.BuildController;
+import org.gradle.tooling.events.ProgressListener;
+import org.gradle.tooling.model.gradle.BasicGradleProject;
+import org.gradle.tooling.model.gradle.GradleBuild;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * A mock-up of Android studio sync.
+ */
+public class AndroidStudioSyncAction implements BuildAction {
+    @Override
+    public String getDisplayName() {
+        return "simulate Android Studio sync";
+    }
+
+    @Override
+    public void run(BuildInvoker buildInvoker, List<String> tasks, List<String> gradleArgs, List<String> jvmArgs) {
+        gradleArgs = new ArrayList<>(gradleArgs);
+        gradleArgs.add("-Dcom.android.build.gradle.overrideVersionCheck=true");
+        gradleArgs.add("-Pandroid.injected.build.model.only=true");
+        gradleArgs.add("-Pandroid.injected.build.model.only.versioned=3");
+        gradleArgs.add("-Pandroid.builder.sdkDownload=true");
+        tasks = new ArrayList<>(tasks);
+        tasks.add("generateDebugSources");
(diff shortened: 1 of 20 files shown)


src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * A mock-up of Android studio sync.
+ */
+public class AndroidStudioSyncAction implements BuildAction {
```
oehme: The way Android studio syncs is evolving quite rapidly, so what we do here is probably already outdated. I think I copied it when Android 3.0 came out. It would be great if we had a standalone library that we could just reference.

cc @ducrohet

src/main/java/org/gradle/profiler/BuildActionInvoker.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public BuildActionInvoker(List<String> jvmArgs, List<String> gradleArgs, BuildInvoker buildInvoker, PidInstrumentation pidInstrumentation, Consumer<BuildInvocationResult> resultsConsumer) {
+        this.jvmArgs = jvmArgs;
+        this.gradleArgs = gradleArgs;
+        this.buildInvoker = buildInvoker;
+        this.pidInstrumentation = pidInstrumentation;
+        this.resultsConsumer = resultsConsumer;
+    }
+
+    public BuildInvocationResult runBuild(Phase phase, int buildNumber, BuildStep buildStep, List<String> tasks, BuildAction buildAction) {
```
oehme: I think it would be clearer if the build action contained the tasks it will run instead of taking them as an argument.

src/main/java/org/gradle/profiler/BuildActionInvoker.java
```diff
@@ -0,0 +1,68 @@
+package org.gradle.profiler;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import static org.gradle.profiler.Logging.startOperation;
+
+public class BuildActionInvoker {
```
oehme: This name is a bit too close to `BuildInvoker`. Maybe this could be called `InstrumentedBuild` or `BuildUnderTest`?

src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
ducrohet: Yes we want to extract this code in a library. I think it'll happen in a few months.

src/main/java/org/gradle/profiler/BuildActionInvoker.java
adammurdoch: Yes, I thought so too, but was going to do this as a follow-up PR. I can change it in this PR before merging if you'd prefer.

src/main/java/org/gradle/profiler/BuildActionInvoker.java
adammurdoch: I'll see if I can come up with some better names.

src/main/java/org/gradle/profiler/BuildActionInvoker.java
oehme: Nothing that holds up this PR.

PR #111, opened 2019-01-21
Fix build scan configuration for build scan plugin v2.0+

eriwen:
build scan plugin v2 removed 'licenseAgreementUrl' in favor of
'termsOfServiceUrl'.

This probably breaks build scan integration for early plugin versions, however.


eriwen:
@oehme sure... this is done with c566b98. All good now?

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -340,6 +340,23 @@ println "<daemon: " + gradle.services.get(org.gradle.internal.environment.Gradle
         assertBuildScanPublished(BuildScanProfiler.defaultBuildScanVersion(GradleVersion.version(minimalSupportedGradleVersion)))
     }
 
+    def "profiles build using Build Scans with latest supported Gradle version"() {
```
oehme: You can replace this new test with a where: clause in the existing test.

PR #115, opened 2019-02-08
Add mutation support for Android Layout files

kkoser:
Changes to Android layout files are a common situation to profile, especially when using Android Data binding, which
will trigger its own compiler. This PR adds support for this form of mutations to the profiler. It supports mutating
both normal layout files and data binding layouts, which have a special format.

This is done by inserting a non-visible view into the bottom of the layout with a random ID. The ID is important
as it guanartees that the API of the data binding model for this layout will also change.

The one thing not supported currently is mutating a layout file with a single element inside it, as in that case there
is nowhere for a fake element to be inserted. Likely the solution to this is to make the fake element a wrapper layout,
but I don't see much use for that use case right now, so I left it out.

### DIFF
--- a/src/main/java/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutator.java
@@ -2,6 +2,14 @@ package org.gradle.profiler.mutations;
 
 import java.io.File;
 
+/**
+ * This class applies a mutation to an Android layout file by adding an extra, hidden view at the bottom of the layout.
+ * It supports both DataBinding layouts and traditional Android layouts.
+ *
+ * Note: This mutator does not support layouts with a single view in them - it requires a valid ViewGroup to contain the
+ * new view. Attempting to mutate a layout that does not have a ViewGroup as its root element will result in an
+ * invalid layout file.
+ */
 public class ApplyChangeToAndroidLayoutFileMutator extends AbstractFileChangeMutator {
     public ApplyChangeToAndroidLayoutFileMutator(File sourceFile) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutator.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.File;
+
+public class ApplyChangeToAndroidLayoutFileMutator extends AbstractFileChangeMutator {
+    public ApplyChangeToAndroidLayoutFileMutator(File sourceFile) {
+        super(sourceFile);
+    }
+
+    @Override
+    protected void applyChangeTo(StringBuilder text) {
+        // We need to handle both databinding layouts, which end with </layout>, and non, which can end with any layout item closing tag
```
oehme: Can you replace all these inline comments with a class-level comment describing what this does instead?

kkoser:
Thanks @oehme for the quick response! I added some class level comments, and also added the flag in the README since I realized I forgot that part :) LMK if there's anything else I can do to improve it. Love the tool!

src/main/java/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutator.java
oehme: I meant to also remove the inline comments, as they go stale and are redundant with the method names :)

kkoser: Ah sorry I misunderstood, let me update

oehme:
Thank you very much! :)

PR #118, opened 2019-03-22
Add support for benchmarking or profiling a cold daemon

adammurdoch:
Add a `--cold-daemon` command-line option, which stops the daemons prior to each warm up and measured build. Profiling a cold daemon is supported for all of the profilers.

Also automatically download an async profiler distribution if one is not available locally.

Start to bust up the integration test into separate classes, and add some test coverage for each of the profilers.


src/main/java/org/gradle/profiler/InstrumentingProfiler.java
```diff
@@ -0,0 +1,172 @@
...
+ * </ul>
+ *
+ * <p>The profiler may support starting recording multiple times for a given JVM. The implementation should indicate this by overriding {@link #canRestartRecording()}.</p>
+ */
+public abstract class InstrumentingProfiler extends Profiler {
+    @Override
+    public JvmArgsCalculator newJvmArgsCalculator(ScenarioSettings settings) {
+        validate(settings);
+        if (!settings.getScenario().getInvoker().isReuseDaemon()) {
+            // When the daemon is not reused, there is no need to instrument the warm ups
```
oehme: I'd very much prefer method-level comments instead of inline ones, as the latter grow stale so quickly.

adammurdoch:
Thanks for reviewing

src/main/java/org/gradle/profiler/InstrumentingProfiler.java
adammurdoch: I've updated the javadocs.

PR #119, opened 2019-03-27
Allow time to first task to be measured when benchmarking

adammurdoch:
Add a `--benchmark-config-time` to enable some coarse-grained measurement of configuration time when benchmarking. In this PR the time to first task execution is measured, or actually the time to the start of the root 'run task graph' build operation (which isn't quite the same thing).


src/main/java/org/gradle/profiler/Main.java
```diff
@@ -61,21 +62,29 @@ public void run(String[] args) throws Exception {
...
-                    } else {
-                        gradleScenarioInvoker.run((GradleScenarioDefinition) scenario, settings, benchmarkResults);
-                    }
-
+                    invoker.run(scenario, settings, resultConsumer);
                 } catch (Throwable t) {
                     t.printStackTrace();
                     failures.add(t);
+                } finally {
+                    // write the results up to this point
```
oehme: What does this mean? Does it write more results later?

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -0,0 +1,50 @@
...
+import org.gradle.initialization.BuildRequestMetaData;
+import org.gradle.internal.operations.*;
+import org.gradle.trace.stream.AsyncWriter;
+
+import java.io.File;
+
+@SuppressWarnings("unused")
+public class BuildOperationTrace {
+    public static void start(GradleInternal gradle, File outFile) {
+        BuildOperationListenerManager manager = gradle.getServices().get(BuildOperationListenerManager.class);
```
oehme: ❌ I don't think this will work in all supported Gradle versions. Could we reuse the logic we have in chrome-trace here?

adammurdoch: At this stage, we're really only interested in measuring this for recent versions of Gradle, so I don't want to put any real effort into making this work for older versions. I did have a quick look at reusing the logic but didn't want the overhead of the adapting layers. But we can if you think we should. For now, I'd rather we just fail if configuration time instrumentation is requested for a Gradle version that it doesn't work with. We can support older versions later if someone needs it.

We also have a similar question with Java version. Some of the instrumentation code is Java 8 but the older versions will work even back to Java 6. I think it's probably ok to just work with Java 8 for now and fail if people try to measure using something older.

adammurdoch:
Mostly people don't care about the configuration time details, so I used a switch to keep instrumentation overhead to a minimum when the information isn't required. It's also there so we don't have to support the instrumentation on every Gradle version.

I'd also like to add a few more samples, which means a little more overhead and backwards compatibility effort.

src/main/java/org/gradle/profiler/BenchmarkResultCollector.java
```diff
@@ -56,15 +54,21 @@ public void write() throws IOException {
         private final ScenarioDefinition scenario;
         private final BuildScenarioResult baseline;
         private final List<BuildInvocationResult> results = new ArrayList<>();
-        private DescriptiveStatistics statistics;
+        private final List<String> samples;
```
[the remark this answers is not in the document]
adammurdoch: Good idea. I'll update the PR

src/main/java/org/gradle/profiler/Main.java
adammurdoch: I've updated the comment. Hopefully it's clearer now.

adammurdoch:
@oehme I've pushed a few updates. Could you take another look?

src/main/java/org/gradle/profiler/Main.java
```diff
@@ -74,27 +73,31 @@ public void run(String[] args) throws Exception {
...
-                Consumer<BuildInvocationResult> resultConsumer = benchmarkResults.scenario(scenario, invoker.samplesFor(settings));
                 try {
-                    invoker.run(scenario, settings, resultConsumer);
+                    invoker.run(scenario, settings, benchmarkResults);
                 } catch (Throwable t) {
                     t.printStackTrace();
                     failures.add(t);
                 } finally {
-                    // write the results up to this point
+                    // Write the current results and generate the reports, so that if this process crashes the results (which may have taken quite some time to collect) are not lost.
```
oehme: Is this a common scenario?

adammurdoch: Maybe not any more. By "crash" I mean "some error that causes gradle-profiler to exit without writing the results". This happened to me a few times in a few different ways. For example, starting some long running benchmarks on a remote machine, forgetting to use `screen` or similar and then having the ssh connection die and the gradle-profiler JVM killed by the OS. Or when we messed up the build failure handling logic and the gradle-profiler JVM exited with an exception without writing the results instead of continuing with the next scenario. I can't say I ever saw the JVM crash, eg segfault, though.

It's just to add a little bit of robustness in cases where we don't deal with problems in a graceful way.

adammurdoch:
@oehme, thanks for reviewing the changes.

PR #120, opened 2019-04-06
Allow a scenario to specify that a cold daemon should be used

adammurdoch:
This PR adds a new `daemon` property that can be used in a scenario file.


adammurdoch:
@oehme thanks for reviewing.

PR #121, opened 2019-04-11
Add ability to override jvm arguments

AndrewReitz:
Adds `jvm-args` to the scenario file to allow for
overriding and testing of different jvm options.

### DIFF
--- a/src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
+++ b/src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
@@ -70,6 +70,8 @@ public class GradleScenarioInvoker extends ScenarioInvoker<GradleScenarioDefinit
             buildConfiguration.printVersionInfo();
 
             List<String> allBuildsJvmArgs = new ArrayList<>(buildConfiguration.getJvmArguments());
+            allBuildsJvmArgs.addAll(scenario.getJvmArgs());
+
             for (Map.Entry<String, String> entry : scenario.getSystemProperties().entrySet()) {
                 allBuildsJvmArgs.add("-D" + entry.getKey() + "=" + entry.getValue());
             }
--- a/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
@@ -1605,4 +1605,33 @@ buildTarget {
         logFile.find("> org.gradle.profiler.number = 1").size() == 4
         logFile.find("> org.gradle.profiler.number = 2").size() == 4
     }
+
+    def "override jvm args"() {
+        given:
+        def scenarios = file('performance.scenario')
+        scenarios.text = """
+            help {
+                tasks = ["jvmArgs"]
+                jvm-args = ["-Xmx2G", "-Xms1G"]
+            }
+        """
+
+        getBuildFile() << """
+            import java.lang.management.ManagementFactory
+
+            task jvmArgs {
+                doFirst {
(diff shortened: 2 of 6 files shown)


src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
```diff
@@ -69,7 +67,14 @@ public void doRun(GradleScenarioDefinition scenario, InvocationSettings settings
         try {
             buildConfiguration.printVersionInfo();
 
-            List<String> allBuildsJvmArgs = new ArrayList<>(buildConfiguration.getJvmArguments());
+            List<String> allBuildsJvmArgs = new ArrayList<>();
+            if (!scenario.getJvmArgs().isEmpty()) {
+                allBuildsJvmArgs.addAll(buildConfiguration.getJvmArguments().stream().filter(arg -> !arg.startsWith("-X")).collect(Collectors.toList()));
```
oehme: ❌ There should be no filtering logic here.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1605,4 +1605,24 @@ buildTarget {
         logFile.find("> org.gradle.profiler.number = 1").size() == 4
         logFile.find("> org.gradle.profiler.number = 2").size() == 4
     }
+
+    def "over rides jvm args"() {
```
oehme: overrides

src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
AndrewReitz: Screen Shot 2019-04-11 at 7 10 54 PM

It looks like we will end up with two -Xmx (and possibly other) args if we don't filter, although the test still pass. Does the last argument always win?

AndrewReitz:
@oehme would you like me to update the README as well?

src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
oehme: Yes that's how the jvm and most other command line programs handle arguments

oehme:
Yes, that would be great

AndrewReitz:
@oehme Looks like the test I added causes the build to report a failure even though all the tests pass. Do you have a suggestion for testing?

gradle-profiler.iml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
oehme: Can you please remove this file?

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1605,4 +1605,21 @@ buildTarget {
...
         logFile.find("> org.gradle.profiler.number = 2").size() == 4
     }
+
+    def "overrides jvm args"() {
+        given:
+        def scenarios = file('performance.scenario')
+        scenarios.text = """
+help {
+    tasks = ["help"]
+    jvm-args = ["-Xmx1m"]
```
oehme: The out of memory error is coming from this test - Teamcity scrapes the log, sees it and reports it as an error. I'd instead use the MemoryMxBean in the build script to print out the max memory in order to verify the setting.

oehme: We could of course deactivate this check, but I think writing the test without relying on an error would be cleaner.

oehme:
Thanks @AndrewReitz!

PR #123, opened 2019-04-28
Fix NPE when cleaning transform or build cache when files with matching names are present in the caches directory

adammurdoch:
Also don't silently ignore failures to delete these directories, as leaving these files behind can affect the results.


adammurdoch:
@oehme thanks for reviewing.

PR #127, opened 2019-05-06
Add git-commit to Profiler Scenarios

AndrewReitz:
Adds ability to checkout a previous commit, run the profiler
against that commit, and then reverts back to the original checked
out branch.

Signed-off-by: Andrew Reitz <aj.reitz@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ A scenario can define changes that should be applied to the source before each b
 - `show-build-cache-size`: Shows the number of files and their size in the build cache before scenario execution, and after each cleanup and build round..
 - `git-checkout`: Checks out a specific commit for the build step, and a different one for the cleanup step.
 - `git-revert`: Reverts a given set of commits before the build and resets it afterward.
+- `git-commit`: Checks out a commit and runs the profiler on that commit.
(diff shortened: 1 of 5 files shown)


oehme:
How is this different from git-checkout?

AndrewReitz:
I thought that git-checkout was supposed to simulate large changes between the current branch and a previous commit, where as this is supposed to profile a specific commit.

AndrewReitz:
Ok, I'll update to so that. Would you be open to a mutator that checks out different branches as well?

oehme:
I don't think there should be a difference. You checkout a reference, which can be any commit or branch or tag.

AndrewReitz:
@oehme When you say to change the second commit is optional, do you mean just allow cleanup to be null, since it looks like that is currently possible without any modifications to the code, or is there something else that would need to be done?

oehme:
If it already works, all the better :) I didn't check.

AndrewReitz:
I created tests to verify. Would you be interested in me submitting a pr?

oehme:
Of course 👍

PR #129, opened 2019-05-11
Add tests for null clean in GitCheckoutMutator

AndrewReitz:
Tests to ensure only the build value can be set on the
GitCheckoutMutator allowing for checkouts of different commits and
running benchmarks on them.

Signed-off-by: Andrew Reitz <aj.reitz@gmail.com>


oehme:
Thanks @AndrewReitz!

PR #139, opened 2019-07-02
Update README to point to appropriate flag

seanabraham:
The flag `--benchmark-config-time` does exist wheres `--measure-config-time` does not.


oehme:
Thanks Sean! Would you mind signing off your commit so the DCO check passes?

seanabraham:
@oehme done!

oehme:
Thanks!

Issue #10, opened 2016-12-27
Allow TAPI model request to be profiled and benchmarked


adammurdoch:

oehme:
Initial support was added in https://github.com/gradle/gradle-profiler/commit/9e480b1e963f21b6d3b5456b204bb31ec9d5d713, made more generic in https://github.com/gradle/gradle-profiler/commit/2cab59c1ea34808f252eeedb63db4ba675d9e5b3

PR #140, opened 2019-07-22
Allow capturing build operation times

wolfs:
This PR allows using a `--benchmark-build-op` command line argument which then captures the measurements for a particular build operation.

E.g. running
```
gradle-profiler --scenario-file performance.scenarios --benchmark --benchmark-build-op org.gradle.api.internal.tasks.SnapshotTaskInputsBuildOperationType --benchmark-build-op org.gradle.api.internal.tasks.execution.ExecuteTaskBuildOperationType --gradle-version 5.5.1 upToDate
```

yields the output:

Sample | execution | org.gradle.api.internal.tasks.SnapshotTaskInputsBuildOperationType | org.gradle.api.internal.tasks.execution.ExecuteTaskBuildOperationType
-- | -- | -- | --
warm-up build #1 | 6397 | 4011 | 4555
warm-up build #2 | 1011 | 1005 | 1109
warm-up build #3 | 873 | 1003 | 1098
warm-up build #4 | 811 | 928 | 993
warm-up build #5 | 734 | 778 | 985
warm-up build #6 | 808 | 920 | 974
measured build #1 | 626 | 758 | 820
measured build #2 | 660 | 800 | 855
measured build #3 | 633 | 900 | 963
measured build #4 | 593 | 724 | 769
measured build #5 | 552 | 684 | 733
measured build #6 | 564 | 735 | 787
measured build #7 | 534 | 803 | 849
measured build #8 | 587 | 811 | 866
measured build #9 | 512 | 725 | 764
measured build #10 | 608 | 687 | 728
mean | 586,90 | 762,70 | 813,40
min | 512,00 | 684,00 | 728,00
25th percentile | 547,50 | 714,75 | 756,25
median | 590,00 | 746,50 | 803,50
75th percentile | 627,75 | 805,00 | 857,75
max | 660,00 | 900,00 | 963,00
stddev | 46,70 | 66,53 | 72,31

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,9 +41,7 @@ dependencies {
     testRuntime 'org.objenesis:objenesis:2.6'
 }
 
-allprojects {
-    sourceCompatibility = 1.8
-}
+sourceCompatibility = 1.8
 mainClassName = "org.gradle.profiler.Main"
 
 processResources {
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
 }
 
 allprojects {
-    plugins.withId('java') {
+    pluginManager.withPlugin('java') {
         java {
             sourceCompatibility = JavaVersion.VERSION_1_8
             targetCompatibility = JavaVersion.VERSION_1_8
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,9 @@ dependencies {
     testRuntime 'org.objenesis:objenesis:2.6'
 }
 
-sourceCompatibility = 1.8
+allprojects {
+    sourceCompatibility = 1.8
+}
 mainClassName = "org.gradle.profiler.Main"
 
 processResources {
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,14 @@ dependencies {
     testRuntime 'org.objenesis:objenesis:2.6'
 }
 
-sourceCompatibility = 1.8
+allprojects {
+    plugins.withId('java') {
+        java {
+            sourceCompatibility = JavaVersion.VERSION_1_8
+            targetCompatibility = JavaVersion.VERSION_1_8
+        }
+    }
+}
 mainClassName = "org.gradle.profiler.Main"
 
 processResources {
--- a/src/main/java/org/gradle/profiler/CommandLineParser.java
+++ b/src/main/java/org/gradle/profiler/CommandLineParser.java
@@ -1,6 +1,10 @@
 package org.gradle.profiler;
 
-import joptsimple.*;
+import joptsimple.ArgumentAcceptingOptionSpec;
+import joptsimple.OptionException;
+import joptsimple.OptionParser;
+import joptsimple.OptionSet;
+import joptsimple.OptionSpecBuilder;
 
 import java.io.File;
 import java.io.IOException;
@@ -39,6 +43,8 @@ class CommandLineParser {
             .defaultsTo("jfr");
         Profiler.configureParser(parser);
         OptionSpecBuilder benchmarkOption = parser.accepts("benchmark", "Collect benchmark metrics");
+        ArgumentAcceptingOptionSpec<String> benchmarkBuildOperation = parser.accepts("benchmark-build-op", "Collect benchmark metrics for build operation")
+            .withRequiredArg();
         OptionSpecBuilder noDaemonOption = parser.accepts("no-daemon", "Do not use the Gradle daemon");
         OptionSpecBuilder coldDaemonOption = parser.accepts("cold-daemon", "Use a cold Gradle daemon");
         OptionSpecBuilder cliOption = parser.accepts("cli", "Uses the command-line client to connect to the daemon");
@@ -73,6 +79,7 @@ class CommandLineParser {
         Integer warmups = parsedOptions.valueOf(warmupsOption);
         Integer iterations = parsedOptions.valueOf(iterationsOption);
         boolean measureConfig = parsedOptions.has(measureConfigOption);
+        List<String> benchmarkedBuildOperations = parsedOptions.valuesOf(benchmarkBuildOperation);
 
         List<String> targetNames = parsedOptions.nonOptionArguments().stream().map(Object::toString).collect(Collectors.toList());
         List<String> versions = parsedOptions.valuesOf(versionOption);
@@ -110,7 +117,23 @@ class CommandLineParser {
                 sysProperties.put(value.substring(0, sep), value.substring(sep + 1));
             }
         }
-        return new InvocationSettings(projectDir, profiler, benchmark, outputDir, invoker, dryRun, scenarioFile, versions, targetNames, sysProperties, gradleUserHome, warmups, iterations, measureConfig);
+        return new InvocationSettings(
+            projectDir,
+            profiler,
+            benchmark,
+            outputDir,
+            invoker,
+            dryRun,
+            scenarioFile,
+            versions,
+            targetNames,
+            sysProperties,
+            gradleUserHome,
+            warmups,
+            iterations,
+            measureConfig,
+            benchmarkedBuildOperations
+        );
     }
 
     private File findOutputDir() {
--- a/src/main/java/org/gradle/profiler/CommandLineParser.java
+++ b/src/main/java/org/gradle/profiler/CommandLineParser.java
@@ -50,7 +50,7 @@ class CommandLineParser {
         OptionSpecBuilder noDaemonOption = parser.accepts("no-daemon", "Do not use the Gradle daemon");
         OptionSpecBuilder coldDaemonOption = parser.accepts("cold-daemon", "Use a cold Gradle daemon");
         OptionSpecBuilder cliOption = parser.accepts("cli", "Uses the command-line client to connect to the daemon");
-        OptionSpecBuilder measureConfigOption = parser.accepts("benchmark-config-time", "Include a breakdown of configuration time in benchmark results");
+        OptionSpecBuilder measureConfigOption = parser.accepts("measure-config-time", "Include a breakdown of configuration time in benchmark results");
         OptionSpecBuilder dryRunOption = parser.accepts("dry-run", "Verify configuration");
         OptionSpecBuilder bazelOption = parser.accepts("bazel", "Benchmark scenarios using Bazel");
         OptionSpecBuilder buckOption = parser.accepts("buck", "Benchmark scenarios using buck");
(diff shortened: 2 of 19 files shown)


src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
```diff
@@ -21,9 +24,24 @@ public Duration extractFrom(GradleBuildInvocationResult result) {
...
+    public static Sample<GradleBuildInvocationResult> sampleBuildOperation(String buildOperationDetailsClass) {
+        return new Sample<GradleBuildInvocationResult>() {
+            @Override
+            public String getName() {
+                return buildOperationDetailsClass;
+            }
+
+            @Override
+            public Duration extractFrom(GradleBuildInvocationResult result) {
+                return Optional.ofNullable(result.cumulativeBuildOperationTimes.get(buildOperationDetailsClass)).orElse(Duration.ofMillis(0));
```
lptr: It threw me off a bit that the method should return a `Duration` yet from the start of the `return` statement I expected it to return an `Optional`. If we keep the use of `Optional` here, I think it's important to break `orElse()` to a new line.

OTOH, I don't think this is a good use of `Optional`: we create an object just to be thrown away immediately, and I don't think it's easier to read than this:

```java
Duration duration = result.cumulativeBuildOperationTimes.get(buildOperationDetailsClass);
return duration != null
    ? duration
    : Duration.ofMillis(0);
```

Furthermore, `Duration.ofMillis(0)` should be replaced with `Duration.ZERO`.

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
```diff
@@ -1,31 +1,71 @@
...
 
-    public BuildOperationInstrumentation() throws IOException {
-        dataFile = File.createTempFile("gradle-profiler", "build-ops");
-        dataFile.deleteOnExit();
+    public BuildOperationInstrumentation(boolean measureConfigTime, ImmutableList<String> measuredBuildOperations) throws IOException {
+        this.measureConfigTime = measureConfigTime;
+        this.configurationTimeDataFile = File.createTempFile("gradle-profiler", "build-ops-config-time");
+        this.configurationTimeDataFile.deleteOnExit();
+        this.buildOperationDataFiles = measuredBuildOperations.stream()
+            .collect(ImmutableMap.toImmutableMap(Function.identity(), BuildOperationInstrumentation::createOpTmpFile));
```
lptr: Does it make sense to put each operation type into its own file? What is the performance impact of this?

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
```diff
@@ -1,31 +1,71 @@
...
+    }
+
+    public boolean requiresInitScript() {
+        return measureConfigTime || !buildOperationDataFiles.isEmpty();
+    }
+
+    private static File createOpTmpFile(String op) {
```
lptr: Nit: I don't feel very strongly, but it seems unnecessary to abbreviate this method name twice.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -46,6 +57,51 @@ public void started(BuildOperationDescriptor buildOperationDescriptor, Operation
...
+
+    private static class BuildOperationDurationRecordingListener extends DetachingBuildOperationListener {
+
+        private final Class<?> capturedBuildOperation;
+
+        private AsyncWriter<Long> writer;
+
+        BuildOperationDurationRecordingListener(String capturedBuildOperation, File outFile, BuildOperationListenerManager manager) throws ClassNotFoundException {
+            super(outFile, manager);
+            this.capturedBuildOperation = Class.forName(capturedBuildOperation + "$Details");
```
lptr: Ugh...

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -46,6 +57,51 @@ public void started(BuildOperationDescriptor buildOperationDescriptor, Operation
...
+        }
+
+        @Override
+        public void finished(BuildOperationDescriptor buildOperationDescriptor, OperationFinishEvent operationFinishEvent) {
+            Object details = buildOperationDescriptor.getDetails();
+            if (buildOperationDescriptor.getDetails() != null && capturedBuildOperation.isAssignableFrom(details.getClass())) {
+                AsyncWriter<Long> writer = ensureWriter();
+                writer.append(operationFinishEvent.getEndTime() - operationFinishEvent.getStartTime());
```
lptr: I'd expect us to collect the times in-memory and only write the end-result to a file. If we did that we could write the results to the same file in a map format instead of needed a separate file for each measured build operation.

What is the reason with writing each event to a file?

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -46,6 +57,51 @@ public void started(BuildOperationDescriptor buildOperationDescriptor, Operation
                 writer.finished();
             }
         }
+    }
+
+    private static class BuildOperationDurationRecordingListener extends DetachingBuildOperationListener {
```
lptr: > `extends DetachingBuildOperationListener`

Doesn't this mean that the listener detaches before the first task would ever execute?

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
paplorinc: If allocation is important here, we can check whether escape analysis kick in here - otherwise I would vote for the newline.

oehme:
I feel like this should be a generalization of the chrome-trace profiler (maybe rename it to "operation trace"?), where you can specify which operations you want aggregates for. By making this (and possibly the "configuration time breakdown") part of the chrome-trace profiler, we'd also remove some duplicate code. The more such operations we instrument, the more I'd consider this a profiler, not a benchmark.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
wolfs: The listener detaches after all the tasks finished executing (in finished).

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
wolfs: Would you prefer that the `$Details` is passed in?

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
wolfs: I did it like this since it seemed more straight forward. I can also capture the (sum of) the measurements in memory and then write the result to a file.

If I want to write all results to the same file, then I need to either have
- one listener
- synchronize the listeners

I don't see much benefit in that. What would be the advantage of writing all the results into one file?

build.gradle
```diff
@@ -41,7 +41,9 @@ dependencies {
     testRuntime 'org.objenesis:objenesis:2.6'
 }
 
-sourceCompatibility = 1.8
+allprojects {
+    sourceCompatibility = 1.8
```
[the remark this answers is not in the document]
wolfs: We want gradle profiler to be able to measure builds running on Java 8. Therefore at least the subprojects need to be compiled with Java 8 bytecode, since they run as init scripts on the build JVM. It seems straightforward to also make it possible to run Gradle profiler itself on Java 8, so you don't need to have two JDKs installed.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
lptr: Not sure. It would be great to have some typed access to these things is all. Nothing to change here I guess.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
lptr: You don't have to synchronize on the writes; you could simply collect the results in some concurrent structure and write them out at the end of the build instead.

I don't much care about the number of files we end up with (though a single text file would be more tidy and easier to process for a human). What I care about is that we should keep writing to files during the benchmark to a minimum.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
lptr: Ah, right, missed that.

build.gradle
```diff
@@ -41,9 +41,7 @@ dependencies {
     testRuntime 'org.objenesis:objenesis:2.6'
 }
 
-allprojects {
-    sourceCompatibility = 1.8
-}
+sourceCompatibility = 1.8
```
lptr: Being explicit about target compatibility would make this easier to comprehend for me.

wolfs: Done.

src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -39,6 +43,8 @@ public InvocationSettings parseSettings(String[] args) throws IOException, Setti
             .defaultsTo("jfr");
         Profiler.configureParser(parser);
         OptionSpecBuilder benchmarkOption = parser.accepts("benchmark", "Collect benchmark metrics");
+        ArgumentAcceptingOptionSpec<String> benchmarkBuildOperation = parser.accepts("benchmark-build-op", "Collect benchmark metrics for build operation")
+            .withRequiredArg();
```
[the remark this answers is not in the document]
wolfs: Done.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
wolfs: I replaced the `get` call with `getOrDefault`.

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
wolfs: Renamed.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
wolfs: I am now only writing the cumulative time into the files. I think having one file per build operation we track is not too much overhead.

wolfs:
@lptr PTAL!

lptr:
Are we not going to rename `--benchmark-config-time`?

wolfs:
This PR requires https://github.com/gradle/gradle/pull/10068 so that the capturing of configuration time works together with instant execution.

PR #141, opened 2019-07-23
Upgradle to 5.5.1

wolfs:


src/test/groovy/org/gradle/profiler/ChromeTraceIntegrationTest.groovy
```diff
@@ -17,7 +17,7 @@ class ChromeTraceIntegrationTest extends AbstractProfilerIntegrationTest {
         new File(outputDir, "${versionUnderTest}-trace.html").isFile()
 
         where:
-        versionUnderTest << supportedGradleVersions
+        versionUnderTest << ["4.7"]
```
lptr: Why not test with all versions as before?

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
```diff
@@ -61,8 +61,16 @@ private long getStartTime() {
             try {
                 return buildRequestMetaData.getStartTime();
             } catch (NoSuchMethodError e) {
-                return buildRequestMetaData.getBuildTimeClock().getStartTime();
+                return (long) call(call(buildRequestMetaData, "getBuildTimeClock"), "getStartTime");
             }
         }
     }
+
+    private Object call(Object object, String method) {
```
lptr: Make this `static` and rename it to `invokeGetter`?

src/test/groovy/org/gradle/profiler/ChromeTraceIntegrationTest.groovy
wolfs: Good catch, I changed this when testing out something locally. Reverted to testing all versions again.

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
wolfs: I moved the method to a utility class.

wolfs:
@lptr PTAL!

lptr:
LGTM.

PR #143, opened 2019-07-29
Publish distributions and JARs

### LINKED ISSUE #8: Setup release infrastructure

wolfs:
Fixes #8.


build.gradle.kts
```diff
@@ -65,7 +65,7 @@ tasks.processResources {
 
 allprojects {
     tasks.withType<Jar>().configureEach {
-        archiveVersion.set(null as String?)
+        archiveFileName.set(provider { "${this@configureEach.archiveBaseName.get()}.jar" })
```
lptr: Wow, this is some weird notation. Can we do this in some other way that makes it easier to understand?

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
```diff
@@ -20,15 +19,15 @@
...
+    private final Map<String, File> buildOperationDataFiles;
     private final boolean measureConfigTime;
 
     public BuildOperationInstrumentation(boolean measureConfigTime, List<String> measuredBuildOperations) throws IOException {
         this.measureConfigTime = measureConfigTime;
         this.configurationTimeDataFile = File.createTempFile("gradle-profiler", "build-ops-config-time");
         this.configurationTimeDataFile.deleteOnExit();
         this.buildOperationDataFiles = measuredBuildOperations.stream()
-            .collect(ImmutableMap.toImmutableMap(Function.identity(), BuildOperationInstrumentation::createBuildOperationTempFile));
+            .collect(Collectors.toMap(Function.identity(), BuildOperationInstrumentation::createBuildOperationTempFile));
```
lptr: You can use `ImmutableMap.toImmutableMap()` here to keep this as an `ImmutableMap`.

.teamcity/settings/GradleProfilerPublishing.kt
```diff
@@ -13,12 +13,12 @@ object GradleProfilerPublishing : BuildType({
     params {
         java8Home(Os.linux)
         text("ARTIFACTORY_USERNAME", "bot-build-tool", allowEmpty = true)
-        password("ARTIFACTORY_PASSWORD", "credentialsJSON:2b7529cd-77cd-49f4-9416-9461f6ac9018", display = ParameterDisplay.HIDDEN)
+        password("ARTIFACTORY_PASSWORD", "credentialsJSON:d94612fb-3291-41f5-b043-e2b3994aeeb4", display = ParameterDisplay.HIDDEN)
```
lptr: Why do we have credentials baked in here?

oehme: This is just a pointer to credentials in teamcity, i.e. they still need to be configured there and an attacker can't do anything with this pointer.

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
wolfs: I can't, since on the `gradle/gradle` side we use the android variant of guava, which does not contain the method. That is why I switched to using a regular map instead of an immutable one.

build.gradle.kts
wolfs: What weird notation are you referring to? `this@configureEach`?

I can assign `this` to a `val` and then use the `val` in the inner lambda.

lptr: There are several levels of indirection to unpack here. I think separating some of it out would be a good idea. Introducing a `val` for `this` could work. Though can't you simply refer to `${archiveBaseName.get()}`?

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
lptr: Right. And we need to access the profiler in an embedded way?

wolfs: Yeah, I want to use the Java API instead of spawning yet another process and parsing the CSV result.

build.gradle.kts
wolfs: I simplified it to
```
archiveFileName.set(archiveBaseName.map { "$it.jar" })
```

lptr: Looks much cleaner, thanks.

lptr:
LGTM.

build.gradle.kts
```diff
@@ -65,6 +65,8 @@ tasks.processResources {
 
 allprojects {
     tasks.withType<Jar>().configureEach {
+        // Removing the version from the JARs here, since JARs in `profilerPlugins` are referenced by name in production code.
```
oehme: I'd probably do that renaming on the consumer side. The producing project really shouldn't care about how it is consumed.

Alternatively, the consumer can have its version number in a manifest/properties file and use that to look up the file.

wolfs: I am renaming now on the consumer side.

PR #144, opened 2019-07-31
Make it possible to use API programmatically

wolfs:
The changes from this PR are required for https://github.com/gradle/gradle/pull/10135.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ You can use `--measure-build-op` together with the fully qualified class name of
 For example, for Gradle 5.x there is a [`org.gradle.api.internal.tasks.SnapshotTaskInputsBuildOperationType`](https://github.com/gradle/gradle/blob/c671360a3f1729b406c5b8b5b0d22c7b81296993/subprojects/core/src/main/java/org/gradle/api/internal/tasks/SnapshotTaskInputsBuildOperationType.java) which can be used to capture snapshotting time.
 The time recorded is cumulative time, so the wall clock time spent on executing the measured build operations is probably smaller.
 
+### Regression detection
+
+If multiple versions are tested, then Gradle profiler determines whether there is an actual difference in the run times by using a [Mann-Whitney U-Test](https://de.wikipedia.org/wiki/Wilcoxon-Mann-Whitney-Test).
+The result files contain the confidence of that the two samples actually have a different performance behavior - i.e. that one is faster or slower than the other.
+
 ## Profiling a build
 
 Profiling allows you to get deeper insight into the performance of your build.
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ The time recorded is cumulative time, so the wall clock time spent on executing 
 
 ### Regression detection
 
-If multiple versions are tested, then Gradle profiler determines whether there is an actual difference in the run times by using a [Mann-Whitney U-Test](https://de.wikipedia.org/wiki/Wilcoxon-Mann-Whitney-Test).
-The result files contain the confidence of that the two samples actually have a different performance behavior - i.e. that one is faster or slower than the other.
+If multiple versions are tested, then Gradle profiler determines whether there is an statistically significant difference in the run times by using a [Mann-Whitney U-Test](https://de.wikipedia.org/wiki/Wilcoxon-Mann-Whitney-Test).
+The result files contain the confidence if a sample has a different performance behavior - i.e. it is faster or slower - than the baseline.
 
 ## Profiling a build
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,7 +27,7 @@ val profilerPlugins by configurations.creating
 dependencies {
     implementation("org.gradle:gradle-tooling-api:5.2.1")
     implementation("com.google.code.findbugs:annotations:3.0.1")
-    implementation("com.google.guava:guava:27.1-jre")
+    implementation("com.google.guava:guava:27.1-android")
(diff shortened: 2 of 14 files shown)


build.gradle.kts
```diff
@@ -27,7 +27,7 @@ val profilerPlugins by configurations.creating
 dependencies {
     implementation("org.gradle:gradle-tooling-api:5.2.1")
     implementation("com.google.code.findbugs:annotations:3.0.1")
-    implementation("com.google.guava:guava:27.1-jre")
+    implementation("com.google.guava:guava:27.1-android")
```
lptr: I think we should document the reason here.

src/main/java/org/gradle/profiler/BenchmarkResultCollector.java
```diff
@@ -135,7 +139,7 @@ private double getPValue(int sample) {
             if (a.length == 0 || b.length == 0) {
                 return 1;
             }
-            return new MannWhitneyUTest().mannWhitneyUTest(a, b);
+            return (1 - new MannWhitneyUTest().mannWhitneyUTest(a, b)) * 100;
```
lptr: Are we doing this to match how we calculate it in `gradle/gradle`?

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
```diff
@@ -66,7 +65,8 @@ private String newFile(File dataFile) {
         }
         try {
             try (Stream<String> lines = Files.lines(configurationTimeDataFile.toPath(), StandardCharsets.UTF_8)) {
-                return Streams.findLast(lines)
+                return lines
+                    .reduce((first, second) -> second)
```
lptr: :( I hate that we have to make so much code worse because of our inability to upgrade Guava to the JRE version.

src/main/java/org/gradle/profiler/report/CsvGenerator.java
```diff
@@ -93,6 +93,7 @@ protected void write(BenchmarkResult benchmarkResult, BufferedWriter writer) thr
         statistic(writer, "75th percentile", statistics, v -> v.getPercentile(75));
         statistic(writer, "max", statistics, BuildScenarioResult.Statistics::getMax);
         statistic(writer, "stddev", statistics, BuildScenarioResult.Statistics::getStandardDeviation);
+        statistic(writer, "p-value (Mann Whitney U test)", statistics, BuildScenarioResult.Statistics::getPValue);
```
lptr: Can we mention somehow that this represents the confidence?

I would probably only call it "confidence" (or "p-value") and document somewhere else what it means. Having this complicated description breaks up the layout of the CSV quite a bit.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1305,7 +1305,7 @@ buildGoal {
         logFile.containsOne("* Maven targets: [-v]")
 
         def lines = resultFile.lines
-        lines.size() == 27 // 4 headers, 16 executions, 7 stats
+        lines.size() == 28 // 4 headers, 16 executions, 8 stats
```
lptr: These tests look like way too much effort to maintain. If we want to keep them I think we should use the sum of named variables instead of comments.

src/main/java/org/gradle/profiler/BenchmarkResultCollector.java
wolfs: Yes, that gives the confidence which we show in the reports.

build.gradle.kts
wolfs: Done.

src/main/java/org/gradle/profiler/report/CsvGenerator.java
wolfs: I am calling it `confidence` now.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
wolfs: I extracted a method which gives the total number of lines depending on the number of executions.

wolfs:
@lptr I addressed your comments, please have another look!

README.md
```diff
@@ -47,6 +47,11 @@ You can use `--measure-build-op` together with the fully qualified class name of
 For example, for Gradle 5.x there is a [`org.gradle.api.internal.tasks.SnapshotTaskInputsBuildOperationType`](https://github.com/gradle/gradle/blob/c671360a3f1729b406c5b8b5b0d22c7b81296993/subprojects/core/src/main/java/org/gradle/api/internal/tasks/SnapshotTaskInputsBuildOperationType.java) which can be used to capture snapshotting time.
 The time recorded is cumulative time, so the wall clock time spent on executing the measured build operations is probably smaller.
 
+### Regression detection
+
+If multiple versions are tested, then Gradle profiler determines whether there is an actual difference in the run times by using a [Mann-Whitney U-Test](https://de.wikipedia.org/wiki/Wilcoxon-Mann-Whitney-Test).
+The result files contain the confidence of that the two samples actually have a different performance behavior - i.e. that one is faster or slower than the other.
+
```
lptr: You are talking about both multiple versions here and two versions. What happens if we have more than two versions under test?

lptr:
I think this looks good. Let's merge it.

PR #145, opened 2019-08-02
Fix failure on Windows when file paths have spaces in them

lptr:
We don't need to run Gradle using cmd.exe. Doing so messes up the escaping of spaces in the file names passed to it as parameters and results in the following type of failure:

```text
* Stopping daemons
java.lang.RuntimeException: Could not run command cmd.exe /C C:\Users\Lorant Pinter\Workspace\gradle\local-install\bin\gradle --stop --gradle-user-home C:\Users\Lorant Pinter\Workspace\gradle\snapshotting-performance\santa-tracker-android\gradle-user-home
Output:
======
'C:\Users\Lorant' is not recognized as an internal or external command,
operable program or batch file.
======
        at org.gradle.profiler.CommandExec$RunHandle.waitForSuccess(CommandExec.java:178)
        at org.gradle.profiler.CommandExec.run(CommandExec.java:68)
        at org.gradle.profiler.CommandExec.run(CommandExec.java:34)
        at org.gradle.profiler.CommandExec.run(CommandExec.java:29)
        at org.gradle.profiler.GradleBuildConfiguration.runGradle(GradleBuildConfiguration.java:59)
```

Instead we can invoke `gradle.bat` directly that has none of this problem.

### DIFF
--- a/src/main/java/org/gradle/profiler/GradleBuildConfiguration.java
+++ b/src/main/java/org/gradle/profiler/GradleBuildConfiguration.java
@@ -60,10 +60,9 @@ public class GradleBuildConfiguration {
     }
 
     public void addGradleCommand(List<String> commandLine) {
-        if (OperatingSystem.isWindows()) {
-            commandLine.add("cmd.exe");
-            commandLine.add("/C");
-        }
-        commandLine.add(new File(gradleHome, "bin/gradle").getAbsolutePath());
+        String gradleScriptName = OperatingSystem.isWindows()
+            ? "gradle.bat"
+            : "gradle";
+        commandLine.add(new File(gradleHome, "bin/" + gradleScriptName).getAbsolutePath());
     }
 }
--- a/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
@@ -1048,7 +1048,7 @@ println "User home: \$gradle.gradleUserHomeDir"
         logFile.find("User home: " + new File("gradle-user-home").absolutePath)
     }
 
-    def "Can specify user home"() {
+    def "can specify user home"() {
         given:
         buildFile.text = """
 apply plugin: 'base'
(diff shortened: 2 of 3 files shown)


src/main/java/org/gradle/profiler/GradleBuildConfiguration.java
```diff
@@ -60,10 +60,9 @@ public void runGradle(String... arguments) {
...
     public void addGradleCommand(List<String> commandLine) {
-        if (OperatingSystem.isWindows()) {
-            commandLine.add("cmd.exe");
-            commandLine.add("/C");
-        }
-        commandLine.add(new File(gradleHome, "bin/gradle").getAbsolutePath());
+        String gradleScriptName = OperatingSystem.isWindows()
+            ? "gradle.bat"
+            : "gradle";
+        commandLine.add(new File(gradleHome, "bin/" + gradleScriptName).getAbsolutePath());
```
wolfs: How about adding a test where the project directory contains a space?

lptr: That wouldn't cause a problem. We only have problems here when either the Gradle installation or the Gradle home is in a spaced location.

We don't have a test for a locally installed Gradle version, but I made the test specifying a Gradle home to use one with a space in the name.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1057,10 +1057,10 @@ println "User home: \$gradle.gradleUserHomeDir"
 
         when:
         new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--gradle-version", minimalSupportedGradleVersion,
-            "--benchmark", "--gradle-user-home", "foobar", "help")
+            "--benchmark", "--gradle-user-home", "home with spaces", "help")
```
wolfs: The test is not specifying a Gradle home but a Gradle _user_ home. IIUC, that does not cause problems. I didn't find any test which uses a not auto-installed Gradle version :(

src/main/java/org/gradle/profiler/GradleBuildConfiguration.java
wolfs: Understood.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
lptr: Sorry, I meant Gradle user home, and it does cause a problem (I checked). I think we're good.

PR #147, opened 2019-08-05
Only publish the root project

wolfs:
The chrome-trace subproject did contain already a publication added
by the `java-gradle-plugin`. Adding a second publication caused
everything to be published twice which caused a build failure when
releasing.

Since the subprojects are not meant to be consumed, we shouldn't publish them at all.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -85,31 +85,14 @@ publishing {
     }
 }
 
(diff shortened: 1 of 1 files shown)


wolfs:
@lptr PTAL, I am now only publishing the root project.

PR #148, opened 2019-08-05
Publish build scans when running on CI

wolfs:


.teamcity/settings/extensions.kt
```diff
@@ -19,3 +19,5 @@ fun BuildType.gradleProfilerVcs() {
         checkoutMode = CheckoutMode.ON_SERVER
     }
 }
+
+val useBuildScanInitScript = "-I gradle/init-scripts/build-scan.init.gradle.kts"
```
lptr: This is not really about using build scans, but about configuring our own server, no? I mean, the `build-scan` plugin would be applied anyway from `build.gradle.kts`, no?

wolfs: Do you suggest to rename the `build-scan.init.gradle.kts` script? Or is this more about renaming `useBuildScanInitScript`? I think the variable `useBuildScanInitScript` is fine, since the command line argument instructs Gradle to use the `build-scan` init script.

lptr: How about something like `useGradleInternalScansServer` and `build-scan-gradle-internal-server.init.gradle.kts`?

wolfs: I renamed the variable and the script.

PR #150, opened 2019-08-12
Add clear project caches and Gradle user home mutators

wolfs:
So it is possible to define the first use performance test in Gradle profiler.

### DIFF
--- a/src/main/java/org/gradle/profiler/BuildMutatorFactory.java
+++ b/src/main/java/org/gradle/profiler/BuildMutatorFactory.java
@@ -17,13 +17,14 @@ public class BuildMutatorFactory implements Supplier<BuildMutator> {
             .map(Supplier::get)
             .filter(it -> it != BuildMutator.NOOP)
             .collect(Collectors.toList());
-        if (mutators.isEmpty()) {
-            return BuildMutator.NOOP;
+        switch (mutators.size()) {
+            case 0:
+                return BuildMutator.NOOP;
+            case 1:
+                return mutators.get(0);
+            default:
+                return new CompositeBuildMutator(mutators);
         }
-        if (mutators.size() == 1) {
-            return mutators.get(0);
-        }
-        return new CompositeBuildMutator(mutators);
     }
 
     @Override
--- a/src/main/java/org/gradle/profiler/BuildMutatorFactory.java
+++ b/src/main/java/org/gradle/profiler/BuildMutatorFactory.java
@@ -13,13 +13,16 @@ public class BuildMutatorFactory implements Supplier<BuildMutator> {
 
     @Override
     public BuildMutator get() {
-        if (factories.isEmpty()) {
-            return new NoOpMutator();
+        List<BuildMutator> mutators = factories.stream()
+            .map(Supplier::get)
+            .filter(it -> it != BuildMutator.NOOP)
+            .collect(Collectors.toList());
+        if (mutators.isEmpty()) {
+            return BuildMutator.NOOP;
         }
-        if (factories.size() == 1) {
-            return factories.get(0).get();
+        if (mutators.size() == 1) {
+            return mutators.get(0);
         }
-        List<BuildMutator> mutators = factories.stream().map(Supplier::get).collect(Collectors.toList());
         return new CompositeBuildMutator(mutators);
     }
 
@@ -28,12 +31,4 @@ public class BuildMutatorFactory implements Supplier<BuildMutator> {
     {
         return get().toString();
     }
-
-    private static class NoOpMutator implements BuildMutator {
-        @Override
-        public String toString() {
-            return "none";
-        }
-    }
-
 }
--- a/src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
@@ -10,7 +10,7 @@ import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.util.function.Supplier;
 
-public class ClearGradleUserHomeMutator implements BuildMutator {
+public class ClearGradleUserHomeMutator extends AbstractBuildMutator {
     private final File gradleUserHome;
 
     public ClearGradleUserHomeMutator(File gradleUserHome) {
--- a/src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
@@ -1,0 +1,59 @@
+package org.gradle.profiler.mutations;
+
+import com.typesafe.config.Config;
+import org.apache.commons.io.FileUtils;
+import org.gradle.profiler.BuildMutator;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.util.function.Supplier;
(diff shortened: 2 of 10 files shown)


src/main/java/org/gradle/profiler/BuildMutatorFactory.java
```diff
@@ -13,13 +13,16 @@ public BuildMutatorFactory(List<? extends Supplier<BuildMutator>> factories) {
...
+        if (mutators.isEmpty()) {
+            return BuildMutator.NOOP;
         }
-        if (factories.size() == 1) {
-            return factories.get(0).get();
+        if (mutators.size() == 1) {
+            return mutators.get(0);
         }
-        List<BuildMutator> mutators = factories.stream().map(Supplier::get).collect(Collectors.toList());
         return new CompositeBuildMutator(mutators);
```
lptr: Nit: you could switch on the size of `mutators`.

src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
```diff
@@ -0,0 +1,59 @@
...
+        if (!gradleUserHome.exists()) {
+            throw new IllegalArgumentException(String.format(
+                "Cannot delete Gradle user home directory (%s) since it does not exist",
+                gradleUserHome
+            ));
+        }
+        try {
+            Files.list(gradleUserHome.toPath())
+                // Don't delete the wrapper dir, since this is where the Gradle distribution we are going to run is located
+                .filter(it -> !it.getFileName().toString().equals("wrapper"))
```
lptr: Retaining the wrapper dir seems like something to be mentioned in the docs.

src/main/java/org/gradle/profiler/mutations/ClearProjectCacheMutator.java
```diff
@@ -0,0 +1,57 @@
...
+    public String toString() {
+        return getClass().getSimpleName();
+    }
+
+    public static class Configurator implements BuildMutatorConfigurator {
+        @Override
+        public Supplier<BuildMutator> configure(Config scenario, String scenarioName, File projectDir, String key) {
+            boolean enabled = scenario.getBoolean(key);
+            return () -> enabled ? new ClearProjectCacheMutator(projectDir) : BuildMutator.NOOP;
+        }
```
lptr: Can we handle the `enabled` state in a more general manner, so we don't have to duplicate this code?

src/main/java/org/gradle/profiler/mutations/ClearProjectCacheMutator.java
```diff
@@ -0,0 +1,57 @@
...
+            FileUtils.deleteDirectory(gradleCache);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName();
+    }
```
lptr: This should be consistent, but the other mutator doesn't add it. Shall we make it the default?

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1502,6 +1503,80 @@ buildTarget {
...
+buildTarget {
+    versions = ["${latestSupportedGradleVersion}"]
+    clear-gradle-user-home = true
+    daemon = none
+    tasks = ["compileJava"]
+}
+"""
+
+        when:
+        new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--benchmark", "--scenario-file", scenarios.absolutePath, "buildTarget", "--warmups", "1", "--iterations", "1")
```
lptr: Please break this up into multiple lines to keep it readable.

src/main/java/org/gradle/profiler/mutations/ClearProjectCacheMutator.java
wolfs: I was wondering if we need `enabled` at all. Something like this in the scenario file should be enough:
```
default {
    clear-project-caches
}
```

Same for the build cache logging mutator. WDYT?

lptr: That could work, too. The only use of `enabled = false` would be if we had some elaborate inheritance scheme where it would make sense to enable things on top and disable them below.

Another angle to look at is if we want the configuration to be more declarative or imperative. I'd say what you are proposing takes us towards imperative.

I don't feel strongly either way.

wolfs: I don't see leaving out the boolean value of the property as moving towards less declarative. The declaration `clear-project-caches` means to add the given mutator which requires no configuration options. The other mutators don't have enable/disable options either (e.g. `apply-abi-change`).

wolfs: Looks like you can't use
```
default {
    clear-project-caches
}
```

in HOCON, so I introduced a base class for the mutators without options.

src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
wolfs: Mentioned now.

src/main/java/org/gradle/profiler/BuildMutatorFactory.java
wolfs: Done.

src/main/java/org/gradle/profiler/mutations/ClearProjectCacheMutator.java
wolfs: I added a new base class which makes it the default. It can't be a default method on the interface, since the method is defined in `Object`...

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
wolfs: Done. I also extracted a method to run the benchmark.

wolfs:
@lptr PTAL!

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1676,7 +1676,8 @@ buildTarget {
 """
 
         when:
-        new Main().run("--project-dir", repoDir.absolutePath, "--output-dir", outputDir.absolutePath, "--benchmark", "--scenario-file", scenarios.absolutePath, "--gradle-version", minimalSupportedGradleVersion, "buildTarget", "--warmups", "1", "--iterations", "1")
+        benchmarkScenario(repoDir, scenarios)
+        new Main().run("--project-dir", repoDir.absolutePath, "--output-dir", outputDir.absolutePath, "--benchmark", "--scenario-file", scenarios.absolutePath, "--gradle-version", latestSupportedGradleVersion, "buildTarget", "--warmups", "1", "--iterations", "1")
```
lptr: Looks like this is now redundant?

lptr:
LGTM apart from that one change.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
wolfs: Good catch! Fixed!

Issue #132, opened 2019-05-23
Provide option to disable generating debug sources in sync action


lukaville:
In the current Android Studio version, there is an option that disables source generation during sync. It will be great if we will be able to disable it in the profiler action.

Issue #128, opened 2019-05-08
Android Sync Does Not Work on Projects with Flavors


AndrewReitz:
Looks like this feature requires the application to have `generateDebugSources` which if you add flavors to your application this task would be something like `generateFlavorDebugSources`, and there would be multiple of these which I think should all be run in order to get a correct sync profile report.

PR #155, opened 2019-09-21
Support build variants and source generation avoidance in AS sync

rom4ek:
This PR addresses both #128 and #132. I didn't come up with anything better than adding a new config in the scenario for `android-studio-sync`. Hope this works well for you.

### DIFF
--- a/src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
+++ b/src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
@@ -1,6 +1,8 @@
 package org.gradle.profiler;
 
 import com.android.builder.model.AndroidProject;
+
+import org.gradle.internal.impldep.org.apache.commons.lang.StringUtils;
 import org.gradle.tooling.BuildController;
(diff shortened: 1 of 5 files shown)


src/main/java/org/gradle/profiler/AndroidStudioSyncAction.java
```diff
@@ -35,10 +46,15 @@ public void run(GradleInvoker buildInvoker, List<String> gradleArgs, List<String
...
-        List<String> tasks = Collections.singletonList("generateDebugSources");
-        buildInvoker.runToolingAction(tasks, gradleArgs, jvmArgs, new GetModel(), (builder) -> {
-            builder.addProgressListener(noOpListener());
-        });
+
+        List<String> tasks;
+        if (skipSourceGeneration) {
+            tasks = Collections.emptyList();
+        } else {
+            String taskName = String.join("", "generate", StringUtils.capitalize(buildFlavor), "Sources");
```
oehme: Why the join call with an empty String? A simple concatenation would work :)

PR #167, opened 2019-11-20
Make unique text used by file mutators actually unique

### LINKED ISSUE #166: Unique text used by file mutators is not always unique
File mutators use a timestamp to generate unique text to ensure that the inputs they mutate don't accidentally match up with anything seen previously. This is to prevent caches to produce a hit when we don't want to.

Unfortunately, this timestamp is not always unique. It is generated when the mutators are created during scenario loading. If things happen fast enough, multiple mutators can be created with the same timestamp. In some cases this can lead to unintentional effects like one scenario storing some files in the build cache and the next scenario loading it.

lptr:
Fixes #166.

Instead of a timestamp to identify the scenario use a unique ID for the profiler invocation (using `UUID.randomUUID()`) and append the scenario name to it.

Instead of using a counter to distinguish between invocations append the phase name (`WARM_UP` or `MEASURE`) and the build number to the scenario ID to generate unique text for each build.

### DIFF
--- a/src/main/java/org/gradle/profiler/ScenarioContext.java
+++ b/src/main/java/org/gradle/profiler/ScenarioContext.java
@@ -16,7 +16,7 @@ public class ScenarioContext {
     }
 
     public String getUniqueScenarioId() {
-        return String.format("%s-%s", invocationId, scenarioName);
+        return String.format("_%s_%s", invocationId.toString().replaceAll("-", "_"), scenarioName);
     }
 
     public BuildContext withBuild(Phase phase, int count) {
--- a/src/main/java/org/gradle/profiler/ScenarioContext.java
+++ b/src/main/java/org/gradle/profiler/ScenarioContext.java
@@ -1,0 +1,25 @@
+package org.gradle.profiler;
+
+import java.util.UUID;
+
+public class ScenarioContext {
+    private final UUID invocationId;
+    private final String scenarioName;
+
+    public static ScenarioContext from(InvocationSettings invocationSettings, ScenarioDefinition scenarioDefinition) {
+        return new ScenarioContext(invocationSettings.getInvocationId(), scenarioDefinition.getName());
+    };
+
+    protected ScenarioContext(UUID invocationId, String scenarioName) {
+        this.invocationId = invocationId;
+        this.scenarioName = scenarioName;
+    }
+
+    public String getUniqueScenarioId() {
+        return String.format("%s-%s", invocationId, scenarioName);
+    }
+
+    public BuildContext withBuild(Phase phase, int count) {
+        return new BuildContext(invocationId, scenarioName, phase, count);
+    }
+}
--- a/src/main/java/org/gradle/profiler/ScenarioContext.java
+++ b/src/main/java/org/gradle/profiler/ScenarioContext.java
@@ -1,5 +1,7 @@
 package org.gradle.profiler;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import java.util.UUID;
 
 public class ScenarioContext {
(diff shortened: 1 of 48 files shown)


src/main/java/org/gradle/profiler/mutations/ApplyNonAbiChangeToJavaSourceFileMutator.java
```diff
@@ -15,9 +16,10 @@ public ApplyNonAbiChangeToJavaSourceFileMutator(File sourceFile) {
     }
 
     @Override
-    protected void applyChangeTo(CompilationUnit compilationUnit) {
+    protected void applyChangeTo(BuildContext context, CompilationUnit compilationUnit) {
         MethodDeclaration existingMethod = getExistingMethod(compilationUnit);
-        existingMethod.getBody().get().addStatement(0, JavaParser.parseStatement("System.out.println(\"" + getUniqueText() + "\");"));
+        existingMethod.getBody()
+            .ifPresent(body -> body.addStatement(0, JavaParser.parseStatement("System.out.println(\"" + context.getUniqueBuildId() + "\");")));
```
wolfs: I think it is better to fail here instead of doing no change to the body of the method. This way at least you know that the non-ABI change was not successfully applied.

src/main/java/org/gradle/profiler/ScenarioContext.java
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    @VisibleForTesting
+    public ScenarioContext(UUID invocationId, String scenarioName) {
+        this.invocationId = invocationId;
+        this.scenarioName = scenarioName;
+    }
+
+    public String getUniqueScenarioId() {
+        return String.format("_%s_%s", invocationId.toString().replaceAll("-", "_"), scenarioName);
```
wolfs: This id will be used as part of a Java identifier. I think it is dangerous to use the scenario name here as-is. I could run the same scenario twice in one Gradle profiler invocation. How about we use the index of the scenario here?

src/main/java/org/gradle/profiler/mutations/AbstractFileChangeMutator.java
```diff
@@ -57,11 +43,8 @@ private void revert() {
     }
 
     @Override
-    public void afterScenario() {
-        if (counter > 0) {
-            revert();
-            counter = 0;
-        }
+    public void afterScenario(ScenarioContext context) {
```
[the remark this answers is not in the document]
lptr: So you'd always have a single change? That could work, but let's do it in another PR. Open an issue so we can discuss it.

src/main/java/org/gradle/profiler/mutations/ApplyNonAbiChangeToJavaSourceFileMutator.java
lptr: Good point.

src/main/java/org/gradle/profiler/ScenarioContext.java
lptr: We didn't have a different "unique text" for multiple scenario executions before either. We could make this change, but I don't want to do it in this PR.

src/main/java/org/gradle/profiler/mutations/AbstractFileChangeMutator.java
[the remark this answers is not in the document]
wolfs: https://github.com/gradle/gradle-profiler/issues/172

src/main/java/org/gradle/profiler/ScenarioContext.java
wolfs: I am not worried about the text being unique, but about the text not being good for a method name. For example, in the `gradle/gradle` build we use the display name for the Gradle profiler scenario name: https://github.com/gradle/gradle/blob/14acd460ee399dd78fe3a6938754e5687203df90/subprojects/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/GradleProfilerBuildExperimentRunner.java#L168

This won't be possible any more, since it contains spaces and other characters and we won't be able to use it as part of a method name.

Issue #169, opened 2019-11-21
Build operation capture with instant execution enabled is broken


eskatos:
This prevents measuring time to first task and other build operations.

eskatos:
See https://github.com/gradle/gradle/issues/12488#issuecomment-602317113

PR #173, opened 2019-11-22
Add coverage for Gradle 6.0

eskatos:
This PR only touches test coverage and raises the latest supported Gradle version to 6.0.1.
It is a prerequisite for #169


eskatos:
Thanks @wolfs

PR #175, opened 2019-11-25
Add long format option for CSV output

### LINKED ISSUE #174: Option to produce one-dimensional (long) CSV as output
These are much easier to use in data processing (like creating pivot tables) than the currently used two-dimensional (wide) table output.

See definition of one/two dimensional data tables here: https://github.com/rgertenbach/Reshape#one-dimensional-long-tables

lptr:
Fixes #174.


lptr:
I didn't find any coverage actually testing the existing CSV format. I added an issue so that we can come back to this as part of our performance work next year: https://github.com/gradle/gradle-profiler/issues/176.

Issue #164, opened 2019-11-20
Add human readable title to scenarios


lptr:
These should be shown in reports.

PR #179, opened 2019-11-29
Add scenario title to use in reports

lptr:
Fixes https://github.com/gradle/gradle-profiler/issues/164

If a title is specified for the scenario, it will be used in reports instead of the name of the scenario.

```hocon
default {
    title = "Base scenario"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,10 +184,13 @@ Here is an example:
     
     # Scenarios are run in alphabetical order
     assemble {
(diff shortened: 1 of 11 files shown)


wolfs:
We should probably make sure that we set the title to the human readable format when using the new profiler version to run performance tests in `gradle/gradle`.

wolfs:
Can you add `title` to the documentation as well?

PR #79, opened 2018-08-17
Throw an error when the Chrome trace file fails to be moved.

liutikas:

PR #180, opened 2019-12-04
Make chrome-trace tmp file copy work correctly

liutikas:
File#moveTo only works when the two files are within the same file system. Some Linux distributions set /tmp/ to be mounted to a different file system, thus causing moveTo fail with:
Caused by: java.lang.RuntimeException: Failed to move the trace file from a temporary location (/tmp/gradle-trace9055990495551523261.html) to the final location (/ssd/ssd1/temp/gradle-profiler/profile-out-9/5.6.4-trace.html)
	at org.gradle.trace.AsynchronousTraceWriter.finish(AsynchronousTraceWriter.java:40)
	at org.gradle.trace.TraceResult.finalizeTraceFile(TraceResult.java:48)
	at org.gradle.trace.GradleTracingPlugin$TraceFinalizerAdapter.buildFinished(GradleTracingPlugin.java:57)
...

This change replaces File#moveTo with Apache FileUtils#copyFile that works correctly in such situations.

### DIFF
--- a/subprojects/chrome-trace/src/main/java/org/gradle/trace/AsynchronousTraceWriter.java
+++ b/subprojects/chrome-trace/src/main/java/org/gradle/trace/AsynchronousTraceWriter.java
@@ -1,27 +1,34 @@
 package org.gradle.trace;
 
-import org.gradle.api.invocation.Gradle;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.PrintWriter;
+import java.io.Reader;
 import org.gradle.api.logging.Logging;
 import org.gradle.internal.UncheckedException;
 import org.gradle.trace.stream.AsyncWriter;
 
-import java.io.*;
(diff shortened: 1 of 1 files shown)


liutikas:
This is a follow up to actually fix https://github.com/gradle/gradle-profiler/pull/79

liutikas:
> Instead of working around this, I think we should instead create the temp file right next to the final destination and then rename it once done. This is guaranteed to work, as both will be on the same filesystem.

Should we just write to the `traceFile` file directly? As far as I can tell the only reason we have a temp file is to prevent a write if generating a trace fails which sounds like a bug in the first place.

oehme:
That's a fair point, it's not like we're writing to some sensitive or contested location 👍

liutikas:
Moved to writing directly to the final destination.

I had to add if (File#exists()) { File#delete() } as we run tracer several times (warm-up).

PR #181, opened 2019-12-05
Use stable APIs to instrument build operations and time to first task

adammurdoch:
Use the metrics collector APIs added in Gradle 6.1 to implement `--measure-config-time` and `--measure-build-op`. These APIs work with instant execution and will be stable going forward. This should mean the instrumentation is more reliable and that we can remove the workarounds for gradle-profiler from Gradle itself.


lptr:
Does this mean I can run instant execution benchmarks with measured dependency resolution?

adammurdoch:
it should, yes

PR #197, opened 2020-03-16
Allow scenarios to use different commands than `build`

bmuskalla:


lptr:
This will break existing scenarios, right?

bmuskalla:
Exactly. Do you know any scenario files that use Bazel and would require updating?

lptr:
I don't know of anything we publish that'd need to be updated. This has never been updated to include Bazel: https://github.com/gradle/perf-android-large

bmuskalla:
That's something I might do at some point in the future though. So given nobody is using `gradle-profiler` using bazel scenarios right now, are we good with breaking people? Do we need to announce this in some way?

bmuskalla:
Totally forgot about this one. So it seems no one (except myself) is using the bazel support, so no harm done breaking the compatibility. Going to merge this after rebasing if nobody objects.

PR #198, opened 2020-03-24
Fix build operation tracking

lptr:
There was a bug in `gradle/gradle` about how the build services are closed for builds that have a `:buildSrc`. Now it's fixed and we need to follow up. See: https://github.com/gradle/gradle/issues/12488#issuecomment-602317113


lptr:
Looks like we don't need to do anything special. As long as the Gradle version used for profiling contains the fix we are good.

PR #206, opened 2020-05-27
Update the gradle-profiler for Gradle 6.5-rc-1 and the configuration cache

eskatos:
See the individual commits

### DIFF
--- a/subprojects/build-operations/src/main/java/org/gradle/trace/pid/PidCollector.java
+++ b/subprojects/build-operations/src/main/java/org/gradle/trace/pid/PidCollector.java
@@ -1,10 +1,8 @@
 package org.gradle.trace.pid;
 
-import org.gradle.BuildAdapter;
-import org.gradle.BuildResult;
 import org.gradle.api.internal.GradleInternal;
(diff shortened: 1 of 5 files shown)


subprojects/build-operations/src/main/java/org/gradle/trace/pid/PidCollector.java
```diff
@@ -1,25 +1,15 @@
...
-        AsyncWriter<Long> writer = new AsyncWriter<>(outFile, (l, w) -> w.print(l));
-        writer.append(processEnvironment.getPid());
-        writer.finished();
-        gradle.addBuildListener(new BuildAdapter(){
-            @Override
-            public void buildFinished(BuildResult result) {
-                writer.stop();
-            }
-        });
+        GFileUtils.writeFile(processEnvironment.getPid().toString(), outFile);
```
lptr, line 13: Looks nicer. Any idea why we had the complex solution originally?

eskatos: It was meant to avoid hiting the disk at configuration time as explained in the commit message

PR #208, opened 2020-06-16
Add support for specifying # of builds in a scenario definition

paragchandra:
I thought I would be able to specify the # of builds to measure, just like I can already specify # of warm-ups. Since it wasn't supported out of the box, I thought I'd add it in - but let me know if I'm missing something here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,8 @@ A scenario can define changes that should be applied to the source before each b
 - `git-checkout`: Checks out a specific commit for the build step, and a different one for the cleanup step.
 - `git-revert`: Reverts a given set of commits before the build and resets it afterward.
 - `jvm-args`: Sets or overrides the jvm arguments set by `org.gradle.jvmargs` in gradle.properties.
+- `warm-ups`: Number of warmups to perform before measurement
+- `iterations`: Number of builds to actually measure
 
 They can be added to a scenario file like this:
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -50,6 +50,7 @@
     private static final String BUCK = "buck";
     private static final String MAVEN = "maven";
     private static final String WARM_UP_COUNT = "warm-ups";
+    private static final String BUILD_COUNT = "builds";
```
wolfs, line 53:
```suggestion
    private static final String ITERATION_COUNT = "iterations";
```

The command line option is `--iterations`, so I would use the same name in the scenario file. I don't know why the command line option is `--warmups` while the scenario file option ended up being `warm-ups`.

wolfs:
Can you also add some documentation in the README?

src/main/java/org/gradle/profiler/ScenarioLoader.java
paragchandra: Fixed

README.md
```diff
@@ -254,6 +254,8 @@ A scenario can define changes that should be applied to the source before each b
 - `git-checkout`: Checks out a specific commit for the build step, and a different one for the cleanup step.
 - `git-revert`: Reverts a given set of commits before the build and resets it afterward.
 - `jvm-args`: Sets or overrides the jvm arguments set by `org.gradle.jvmargs` in gradle.properties.
+- `warm-ups`: Number of warmups to perform before measurement
+- `iterations`: Number of builds to actually measure
```
lptr, line 258: These should be in alphabetic order. (We should move `show-build-cache-size` to the right location, too.)

src/test/groovy/org/gradle/profiler/ScenarioLoaderTest.groovy
```diff
@@ -248,6 +249,39 @@ class ScenarioLoaderTest extends Specification {
...
+        def benchmarkScenario = benchmarkScenarios[0] as GradleScenarioDefinition
+        benchmarkScenario.warmUpCount == 5
+        benchmarkScenario.buildCount == 2
+
+        def profileScenario = profileScenarios[0] as GradleScenarioDefinition
+        profileScenario.warmUpCount == 5
+        profileScenario.buildCount == 2
+
+        where:
+        runUsing      | daemon | warmups | profileWarmups
```
lptr, line 277: I don't see these parameters referenced, but they should be, right?

paragchandra: Removed `warmups` and `profileWarmups`

paragchandra:
Are you guys good with this? I think I've addressed all of your comments...

lptr:
Thanks @paragchandra!

Issue #212, opened 2020-07-03
Applying non abi change to Kotlin source changes Java ABI


gavra0:
While this is a limitation of compilation avoidance, currently with `apply-non-abi-change-to` in the scenario file, Java ABI will be changed because `KotlinMetadata` annotation will change. This means that the change is not a non-ABI in a typical sense.

If  possible, adding a dummy method and modifying its body would be great.

porcupine7:
Thanks for filing this bug. I wondered why there was no difference between `apply-abi-change-to` and  `apply-non-abi-change-to` in my kotlin module. This seems to be the reason.

lptr:
@eskatos:

> I had a quick look, should be fairly simple, `beforeScenario()` would insert the dummy function, and `applyChangeTo()` would change the function body.

eskatos:
Fixed by https://github.com/gradle/gradle-profiler/pull/215

PR #215, opened 2020-07-17
Fix Kotlin non-abi-change mutator so it produces a non-abi-change for java consumers

eskatos:
Changing the signature of a private function changes kotlin metadata hence not producing a non-abi-change but an abi-change for java consumers.

This commit lets the Kotlin mutator inject a dummy method before running a scenario, and modify it's implementation body on each build.

See https://github.com/gradle/gradle-profiler/issues/212

### DIFF
--- a/src/test/groovy/org/gradle/profiler/mutations/ApplyNonAbiChangeToKotlinSourceFileMutatorTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/mutations/ApplyNonAbiChangeToKotlinSourceFileMutatorTest.groovy
@@ -1,17 +1,32 @@
 package org.gradle.profiler.mutations
 
+import org.gradle.profiler.Phase
+
 class ApplyNonAbiChangeToKotlinSourceFileMutatorTest extends AbstractMutatorTest {
 
-    def "adds and replaces public method at end of source file"() {
+    def "adds public function at end of source file replacing its body"() {
         def sourceFile = tmpDir.newFile("Thing.kt")
         sourceFile.text = "class Thing { fun existingMethod() { }}"
         def mutator = new ApplyNonAbiChangeToKotlinSourceFileMutator(sourceFile)
+        def functionText = { qualifier ->
+            "class Thing { fun existingMethod() { }}\n" +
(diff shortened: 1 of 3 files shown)


src/main/java/org/gradle/profiler/mutations/AbstractFileChangeMutator.java
```diff
@@ -19,21 +21,39 @@ protected AbstractFileChangeMutator(File sourceFile) {
...
+    }
+
+    @Override
+    public void beforeScenario(ScenarioContext context) {
+        scenarioText = modifyFile(modifiedText -> applyScenarioChangeTo(context, modifiedText));
+    }
+
+    protected void applyScenarioChangeTo(ScenarioContext context, StringBuilder text) {
+        // NOOP
     }
```
lptr, line 34: Is this an override, or why do we need to implement a noop method?

src/test/groovy/org/gradle/profiler/mutations/ApplyNonAbiChangeToKotlinSourceFileMutatorTest.groovy
```diff
@@ -1,17 +1,32 @@
...
+    def "adds public function at end of source file replacing its body"() {
         def sourceFile = tmpDir.newFile("Thing.kt")
         sourceFile.text = "class Thing { fun existingMethod() { }}"
         def mutator = new ApplyNonAbiChangeToKotlinSourceFileMutator(sourceFile)
+        def functionText = { qualifier ->
+            "class Thing { fun existingMethod() { }}\n" +
+                "private fun _m_276d92f3_16ac_4064_9a18_5f1dfd67992f_testScenario_3c4925d7() {" +
+                "requireNotNull(\"_276d92f3_16ac_4064_9a18_5f1dfd67992f_testScenario_3c4925d7_$qualifier\")" +
+                "} // _m_276d92f3_16ac_4064_9a18_5f1dfd67992f_testScenario_3c4925d77"
+        }
```
lptr, line 16: Can we share this with the other test as well?

lptr:
This is what I get before the fix: 

image

This is what I get after the fix:

image

Looks like it worked. (I also have no idea how build times can increase without the fix, but that's probably a bug in Kotlin.)

PR #220, opened 2020-07-24
Use reproducible JARs for the packaged artifacts

wolfs:
So that we can get build cache hits for the tests.


wolfs:
We should be able to revert this change once we update the build to Gradle 6.6.

lptr:
Let's do that then.

PR #221, opened 2020-07-24
Some usability improvements

adammurdoch:
This PR is a bucket of minor improvements:

- Add `--profiler async-profiler-heap`. This is intended to provide some preset async-profiler configuration to capture heap allocation snapshots. It also allows both CPU and heap snapshots to be captured by a single `gradle-profiler` invocation.
- Rework the YourKit options so that `--profiler yourkit` measures CPU usage using sampling, `--profiler yourkit-tracing` measure CPU usage using tracing, and `--profiler yourkit-heap` measures heap allocation. 
- Upgrade to new async-profiler version.
- Add a mutator to clear the instrumented build logic cache.


src/main/java/org/gradle/profiler/ProfilerFactory.java
```diff
@@ -0,0 +1,76 @@
...
+import org.gradle.profiler.yourkit.YourKitSamplingProfilerFactory;
+import org.gradle.profiler.yourkit.YourKitTracingProfilerFactory;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Represents some profiling strategy. Produces {@link Profiler} instances from a set of command-line options.
+ */
+public abstract class ProfilerFactory {
```
eskatos, line 21: 👍 for splitting profiler factories out

src/test/groovy/org/gradle/profiler/AsyncProfilerIntegrationTest.groovy
```diff
@@ -4,102 +4,141 @@ package org.gradle.profiler
...
-    @Requires({ OperatingSystem.isMacOS() })
-    def "profiles build using async-profiler with tooling API and warm daemon"() {
+    @Requires({ !OperatingSystem.isWindows() })
+    def "profiles build CPU usage using async-profiler with tooling API and warm daemon"() {
         given:
         instrumentedBuildScript()
 
         when:
-        new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--gradle-version", minimalSupportedGradleVersion, "--profile", "async-profiler",  "assemble")
+        new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--gradle-version", latestSupportedGradleVersion, "--profile", "async-profiler", "assemble")
```
eskatos, line 13: ❓ shouldn't we test both `minimalSupportedGradleVersion` and `latestSupportedGradleVersion`?

adammurdoch: yes, probably. I figured other tests took care of covering the actual Gradle invocation stuff for all of the versions, and this test can assume that it works fine. I changed it to the latest version because this is more likely to be used with the async-profiler integration (rather than Gradle 3.3, which is the minimal version) and also so I didn't need to figure out how to run the tests with Java 8 (Gradle 3.3 doesn't work on Java 11).

wolfs:
I wonder if it would be better to keep which profiler to use and what to profile separately on the command line. Something like:
```bash
> gradle-profiler --profiler async-profiler --profile heap
> gradle-profiler --profiler async-profiler --profile cpu # the default
> gradle-profiler --profiler yourkit --profile heap
> gradle-profiler --profiler jprofiler --profile heap
```

WDYT?

adammurdoch:
With the changes in this PR, I can do `gradle-profiler --profile async-profiler --profile yourkit --profile async-profiler-heap` and just come back when it's finished, and have the reports ready, rather than having to run `gradle-profiler` 2 or 3 times.

wolfs:
Ah, I didn't realize that you now can start multiple profiler runs with a single `gradle-profiler` invocation. That makes sense 👍 .

PR #222, opened 2020-07-24
Use canonical paths for temp-files

wolfs:
when they are inputs to the build. We currently can't watch symlinks on macOS, since we get the events for the canonical path. `/var` is a symlink to `/private/var` on macOS, and the tmp dir is in `/var`.

### DIFF
--- a/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java
+++ b/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java
@@ -34,7 +34,7 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
         this.projectDir = projectDir;
         this.gradleUserHome = gradleUserHome;
         this.daemonControl = daemonControl;
-        initScript = File.createTempFile("gradle-profiler", ".gradle");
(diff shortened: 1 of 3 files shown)


wolfs:
> Shouldn't we be also doing the same in https://github.com/gradle/gradle-profiler/blob/08a96120f0a22c999a652e6e605eb1d4c0803d0c/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java#L37❓

I did that, though it shouldn't make a difference since the init script is only loaded for probing and not for profiling. Moreover, the init script itself is not read via the VFS (yet), Gradle only uses the compiled init script which is in the Gradle user home.

@eskatos Do you have any idea if we still need this file: https://github.com/gradle/gradle-profiler/blob/0a72d39cadfba7114f28043b5ebaac28a5964d73/subprojects/chrome-trace/init.gradle

eskatos:
> @eskatos Do you have any idea if we still need this file: https://github.com/gradle/gradle-profiler/blob/0a72d39cadfba7114f28043b5ebaac28a5964d73/subprojects/chrome-trace/init.gradle

It's not used by anything in the build. I think it's been added as a convenience to use the chrome trace plugin from a clone of this repo. I never used it.

PR #224, opened 2020-07-30
optimized code

abhishek-abhi:


lptr:
Can you explain what you are trying to achieve here please?

abhishek-abhi:
> Can you explain what you are trying to achieve here please?
I have used Optional in few places to avoid scenarios where we are checking for null , which is not a specific type. In this way we are avoiding a scenario where NullPointer Exception can be thrown if we are checking null explicitly. In few places I have updated the conditional logic a little bit to do some computation little faster.

lptr:
I'm sorry, but I don't see how this is making things better. We could potentially improve nullability handling around these methods if we want to, though I'd do that in a different way. However, I don't see why this specific code would need to be improved to start with. Let's keep it the way it is.

PR #226, opened 2020-08-18
Fix help command not provided explicitly

### LINKED ISSUE #217: Recognize -h/--help
```
$ ./bin/gradle-profiler --help
help is not a recognized option
```
```
$ ./bin/gradle-profiler -h
h is not a recognized option
```
Both subsequently print the help output, however.

fajarnuha:
when executing `-h` or `--help` options show unrecognized command:
`h is not a recognized option`

this PR explicitly accepts both commands so that JOpt library can recognize

Fixes #217


lptr:
Thanks for the PR! I've added a bit more polish in https://github.com/gradle/gradle-profiler/pull/248. Let's continue there.

Issue #231, opened 2020-08-28
Expose project directory in BuildContext


lptr:
This would be useful for mutators.

wolfs:
I am not sure the `BuildContext` should expose the project directory. The project directory should be something which stays the same during the performance iterations. The project directory and the Gradle user home is something we expose when instantiating the mutator before running all the iterations.

lptr:
Right now `BuildContext` extends `ScenarioContext`. I think we should remedy that, and instead either add `BuildContext.getScenarioContext()` or pass `ScenarioContext` along with `BuildContext`. Either way, I think we should expose the whole `InvocationSettings` object in `ScenarioContext` (or pass/expose an `InvocationContext` object with similar data).

wolfs:
So you mean that a `BuildMutator` instance should live much longer than for only one scenario and should reset itself in `beforeScenario`?

I had more in mind that for each scenario Gradle profiler creates a new `BuildMutator` for the invocation settings.

lptr:
Let's revert this for now.

lptr:
Reverted in #235.

PR #232, opened 2020-08-28
Expose project dir in BuildContext and ScenarioContext

lptr:
This is a stop-gap measure for now. We should probably expose the whole of `InvocationSettings` and `ScenarioDefinition` in these contexts later on.


wolfs:
I don't think this is a good idea, see my comment on https://github.com/gradle/gradle-profiler/issues/231#issuecomment-683737139.

PR #235, opened 2020-09-01
Revert "Expose projectDir in BuildContext and ScenarioContext"

lptr:
This reverts the exposing of `projectDir` on `BuildContext` and `ScenarioContext`. Let's do this differently, and expose things like the `InvocationSettings` and `ScenarioDefinition` in a more complete way instead.


wolfs (approved):
LGTM!

PR #238, opened 2020-09-03
Fix PID instrumentation for configuration caching

wolfs:
Use a build service for PID instrumentation on Gradle versions >= 6.1, so the PID collector is
compatible with configuration caching.


subprojects/build-operations/src/main/java/org/gradle/trace/pid/PidCollectorBuildService.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public interface Parameters extends BuildServiceParameters {
+        RegularFileProperty getPidFile();
+    }
+
+    public PidCollectorBuildService() throws Exception {
+        GFileUtils.writeFile(getPid().toString(), getParameters().getPidFile().get().getAsFile());
+    }
+
+    private Integer getPid() throws NoSuchFieldException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
```
wolfs, line 35: I am not very happy with how we obtain the PID of the current process here, though I don't have a better idea :(.

src/test/groovy/org/gradle/profiler/AbstractProfilerIntegrationTest.groovy
```diff
@@ -16,7 +16,7 @@ abstract class AbstractProfilerIntegrationTest extends Specification {
         "3.3", "3.4.1", "3.5",
         "4.0", "4.1", "4.2.1", "4.7",
         "5.2.1", "5.5.1", "5.6.3",
-        "6.0.1", "6.1", "6.6-rc-2"
+        "6.0.1", "6.1", "6.6"
```
blindpirate, line 19: `6.6.1`?

wolfs: Done. I also upgradled the build to 6.6.1.

subprojects/build-operations/src/main/java/org/gradle/trace/pid/PidCollectorBuildService.java
blindpirate: I guess no better solution on Java 8.

blindpirate (approved):
LGTM, thanks for fixing!

Issue #240, opened 2020-09-04
HTML benchmark report generation fails when earlier scenario has more captured build operations


lptr:
Way to reproduce: start one scenario with some build operations captured, the next one with no build operations captured.

```text
java.lang.IndexOutOfBoundsException: index (1) must be less than size (1)
	at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1402)
	at com.google.common.base.Preconditions.checkElementIndex(Preconditions.java:1384)
	at com.google.common.collect.RegularImmutableList.get(RegularImmutableList.java:77)
	at org.gradle.profiler.report.HtmlGenerator.statistic(HtmlGenerator.java:202)
	at org.gradle.profiler.report.HtmlGenerator.write(HtmlGenerator.java:116)
	at org.gradle.profiler.report.AbstractGenerator.write(AbstractGenerator.java:18)
	at org.gradle.profiler.BenchmarkResultCollector.write(BenchmarkResultCollector.java:54)
	at org.gradle.profiler.Main.run(Main.java:86)
	at org.gradle.profiler.Main.main(Main.java:16)
```

PR #241, opened 2020-09-04
Fix scenario order bug

lptr:
Fixes https://github.com/gradle/gradle-profiler/issues/240


lptr:
Looks like it is fixed now:

image

wolfs (commented):
LGTM!

PR #242, opened 2020-09-04
Use an HTML template for the report

### LINKED ISSUE #196: Switch to order results in benchmark result graph
Currently results are shown like this on the result graph:

image

However, to compare two series it can be better to compare them in a way that every scenario is sorted, like so:

image

Add a switch on the generated report to toggle between the two modes.

### LINKED ISSUE #227: Use a step chart in the HTML report
The current separate points are hard to read and make it difficult to see trends, especially when there are several time series, each with only a handful of values. We should connect the points, but in a step-chart kind of way to avoid the suggestion of performance being a continuous thing.

### LINKED ISSUE #228: Show graphs for captured build operations in HTML report

lptr:
Fixes #196
Fixes #227
Fixes #228


lptr:
Ended up with implementing a bunch more here:

- sorting the timeline
- selecting build operations to compare (multiple ones supported)
- a new look for the data table at the bottom (lacks the baseline comparison feature as of now)

image

lptr:
The form is also ugly, will need some more CSS polish.

lptr:
OK, looks a lot better now I think.

lptr:
image
[benchmark.html.zip](https://github.com/gradle/gradle-profiler/files/5191023/benchmark.html.zip)

lptr:
@wolfs PTAL. I disabled animation. Couldn't find a simple way to fix the bottom scrollbar, and don't think it's worth spending a lot of time on anyway.

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+    const rest = pos - base;
+    if (sorted[base + 1] !== undefined) {
+        return sorted[base] + rest * (sorted[base + 1] - sorted[base]);
+    } else {
+        return sorted[base];
+    }
+};
+
+function mean(data) {
+    return data.reduce((sum, val) => sum += val) / data.length;
```
wolfs, line 176:
```suggestion
    return data.reduce((sum, val) => sum + val) / data.length;
```

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+const Phase = {
+    WARM_UP: "WARM_UP",
+    MEASURE: "MEASURE"
+};
+
+let benchmarkResult =
+@@BENCHMARK_RESULT_JSON@@
+;
+</script>
+    <script id="data-table-template" type="text/x-handlebars-template">
```
wolfs, line 53: Is there an advantage in using the `script` tag for the template instead of a `template` tag?

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+
+let source = document.getElementById("data-table-template");
+let target = document.getElementById("data-table");
+let template = Handlebars.compile(source.innerHTML);
+let maxIterations = Math.max(...(benchmarkResult.scenarios.map(scenario => scenario.iterations.length)));
+let html = template({
+    benchmarkResult: benchmarkResult,
+    iterationTitles: iterationTitles(),
+});
+// console.log(html);
```
wolfs, line 217: Please remove the commented out code.

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+
+Handlebars.registerHelper("calc", function(operation, sample, scenario) {
+    let data = iterations(scenario, Phase.MEASURE).map(iteration => iteration.values[sample.name]);
+    return OPERATIONS[operation](data).toFixed(2);
+});
+
+let source = document.getElementById("data-table-template");
+let target = document.getElementById("data-table");
+let template = Handlebars.compile(source.innerHTML);
+let maxIterations = Math.max(...(benchmarkResult.scenarios.map(scenario => scenario.iterations.length)));
```
wolfs, line 212: `maxIterations` doesn't seem to be used.

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+    benchmarkResult: benchmarkResult,
+    iterationTitles: iterationTitles(),
+});
+// console.log(html);
+target.innerHTML = html;
+</script>
+<script>
+const COLORS = ["#527AB3", "#56ac76", "#f44336", "#d49c3e", "#211f2d"];
+class PaletteGenerator {
+    constructor(colors) {
```
wolfs, line 223: Unused constructor parameter.

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+    };
+}
+
+function convertData(samples, sorted) {
+    let palette = new PaletteGenerator();
+    return samples.map(sample => benchmarkResult.scenarios.map(scenario => convertScenarioToChart(scenario, sample, sorted, palette.nextColor()))).flat();
+}
+
+function updateData() {
+    let sorted = $("input[name='sorting']:checked").val() == "sorted";
```
wolfs, line 262:
```suggestion
    let sorted = $("input[name='sorting']:checked").val() === "sorted";
```

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -0,0 +1,324 @@
...
+$(document).ready(function() {
+    let samples = benchmarkResult.scenarios
+        .map(scenario => scenario.samples)
+        .flat()
+        .map(sample => sample.name)
+        .unique();
+    samples.forEach(sample =>
+        $("#sample").append(new Option(sample, sample))
+    );
+    $("#sample").val(samples[0]);
```
wolfs, line 314: You could extract the selector `$("#sample")` to a local variable.

wolfs (commented):
The report looks much better!

I added some comments.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: This was recommended by [Handlebars' docs](https://handlebarsjs.com/installation/#usage):

> **Always use a script-tag for your template**
> 
> If you use this method, you have to wrap your template with a script-tag. Otherwise the browser may remove or modify parts of your template if you don't. Have a look at ["Unexpected markup in tables"](https://html.spec.whatwg.org/multipage/parsing.html#unexpected-markup-in-tables) for an example.

I tried using `<template>` but it didn't work.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: I've condensed the code a little more, but I think there is no problem with referring to the elements multiple times when it's for different purposes.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Done.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Removed.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Removed.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Removed.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Fixed.

src/main/resources/org/gradle/profiler/report/report-template.html
wolfs: 👍

wolfs (approved):
LGTM!

Issue #243, opened 2020-09-07
Redirect error to log file


blindpirate:
Currently, the error log in build is written into a memory `ByteArray`, which might cause OOM when build log is huge - and it happens.

https://github.com/gradle/gradle-profiler/blob/45c0f693afa1d785b3dddd08d88ff64f0a452a72/src/main/java/org/gradle/profiler/CommandExec.java#L142

https://builds.gradle.org/viewLog.html?buildId=37869263&buildTypeId=Enterprise_Master_Component_GradlePlugin_PerformanceDaily_PerformanceLinux&tab=buildResultsDiv

This build is failing because there're 3 huge `String` capturing the whole build log:

![image](https://user-images.githubusercontent.com/12689835/92341547-af2cfd80-f0f0-11ea-8f47-94da1ae56810.png)

We should redirect the build log to a file, then throw an exception which points to that file, instead of writing everything in memory.

PR #244, opened 2020-09-07
Add buildLog to InvocationSettings

blindpirate:
This is an attempt to resolve https://github.com/gradle/gradle-profiler/issues/243.
Previously, the build log is stored in memory, and we get OOMs when it's huge.
Now we allow users to specify a buildLog file which the build log will be redirected to.

### DIFF
--- a/src/main/java/org/gradle/profiler/InvocationSettings.java
+++ b/src/main/java/org/gradle/profiler/InvocationSettings.java
@@ -1,12 +1,13 @@
 package org.gradle.profiler;
 
-import org.gradle.profiler.report.CsvGenerator;
-
+import javax.annotation.Nullable;
 import java.io.File;
 import java.io.PrintStream;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+
+import static org.gradle.profiler.report.CsvGenerator.Format;
 
 public class InvocationSettings {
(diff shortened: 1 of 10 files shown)


src/main/java/org/gradle/profiler/InvocationSettings.java
```diff
@@ -170,4 +183,134 @@ public void printTo(PrintStream out) {
             }
         }
     }
+
```
wolfs, line 186: Nit: extra newline.

src/main/java/org/gradle/profiler/InvocationSettings.java
```diff
@@ -170,4 +183,134 @@ public void printTo(PrintStream out) {
...
+        private Map<String, String> sysProperties;
+        private File gradleUserHome;
+        private Integer warmupCount;
+        private Integer iterations;
+        private boolean measureConfigTime;
+        private List<String> measuredBuildOperations;
+        private Format csvFormat;
+        private File buildLog;
+
+        public InvocationSettingsBuilder() {
```
wolfs, line 207: You can remove the no-args constructor.

wolfs (approved):
LGTM, apart from the failures on CI.

PR #247, opened 2020-09-09
Allow `CliInvoker` redirect build log to file

blindpirate:
In https://github.com/gradle/gradle-profiler/pull/244/files we only added `buildLog` support for `ScenarioInvoker.measureCommandLineExecution`, but not for `CliInvoker`. This PR adds `buildLog` support to it.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
-profiler.version=0.15.0-alpha2
+profiler.version=0.15.0-alpha6
 org.gradle.caching=true
--- a/src/main/java/org/gradle/profiler/CliGradleClient.java
+++ b/src/main/java/org/gradle/profiler/CliGradleClient.java
@@ -14,12 +14,18 @@ public class CliGradleClient implements GradleInvoker, GradleClient {
     private final File javaHome;
     private final File projectDir;
     private final boolean daemon;
+    private final File buildLog;
 
-    public CliGradleClient(GradleBuildConfiguration gradleBuildConfiguration, File javaHome, File projectDir, boolean daemon) {
+    public CliGradleClient(GradleBuildConfiguration gradleBuildConfiguration,
+                      File javaHome,
+                      File projectDir,
+                      boolean daemon,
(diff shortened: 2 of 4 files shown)


src/main/java/org/gradle/profiler/CliInvoker.java
```diff
@@ -49,15 +55,21 @@ public void runTasks(List<String> tasks, List<String> gradleArgs, List<String> j
 
         ProcessBuilder builder = new ProcessBuilder(commandLine);
         builder.directory(projectDir);
-        if (!daemon) {
+        if (daemon) {
+            builder.environment().put("GRADLE_OPTS", "-Xmx128m -Xms128m -XX:+HeapDumpOnOutOfMemoryError");
```
wolfs, line 59: Why add these `GRADLE_OPTS` now?

wolfs (commented):
I added a question.

src/main/java/org/gradle/profiler/CliInvoker.java
blindpirate: Same as here: https://github.com/gradle/gradle/commit/f3c5ee6185c19b1e2af44fd8ee90b1c9c8338d26

When the build generates too much log, the log events might congest in client JVM and OOM. gradle-profiler doesn't provide an API to configure the client JVM or environment variables, right?

wolfs: Do we already know we need this? If we don't, I wouldn't add it now.

If we need something like that, I would make it configurable via something like `clientJvmArgs` on `CliInvoker`.

blindpirate: Yes, we need it for build scan performance test. This is just a workaround, I opened a new issue https://github.com/gradle/gradle-profiler/issues/260 to handle this issue and no-daemon jfr recording issue.

wolfs (approved):
LGTM!

PR #249, opened 2020-09-09
Reduce build operation measurement overhead

### LINKED ISSUE #214: Capturing build operations adds large build time overhead
It seems that when capturing build operations is enabled during a benchmark, builds run significantly slower. Here's a non-ABI change on Santa Tracker Kotlin without capturing build operations:

image

Trimmed mean of 50% |   |   |  
-- | -- | -- | --
Scenario | Execution | Difference
0 – Baseline | 2,813 ms | 0 ms | 0.00%
2 – FS watching | 2,712 ms | -100 ms | -3.57%

And here is the same result with capturing a number of operations around snapshotting:

image

Trimmed mean of 50% |   |   |  
-- | -- | -- | --
Scenario | Execution | Difference
0 – Baseline | 3,228 ms | 0 ms | 0.00%
2 – FS watching | 3,109 ms | -119 ms | -3.69%

There is a 12.4% increase at 400 ms when build operations are captured. However, with production builds we've seen far larger impacts at around 40%+.

lptr:
Instead of registering a `BuildOperationListener` for every build operation type captured, we now add only one that handles all the captured build operations. This reduces the overhead of the observation significantly from 68.5%, though it still stays rather high at 14.1%.

Fixes #214

### DIFF
--- a/subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
+++ b/subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
@@ -9,12 +9,21 @@ import org.gradle.api.services.BuildServiceParameters;
 import org.gradle.api.services.BuildServiceRegistry;
 import org.gradle.execution.RunRootBuildWorkBuildOperationType;
 import org.gradle.internal.build.event.BuildEventListenerRegistryInternal;
-import org.gradle.internal.operations.*;
+import org.gradle.internal.operations.BuildOperationDescriptor;
+import org.gradle.internal.operations.BuildOperationListener;
+import org.gradle.internal.operations.OperationFinishEvent;
+import org.gradle.internal.operations.OperationIdentifier;
+import org.gradle.internal.operations.OperationProgressEvent;
+import org.gradle.internal.operations.OperationStartEvent;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
 
 @SuppressWarnings("unused")
@@ -22,6 +31,7 @@ public class BuildOperationTrace {
 
     private final BuildEventListenerRegistryInternal registry;
     private final BuildServiceRegistry sharedServices;
+    private final Map<String, File> capturedBuildOperations = new LinkedHashMap<>();
 
     public BuildOperationTrace(GradleInternal gradle) {
         registry = gradle.getServices().get(BuildEventListenerRegistryInternal.class);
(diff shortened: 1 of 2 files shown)


lptr:
On the Gradle build the median overhead has been reduced significantly:

|                | 8 operations captured | Operation not captured |
| --------------- | ---------------------- | ------------------------- |
Before fix      | 7984.40 ms   (+68.5%)  | 4739.40
After fix       | 5388.82 ms   (+14.1%)  | 4724.26

It's still quite a significant overhead, but it's much better already.

Before:

image

After:

image

lptr:
@wolfs PTAL. I think this is as far as we can go with reducing the overhead. I think it's acceptable at this point, and we also have a better understanding of what's happening.

How and why the overhead disappears at times on macOS is still beyond me, though.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -37,11 +47,11 @@ public BuildOperationTrace measureConfigurationTime(File outFile) {
     }
 
     public BuildOperationTrace measureBuildOperation(String buildOperationName, File outFile) {
-        Provider<BuildOperationDurationRecordingListener> listenerProvider = sharedServices.registerIfAbsent("build-op-" + buildOperationName, BuildOperationDurationRecordingListener.class, spec -> {
-            spec.getParameters().getCapturedBuildOperation().set(buildOperationName);
-            spec.getParameters().getOutputFile().set(outFile);
+        Provider<BuildOperationDurationRecordingListener> listenerProvider = sharedServices.registerIfAbsent("measure-build-operations", BuildOperationDurationRecordingListener.class, spec -> {
```
wolfs, line 50: I don't think registering the same build service twice with different parameters has an effect. We should register the build service only when we know all build operations to register.

This seems to work now since the map is queried each time the `MapProperty` is queried.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -37,11 +47,11 @@ public BuildOperationTrace measureConfigurationTime(File outFile) {
...
+            spec.getParameters().getCapturedBuildOperations().set(capturedBuildOperations);
         });
         registry.onOperationCompletion(listenerProvider);
+        capturedBuildOperations.put(buildOperationName, outFile);
```
wolfs, line 54: I was expecting to find this `put` before the registration.

wolfs (changes requested):

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
lptr: That is the point, yes.

lptr: I see how a different design could work, too. Do you feel strongly to change it?

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
lptr: In the design you propose it would make sense that way, yes.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
lptr: I changed the API, so now we have the full map of captured operations up front.

wolfs (approved):
LGTM!

PR #250, opened 2020-09-09
Show cross-hair in HTML report

lptr:
This should make it easier to see which values belong together, and also make the tooltip easier to discover.

image

Example: 
[benchmark.html.zip](https://github.com/gradle/gradle-profiler/files/5197672/benchmark.html.zip)

### DIFF
--- a/src/main/resources/org/gradle/profiler/report/report-template.html
+++ b/src/main/resources/org/gradle/profiler/report/report-template.html
@@ -9,6 +9,7 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.13/js/select2.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/handlebars.js/4.7.6/handlebars.min.js"></script>
     <script src='https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.9.3/Chart.bundle.min.js'></script>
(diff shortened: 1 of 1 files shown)


src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -272,6 +273,16 @@
         datasets: []
     },
     options: {
+        plugins: {
+            crosshair: {
+                zoom: {
+                    enabled: false
```
wolfs, line 279: Hmm..., there is now a "Reset Zoom" button, though this button seems to do nothing. At least I don't know how to zoom.

wolfs (commented):
Looks good. I added a comment.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Where do you see that button? :o It shouldn't be displayed at all.

lptr:
[benchmark.html.zip](https://github.com/gradle/gradle-profiler/files/5202427/benchmark.html.zip)

wolfs:
I still see the button: 
image

If you say it is gone when you create a new benchmark from `master`, that is fine for me.

wolfs (commented):
LGTM!

PR #251, opened 2020-09-10
Report profiler version

### LINKED ISSUE #171: Record version number
We should print the version number of the tool used in the reports.

We also need a `--version` command line option.

lptr:
Expose the profiler version in the HTML report and on the command-line.

Fixes #171

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -76,6 +76,15 @@ java {
     withSourcesJar()
 }
 
+tasks.withType<Jar>().configureEach {
+    manifest {
+        attributes(
+            "Implementation-Title" to "Gradle Profiler",
+            "Implementation-Version" to archiveVersion
+        )
+    }
+}
+
 application.mainClassName = "org.gradle.profiler.Main"
 
 tasks.processResources {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,7 +80,7 @@ tasks.withType<Jar>().configureEach {
     manifest {
         attributes(
             "Implementation-Title" to "Gradle Profiler",
-            "Implementation-Version" to archiveVersion
+            "Implementation-Version" to project.version
         )
     }
 }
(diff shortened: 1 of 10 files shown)


build.gradle.kts
```diff
@@ -76,6 +76,15 @@ java {
     withSourcesJar()
 }
 
+tasks.withType<Jar>().configureEach {
+    manifest {
+        attributes(
+            "Implementation-Title" to "Gradle Profiler",
+            "Implementation-Version" to archiveVersion
```
wolfs, line 83: I think it would be better to use `project.version` here, since we may remove the version from the archive name.

src/test/groovy/org/gradle/profiler/report/JsonResultWriterTest.groovy
```diff
@@ -41,152 +41,155 @@ class JsonResultWriterTest extends Specification {
         when:
         writer.write([result1, result2], stringWriter)
         then:
-        stringWriter.toString() == """{
-  "scenarios": [
-    {
-      "definition": {
+        stringWriter.toString() ==~ /\{
+  "environment": \{
+    "profilerVersion": ".*"
```
wolfs, line 46: The version should be available here, so you could assert the actual version string, right?

wolfs (changes requested):
Looks mostly good, I added some comments.

Could you upload a `benchmark.html` file, so I can look at it without running an actual benchmark?

src/test/groovy/org/gradle/profiler/report/JsonResultWriterTest.groovy
lptr: It is `UNKNOWN` if you try to run the test from classes. But... yeah, we could do it that way.

build.gradle.kts
lptr: OK, fixed.

wolfs (approved):
LGTM!

PR #252, opened 2020-09-11
Support passing toolHome to Maven/Buck/Bazel invoker

### LINKED ISSUE #245: Allow passing the `MAVEN_HOME` location or the location of `mvn` to `MavenScenarioDefinition`
Similar to what we have in `GradleScenarioDefinition.GradleBuildConfiguration`. The location of the Maven executable or the Maven home should be enough for now.

This allows removing the workaround on the `gradle/gradle` performance test infrastructure: https://github.com/gradle/gradle/pull/14441/commits/c45726a2639c19336a0d0264538ff27d90e7325e

blindpirate:
So that they can be invoked programmatically more easily.

Fixes #245

### DIFF
--- a/src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
+++ b/src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
@@ -1,0 +1,55 @@
+package org.gradle.profiler;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.io.PrintStream;
+import java.util.List;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public abstract class BuildToolCommandLineScenarioDefinition extends ScenarioDefinition {
+    private final List<String> targets;
+    private final File toolHome;
+
+    public BuildToolCommandLineScenarioDefinition(
+        String name,
+        @Nullable String title,
+        List<String> targets,
+        Supplier<BuildMutator> buildMutator,
+        int warmUpCount,
+        int buildCount,
+        File outputDir,
+        @Nullable File toolHome
+    ) {
+        super(name, title, buildMutator, warmUpCount, buildCount, outputDir);
+        this.targets = targets;
+        this.toolHome = toolHome;
+    }
+
+    protected abstract String getExecutableName();
+
+    protected abstract String getToolHomeEnvName();
+
+    @Override
+    protected void printDetail(PrintStream out) {
+        out.println("  Targets: " + getTargets());
+    }
+
+    public String getExecutablePath() {
+        String toolHomePath = toolHome == null ? System.getenv(getToolHomeEnvName()) : toolHome.getAbsolutePath();
+        return toolHomePath == null ? getExecutableName() : toolHomePath + "/bin/" + getExecutableName();
+    }
+
+    @Override
+    public String getTasksDisplayName() {
+        return targets.stream().collect(Collectors.joining(" "));
+    }
+
+    public List<String> getTargets() {
+        return targets;
+    }
+
+    public File getToolHome() {
+        return toolHome;
+    }
+}
--- a/src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
+++ b/src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
@@ -5,7 +5,6 @@ import java.io.File;
 import java.io.PrintStream;
 import java.util.List;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
 public abstract class BuildToolCommandLineScenarioDefinition extends ScenarioDefinition {
     private final List<String> targets;
@@ -36,13 +35,13 @@ public abstract class BuildToolCommandLineScenarioDefinition extends ScenarioDef
     }
 
     public String getExecutablePath() {
-        String toolHomePath = toolHome == null ? System.getenv(getToolHomeEnvName()) : toolHome.getAbsolutePath();
+        String toolHomePath = getToolHome() == null ? System.getenv(getToolHomeEnvName()) : getToolHome().getAbsolutePath();
         return toolHomePath == null ? getExecutableName() : toolHomePath + "/bin/" + getExecutableName();
     }
 
     @Override
     public String getTasksDisplayName() {
-        return targets.stream().collect(Collectors.joining(" "));
+        return String.join(" ", targets);
     }
 
     public List<String> getTargets() {
--- a/src/main/java/org/gradle/profiler/MavenScenarioDefinition.java
+++ b/src/main/java/org/gradle/profiler/MavenScenarioDefinition.java
@@ -1,17 +1,22 @@
 package org.gradle.profiler;
 
+import javax.annotation.Nullable;
 import java.io.File;
-import java.io.PrintStream;
 import java.util.List;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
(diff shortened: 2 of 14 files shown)


src/main/java/org/gradle/profiler/MavenScenarioDefinition.java
```diff
@@ -9,8 +10,17 @@
...
-        super(scenarioName, title, buildMutator, warmUpCount, buildCount, outputDir);
+    public MavenScenarioDefinition(
+        String scenarioName,
+        String title,
+        List<String> targets,
+        Supplier<BuildMutator> buildMutator,
+        int warmUpCount,
+        int buildCount,
+        File outputDir,
+        @Nullable  File mavenHome
```
wolfs, line 21: Adding the tool home to `ScenarioDefinition` doesn't seem like the right place and introduces some oddness with the `GradleScenarioDefinition`. I think it would be better to introduce a new type like `CommandLineBuildToolScenarioDefinition` and add the home and the command name there.

Later there would maybe be a `MavenBuildConfiguration` and a `BazelBuildConfiguration` instead of add a `File` field to `ScenarioDefinition`. The `(Gradle|Maven|Bazel)BuildConfiguration`s then can have a base class `BuildToolConfiguration` which take the `toolHome` and the `executableName` or something like that. At some point, there may also be a `JvmBuildToolConfiguration` which captures the JVM arguments and the Java home for Gradle and Maven. That is something for later, though.

src/main/java/org/gradle/profiler/ScenarioInvoker.java
```diff
@@ -74,10 +76,30 @@ protected void runCleanup(BuildContext buildContext, BuildMutator mutator, Runna
                 new CommandExec().inDir(workingDir).runAndCollectOutput(buildLog, commandLine);
             }
             Duration executionTime = timer.elapsed();
-            return new BuildInvocationResult(buildContext, executionTime);
+            return (T) new BuildInvocationResult(buildContext, executionTime);
         };
     }
 
+    void doRun(T scenario, InvocationSettings settings, Consumer<BuildInvocationResult> resultConsumer, List<String> commandLine) {
```
wolfs, line 83: This method looks like it shouldn't be here, since it is not used by the `GradleScenarioInoker`. Maybe there should be an intermediate class for a `CommandLineBuildToolScenarioInvoker`?

I also saw that the `CliInvoker` (Gradle) uses `ProcessBuilder` directly while `measureCommandLineInvocation` uses `CommandExec`.

src/main/java/org/gradle/profiler/ScenarioDefinition.java
```diff
@@ -83,4 +97,13 @@ public void visitProblems(InvocationSettings settings, Consumer<String> reporter
 
     protected void printDetail(PrintStream out) {
     }
+
+    protected String getExecutableName() {
+        return getBuildToolDisplayName().toLowerCase();
```
wolfs, line 102: This won't work on `Windows` since it needs the `.bat` extension, right?

wolfs (changes requested):
I think we should add a separate base class for the `(Maven|Bazel)ScenarioDefinition` and the `(Maven|Bazel)ScenarioInvoker`. Then we can move the shared code from `ScenarioInvoker` there, instead of leaving the code in `ScenarioInvoker` and not using it in `GradleScenarioInvoker`.

wolfs:
@blindpirate There are quite some changes regarding the invokers going on in https://github.com/gradle/gradle-profiler/pull/253, I think we should hold off with this PR until https://github.com/gradle/gradle-profiler/pull/253 has been merged.

wolfs:
@blindpirate https://github.com/gradle/gradle-profiler/pull/253 has been merged, so we can go forward with this change.

src/main/java/org/gradle/profiler/ScenarioDefinition.java
```diff
@@ -83,4 +97,13 @@ public void visitProblems(InvocationSettings settings, Consumer<String> reporter
...
+    }
+
+    public String getExecutablePath() {
+        String toolHomePath = toolHome == null ? System.getenv(getBuildToolDisplayName().toUpperCase() + "_HOME") : getToolHome().getAbsolutePath();
```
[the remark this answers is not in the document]
blindpirate: Done.

src/main/java/org/gradle/profiler/MavenScenarioDefinition.java
blindpirate: Done.

src/main/java/org/gradle/profiler/ScenarioInvoker.java
blindpirate: Done.

src/main/java/org/gradle/profiler/ScenarioDefinition.java
blindpirate: Why? IIUC, when we say `mvn`, Windows will automatically search for `mvn.exe`/`mvn.bat`/`mvn.cmd`, right? BTW, default `mvn` command is `mvn.cmd`.

wolfs: Does this work for `ProcessBuilder` as well or only from the command-line? Can you try this out on Windows?

blindpirate: Seems like we need the extensions. I update the PR to `mvn.cmd`/`buck.exe`/`bazel.exe`

src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+import java.util.function.Supplier;
+
+public abstract class BuildToolCommandLineScenarioDefinition extends ScenarioDefinition {
+    private final List<String> targets;
+    private final File toolHome;
+
+    public BuildToolCommandLineScenarioDefinition(
+        String name,
+        @Nullable String title,
```
wolfs, line 15: Why is `title` `@Nullable` here, but not in any of the subclasses?

src/main/java/org/gradle/profiler/ScenarioInvoker.java
```diff
@@ -44,23 +44,23 @@ protected R runMeasured(BuildContext buildContext, BuildMutator buildMutator, Bu
     /**
      * Returns a {@link Supplier} that returns the result of the given command.
      */
-    BuildStepAction<BuildInvocationResult> measureCommandLineExecution(List<String> commandLine, File workingDir, File buildLog) {
-        return new BuildStepAction<BuildInvocationResult>() {
+    BuildStepAction<R> measureCommandLineExecution(List<String> commandLine, File workingDir, File buildLog) {
```
wolfs, line 47: Should this method go to `BuildToolCommandLineInvoker` as well?

wolfs (commented):
Nice work! I left two more comments.

src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
blindpirate: Fixed.

src/main/java/org/gradle/profiler/ScenarioInvoker.java
blindpirate: Done.

wolfs (approved):
LGTM! Can you add the new scenario field to the documentation? Maybe as part of the Maven, Bazel, Buck examples?

blindpirate:
Thanks @wolfs ! sure I will.

PR #253, opened 2020-09-11
Refactoring to prepare for improving support for Android Studio sync

adammurdoch:
This is a preparatory PR to abstract what kinds of actions can be measured, to prepare for some improved support for measuring and profiling Android Studio sync.

Specifically, introduce a "Gradle client", which is responsible for running and measuring different kinds of build actions. A dummy "Android studio" implementation is added but this does not yet do anything. The existing `GradleInvoker` implementations also become Gradle clients.

Also extract out the work of each of the build "steps" into various action types to simplify the main loops and make the implementations for each build tool more similar.


wolfs (commented):
I didn't find any problems with the PR, so LGTM!

adammurdoch:
Thanks for reviewing.

PR #254, opened 2020-09-14
Better HTML report model

lptr:
Vue.js allows us to better model things. Let's make use of it.

[test-report.html.zip](https://github.com/gradle/gradle-profiler/files/5218974/test-report.html.zip)

### DIFF
--- a/src/main/resources/org/gradle/profiler/report/report-template.html
+++ b/src/main/resources/org/gradle/profiler/report/report-template.html
@@ -83,9 +83,6 @@ let benchmarkResult =
             </label>
         </div>
     </div>
-    <div class="form-group col-md-1">
-        <button id="reset-zoom" class="btn btn-secondary">Reset Zoom</button>
-    </div>
     <div class="form-group col-md-1 col-form-label">
         <label for="sample">Samples:</label>
     </div>
--- a/src/main/resources/org/gradle/profiler/report/report-template.html
+++ b/src/main/resources/org/gradle/profiler/report/report-template.html
@@ -6,20 +6,26 @@
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.5.2/css/bootstrap.min.css" integrity="sha512-MoRNloxbStBcD8z3M/2BmnT+rg4IsMxPkXaGh2zD6LGNNFE80W3onsAhRcMAMrSoyWL9xD7Ert0men7vR8LUZg==" crossorigin="anonymous">
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.13/css/select2.min.css">
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.14.0/css/all.min.css">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto+Mono&display=swap">
     <script src="https://cdnjs.cloudflare.com/ajax/libs/vue/2.6.12/vue.min.js"></script>
     <script src='https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.9.3/Chart.bundle.min.js'></script>
     <script src='https://cdn.jsdelivr.net/npm/chartjs-plugin-crosshair@1.1.6'></script>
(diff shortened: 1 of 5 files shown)


src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -148,12 +142,21 @@
...
+    <span>Gradle Profiler version {{ benchmarkResult.environment.profilerVersion }}</span>
 </footer>
 </div>
 <script>
 Array.prototype.unique = function() { return [...new Set(this)]; };
 
+// Initialize sample config
+benchmarkResult.scenarios.forEach((scenario, scenarioIndex, scenarios) => {
+    scenario.samples.forEach((sample, sampleIndex, samples) => {
+        sample.color = `hsl(${scenarioIndex * 360 / scenarios.length}, ${100 - 80 * sampleIndex / samples.length}%, ${30 + 40 * sampleIndex / samples.length}%)`;
```
wolfs, line 155: I find the shades of red/green very hard to distinguish: 
image

Do you have any better ideas what colors to use?

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -190,132 +193,118 @@
...
-    let samples = $("#sample").val();
-    chart.data.datasets = convertData(samples, sorted);
-    chart.update();
-}
-
-let ctx = document.getElementById('samples').getContext('2d');
-let maxMeasuredIterations = Math.max(...(benchmarkResult.scenarios.map(scenario => iterations(scenario, Phase.MEASURE).length)));
-let chart = new Chart(ctx, {
-    type: 'line',
+var app = new Vue({
```
wolfs, line 218:
```suggestion
const app = new Vue({
```

No need to leave this mutable.

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -148,12 +142,21 @@
 </div>
 </div>
 <footer>
-    <span>Gradle Profiler version {{ environment.profilerVersion }}</span>
+    <span>Gradle Profiler version {{ benchmarkResult.environment.profilerVersion }}</span>
 </footer>
 </div>
 <script>
 Array.prototype.unique = function() { return [...new Set(this)]; };
```
wolfs, line 150: The `unique` function is now unused.

src/main/resources/org/gradle/profiler/report/report-template.html
```diff
@@ -190,132 +194,118 @@
...
-        if (this.colors.length) {
-            return this.colors.shift();
-        } else {
-            return "#8B899A";
-        }
-    }
-};
-
 function convertScenarioToChart(scenario, sample, sorted, color) {
     var data = iterations(scenario, Phase.MEASURE)
```
wolfs, line 198:
```suggestion
    const data = iterations(scenario, Phase.MEASURE)
```

wolfs (changes requested):
Nice! I added some comments.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: I've fixed this and others.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: I removed it and moved the utility functions into `Array` prototype functions.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Done.

src/main/resources/org/gradle/profiler/report/report-template.html
lptr: Not sure how important it is to distinguish the different shades of green (=different build operations for the same experiment), it seems more important to distinguish the green from the red (same build operations for different experiments).

lptr: That said I'm open to suggestions.

wolfs (approved):
LGTM! I still wonder if we can do something about the colors. Let's merge this PR and discuss to see if we can come up with something.

PR #258, opened 2020-09-17
Choose baseline in HTML report

lptr:
Before this change the baseline for each scenario was chosen based on some complex logic. It was also used to show diffs and confidence values for each measured sample in the narrow CSV report and the HTML report.

Now narrow CSV report don't have the concept of a baseline, so confidence values are gone. Also gone are the calculated values like mean and median from the CSV report – these should be easy to calculate in whatever tool the CSV itself is loaded in.

The concept of a baseline scenario now only lives in the HTML report. And the user can choose any of the scenarios as baseline.

There is one special case where we pick a default baseline: if there are multiple build tools (e.g. different Gradle versions) used, then we assume that the benchmark is a regression test, and thus we choose the first scenario as the baseline.

Example: 
[test-html-report.zip](https://github.com/gradle/gradle-profiler/files/5244064/test-html-report.zip)

image


build.gradle.kts
```diff
@@ -87,13 +89,29 @@ tasks.withType<Jar>().configureEach {
...
+node {
+    download = true
+}
+
+val generateHtmlReportJavaScript = tasks.register<NpxTask>("generateHtmlReportJavaScript") {
+    dependsOn("npmInstall")
+    val source = file("src/main/js/org/gradle/profiler/report/report.js")
+    val outputDir = layout.buildDirectory.dir("html-report")
+    val output = outputDir.map { it.file("org/gradle/profiler/report/report.js") }
+    inputs.file(source)
```
wolfs, line 101: I think we need to add `package.json` or `package-lock.json` as an input here as well.

build.gradle.kts
```diff
@@ -87,13 +89,29 @@ tasks.withType<Jar>().configureEach {
 
 application.mainClassName = "org.gradle.profiler.Main"
 
+node {
+    download = true
+}
+
+val generateHtmlReportJavaScript = tasks.register<NpxTask>("generateHtmlReportJavaScript") {
```
wolfs, line 96: Shall we move this task to a task class in `buildSrc` and maybe add a convention plugin in build src which does set up the Javascript parts?

build.gradle.kts
```diff
@@ -87,13 +89,29 @@ tasks.withType<Jar>().configureEach {
...
+}
+
 tasks.processResources {
     into("META-INF/jars") {
         from(profilerPlugins) {
             // Removing the version from the JARs here, since they are referenced by name in production code.
             rename("""(.*)-\d\.\d.*\.jar""", "${'$'}1.jar")
         }
     }
+    from(generateHtmlReportJavaScript.map { it.outputs })
```
wolfs, line 114: Can't you use `from(generateHtmlReportJavaScript)` directly?

wolfs (changes requested):
The report looks really nice now!

build.gradle.kts
```diff
@@ -100,7 +100,7 @@ val generateHtmlReportJavaScript = tasks.register<NpxTask>("generateHtmlReportJa
     val output = outputDir.map { it.file("org/gradle/profiler/report/report.js") }
     inputs.file(source)
     outputs.dir(outputDir)
-    command = "browserify"
+    command = "browserify@16.5.2"
```
wolfs, line 103: Why is this better? Isn't the lock-file better so we also lock in all the transitive dependencies?

wolfs (changes requested):

wolfs:
If it is easy, we could hide the whole `baseline` column if there is only one scenario.

build.gradle.kts
lptr: Isn't the dependency on `npmInstall` enough?

build.gradle.kts
lptr: #259 adds `buildSrc`. Let's not conflict with them, and do this afterwards.

build.gradle.kts
lptr: Good point, done.

build.gradle.kts
wolfs: The task does depend on `npmInstall`, though it will be up-to-date if `npmInstall` changes something, since `npmInstall` does not contribute to the inputs of this task. Or does it?

build.gradle.kts
wolfs: 👍

build.gradle.kts
lptr: I think this is fine like this, and it's fewer changes to the codebase. I don't expect this to break in a way that would otherwise be prevented by putting everything in a lock file. If it does, we can add back the lock file.

lptr:
> If it is easy, we could hide the whole baseline column if there is only one scenario.

I like this, let me take a look.

build.gradle.kts
wolfs: Fine with me.

build.gradle.kts
lptr: It should have `node_modules` as the input I think...

wolfs: Maybe it should, currently the `NpxTask` does not have the `package.json` file nor the `node_modules` directory as an input.

wolfs (approved):
LGTM!

PR #262, opened 2020-09-21
Support generating flame graphs from multiple jfr files

blindpirate:
Previously in `no-daemon` scenario, only the jfr file of last iteration will be used to generate flame graph, which makes troubleshooting no-daemon scenario extremely difficult. This PR support generating multiple jfr files and generate flame graphs from these files by merging them.

### DIFF
--- a/src/main/java/org/gradle/profiler/jfr/JFRJvmArgsCalculator.java
+++ b/src/main/java/org/gradle/profiler/jfr/JFRJvmArgsCalculator.java
@@ -35,7 +35,7 @@ public class JFRJvmArgsCalculator implements JvmArgsCalculator {
         if (profileOnStart) {
             String startArgs = "name=profile,settings=" + args.getJfrSettings();
             if (captureOnExit) {
-                startArgs += ",dumponexit=true,filename=" + outputFile.getAbsolutePath();
+                startArgs += ",dumponexit=true,dumponexitpath=" + outputFile.getParentFile().getAbsolutePath();
             }
             jvmArgs.add("-XX:StartFlightRecording=" + startArgs);
         }
--- a/src/main/java/org/gradle/profiler/jfr/JFRJvmArgsCalculator.java
+++ b/src/main/java/org/gradle/profiler/jfr/JFRJvmArgsCalculator.java
@@ -29,16 +29,21 @@ public class JFRJvmArgsCalculator implements JvmArgsCalculator {
             jvmArgs.add("-XX:+UnlockCommercialFeatures");
         }
         jvmArgs.add("-XX:+FlightRecorder");
-        jvmArgs.add("-XX:FlightRecorderOptions=stackdepth=1024");
         jvmArgs.add("-XX:+UnlockDiagnosticVMOptions");
         jvmArgs.add("-XX:+DebugNonSafepoints");
+
+        StringBuilder flightRecorderOptions = new StringBuilder("-XX:FlightRecorderOptions=stackdepth=1024");
+
         if (profileOnStart) {
-            String startArgs = "name=profile,settings=" + args.getJfrSettings();
+            jvmArgs.add("-XX:StartFlightRecording=name=profile,settings=" + args.getJfrSettings());
             if (captureOnExit) {
-                startArgs += ",dumponexit=true,dumponexitpath=" + outputFile.getParentFile().getAbsolutePath();
+                flightRecorderOptions.append(",defaultrecording=true,dumponexit=true")
+                    .append(",dumponexitpath=")
+                    .append(outputFile.getParentFile().getAbsolutePath());
             }
-            jvmArgs.add("-XX:StartFlightRecording=" + startArgs);
         }
+
+        jvmArgs.add(flightRecorderOptions.toString());
     }
 
     private boolean isOracleVm() {
--- a/src/main/java/org/gradle/profiler/jfr/JfrProfiler.java
+++ b/src/main/java/org/gradle/profiler/jfr/JfrProfiler.java
@@ -1,5 +1,6 @@
 package org.gradle.profiler.jfr;
 
+import org.gradle.profiler.GradleScenarioDefinition;
 import org.gradle.profiler.InstrumentingProfiler;
 import org.gradle.profiler.JvmArgsCalculator;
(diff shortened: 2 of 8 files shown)


src/main/java/org/gradle/profiler/jfr/JFRJvmArgsCalculator.java
```diff
@@ -29,16 +29,21 @@ public void calculateJvmArgs(List<String> jvmArgs) {
...
             if (captureOnExit) {
-                startArgs += ",dumponexit=true,filename=" + outputFile.getAbsolutePath();
+                flightRecorderOptions.append(",defaultrecording=true,dumponexit=true")
+                    .append(",dumponexitpath=")
+                    .append(outputFile.getParentFile().getAbsolutePath());
             }
-            jvmArgs.add("-XX:StartFlightRecording=" + startArgs);
         }
+
+        jvmArgs.add(flightRecorderOptions.toString());
```
blindpirate, line 46: According to https://docs.oracle.com/javacomponents/jmc-5-4/jfr-runtime-guide/run.htm#JFRUH164 it should be `-XX:FlightRecorderOptions=defaultrecording=true,dumponexit=true,dumponexitpath=path`. Previous `-XX:StartFlightRecording=dumponexit=true` is wrong.

src/main/java/org/gradle/profiler/jfr/JfrProfiler.java
```diff
@@ -45,4 +46,12 @@ protected JvmArgsCalculator jvmArgsWithInstrumentation(ScenarioSettings settings
     private File getJfrFile(ScenarioSettings settings) {
         return new File(settings.getScenario().getOutputDir(), settings.getScenario().getProfileName() + PROFILE_JFR_SUFFIX);
     }
+
+    @Override
+    public void validate(ScenarioSettings settings, Consumer<String> reporter) {
```
blindpirate, line 51: This overwrites parent class' `validate()` https://github.com/gradle/gradle-profiler/blob/966f032d33f9a91847afe00084c62e69f58c85a1/src/main/java/org/gradle/profiler/InstrumentingProfiler.java#L86

src/main/java/org/gradle/profiler/jfr/JfrProfiler.java
```diff
@@ -45,4 +46,12 @@ protected JvmArgsCalculator jvmArgsWithInstrumentation(ScenarioSettings settings
...
+        GradleScenarioDefinition scenario = settings.getScenario();
```
wolfs, line 52: Could you extract this validation into a method, so you don't need to copy all the code here?

wolfs (commented):
LGTM, I added one comment. Do we need to changes something in the documentation as well?

blindpirate:
@wolfs I don't see any doc mentioning this only-last-iteration-is-recorded-limitation, so I think doc doesn't need to be updated.

src/main/java/org/gradle/profiler/jfr/JfrProfiler.java
```diff
@@ -45,4 +45,9 @@ protected JvmArgsCalculator jvmArgsWithInstrumentation(ScenarioSettings settings
     private File getJfrFile(ScenarioSettings settings) {
         return new File(settings.getScenario().getOutputDir(), settings.getScenario().getProfileName() + PROFILE_JFR_SUFFIX);
     }
+
+    @Override
+    public void validate(ScenarioSettings settings, Consumer<String> reporter) {
+        validateMultipleIterationsWithCleanupAction(settings, reporter);
```
wolfs, line 51: Thinking about this some more: When the test doesn't re-use the daemon, then there can be cleanup actions. So `JfrProfiler.canRestartRecording()` should return true if the scenario doesn't re-use the daemon, right?

That is important for our first use scenarios, which have cleanup actions.

wolfs (changes requested):

src/main/java/org/gradle/profiler/jfr/JfrProfiler.java
blindpirate: Yes, correct.

src/test/groovy/org/gradle/profiler/JFRProfilerIntegrationTest.groovy
```diff
@@ -138,32 +139,44 @@ class JFRProfilerIntegrationTest extends AbstractProfilerIntegrationTest {
         latestSupportedGradleVersion  | _
     }
 
+    @Requires({ !OperatingSystem.isWindows() })
```
wolfs, line 142: Is it enough to only check the flamegraphs dir on Linux, and run the remainder of the test on Windows as well?

src/test/groovy/org/gradle/profiler/JFRProfilerIntegrationTest.groovy
```diff
@@ -138,32 +139,44 @@ class JFRProfilerIntegrationTest extends AbstractProfilerIntegrationTest {
...
     @Unroll
-    def "can profile Gradle #versionUnderTest using JFR, `gradle` command and no daemon"() {
+    def "can profile Gradle no daemon #versionUnderTest with #iterations iterations"(String versionUnderTest, int iterations) {
         given:
         instrumentedBuildScript()
 
         when:
-        new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--gradle-version", versionUnderTest, "--profile", "jfr", "--no-daemon", "assemble")
+        new Main().run(
+            "--project-dir", projectDir.absolutePath,
```
wolfs, line 151: It would be nice to add a cleanup action here to see that this is supported.

src/test/groovy/org/gradle/profiler/JFRProfilerIntegrationTest.groovy
blindpirate: Done.

src/test/groovy/org/gradle/profiler/JFRProfilerIntegrationTest.groovy
blindpirate: I created a new test.

wolfs (approved):
LGTM!

PR #264, opened 2020-09-21
Show benchmark title in HTML report

### LINKED ISSUE #257: Add title and description to benchmarks
Split off of https://github.com/gradle/gradle-profiler/issues/161.

Add a title along with some description for benchmark reports. It should be possible to declare these in the scenario file as well as on the command-line.

Add an option like this:

```console
$ gradle-profiler --benchmark --title "Test file system watching" --description "File system watching reduces snapshotting time. This should show up most when running incremental builds after non-ABI changes." ...
```

We should show both title and description on the HTML report.

----
/cc @gradle/execution

lptr:
Fixes #257.

image

[test-html-report.zip](https://github.com/gradle/gradle-profiler/files/5255454/test-html-report.zip)

### DIFF
--- a/src/main/java/org/gradle/profiler/InvocationSettings.java
+++ b/src/main/java/org/gradle/profiler/InvocationSettings.java
@@ -167,6 +167,9 @@ public class InvocationSettings {
         return csvFormat;
     }
 
+    /**
+     * The title of the benchmark. Shown on the HTML report and on the console.
+     */
     public String getBenchmarkTitle() {
         return benchmarkTitle;
     }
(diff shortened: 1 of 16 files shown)


lptr:
I've moved the controls to below the graph, removed the legend, and added buttons to select/deselect all entries and to show only the selected items. This makes it much easier to work with benchmarks with many scenarios and/or build operations.

src/main/java/org/gradle/profiler/InvocationSettings.java
```diff
@@ -164,6 +167,10 @@ public Format getCsvFormat() {
         return csvFormat;
     }
 
+    public String getTitle() {
```
wolfs, line 170: I think it would be good to have some Javadoc in this method which also helps distinguish between `InvocationSettings.title` and `ScenarioDefinition.title`.

wolfs (commented):
LGTM! I left a comment.

src/main/java/org/gradle/profiler/InvocationSettings.java
lptr: I opted to rename it to `benchmarkTitle` instead. WDYT?

PR #265, opened 2020-09-21
Revert toolchains

lptr:
Toolchain support seems to make builds fail now, because Oracle JDK cannot be found.


lptr:
Not so easy, unfortunately. Going to disable the JFR-related tests on macOS for now: see #266.

PR #267, opened 2020-09-22
Extract publication logic to buildSrc

blindpirate:
After introducing `client-protocol` subproject, we need to publish
more than 1 project, so we extract the common logic to buildSrc.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,11 +1,10 @@
-import java.net.URI
 import com.moowork.gradle.node.npm.NpxTask
 
 plugins {
     id("profiler.java-library")
     groovy
     application
-    `maven-publish`
+    id("profiler.publication")
     id("com.github.node-gradle.node") version "2.2.4"
 }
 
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/profiler.publication.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+publishing {
+    publications {
+        register<MavenPublication>("mavenJava") {
+            from(components["java"])
+
+            if (project.parent == null) {
+                afterEvaluate {
+                    val profilerDistribution = artifacts.add("archives", tasks.named<Zip>("distZip").flatMap { it.archiveFile }) {
```
wolfs, line 14: This should go into the root project. I don't see why the general publication plugin would add the zip.

wolfs (changes requested):

build.gradle.kts
```diff
@@ -141,33 +141,13 @@ val profilerDistribution = artifacts.add("archives", tasks.distZip.flatMap { it.
...
-            credentials {
-                username = project.findProperty("artifactoryUsername") as String?
-                password = project.findProperty("artifactoryPassword") as String?
-            }
         }
     }
 }
 
-fun Project.gradleInternalRepositoryUrl(): URI {
+fun Project.gradleInternalRepositoryUrl(): java.net.URI {
```
wolfs, line 150: Any reason why we don't use the import here any more?

wolfs (approved):
LGTM! I left one comment.

build.gradle.kts
blindpirate: This is an IDEA auto-format issue, will fix.

PR #268, opened 2020-09-22
Fix two issues with Android Studio Sync support 

eskatos:
First commit fixes the case where running the profiler from inside the project under test directory passing `.` as the project dir by using an absolute path.

Second commit fixes the case where the path to studio install directory is actually a JetBrains Toolbox Launcher by resolving the actual Studio install directory from the launcher info file.

With these two fixes I was able to produce flame graphs for Studio Sync on my mac.


adammurdoch (approved):
Thanks for fixing these.

Issue #151, opened 2019-08-21
Document Linux kernel parameters required for async-profiler


lfielke:
Gradle-profile using async-profile did not work out of the box for me on Ubuntu and Arch Linux, instead it showed a cryptic error like the following:

```
java.lang.RuntimeException: Could not run command perl /tmp/flamegraph1180508986208029728.pl /home/user/dev/sandbox/GradleTaskTime/profile-out-2/5.6.simplified-stacks.txt --colors java --minwidth 1 --title CPU Flame Graph --countname samples
Output:
======
ERROR: No stack counts found
...
```

After trying to get async-profiler working stand-alone, I discovered there are some Linux kernel parameters that need to be set to allow non-root profiling. Setting these allowed gradle-profiler to work.

I found this quite confusing, and had tried gradle-profiler on two different computers over a couple of days before I discovered the fix. It would be handy to mention it in the readme under the [Async profiler](https://github.com/gradle/gradle-profiler#async-profiler) section.

For reference, see https://github.com/jvm-profiling-tools/async-profiler#basic-usage

> Basic Usage
> 
> As of Linux 4.6, capturing kernel call stacks using perf_events from a non- root process requires setting two runtime variables. You can set them using sysctl or as follows:
> 
> echo 1 > /proc/sys/kernel/perf_event_paranoid
> echo 0 > /proc/sys/kernel/kptr_restrict

autonomousapps:
This same thing happened to me, but I understand why they wouldn't want to duplicate documentation. As you can see, the Async Profiler docs (linked from this project) do indicate what you need to do.

That said, I have a script I wrote that lets me set and unset these values for when I'm doing profiling on my Linux computer.

```
#!/usr/bin/env bash
# Must run with sudo
# /etc/sudoers.d/ must include this file in its secure_path.
# Without this, ~/.local/async-profiler would not work.
# Original values: 3 (paranoid), 1 (restrict)
if [[ $EUID > 0 ]]; then
    echo "Must run as root" >&2
    exit 1
fi

if [ "$#" -ne 1 ]; then
    echo "Must pass exactly one parameter, either 'set' or 'restore'" >&2
    exit 1
fi

if [ $1 = "set" ]; then
    echo "Setting up async-profiler"
    echo 1 > /proc/sys/kernel/perf_event_paranoid
    echo 0 > /proc/sys/kernel/kptr_restrict
    echo "Success"
elif [ $1 = "restore" ]; then
    echo "Restoring original values"
[...]

PR #270, opened 2020-09-23
Add Troubleshooting section to the project README.md

Mishkun:
After stumbling across mentioned RuntimeException, wasted
enough time to add this error message to the README.md
as proposed in issue #151


Mishkun:
Guess the maintainers are not interested

ljacomet:
Sorry for the late reply

We believe this addition makes sense and will look into merging those changes

PR #273, opened 2020-09-28
Remove all of the transforms cache dir

lptr:
This makes us independent of the structure of said directory. Gradle 6.8 will have a new structure for the directory: https://github.com/gradle/gradle/pull/14588.

### DIFF
--- a/src/main/java/org/gradle/profiler/mutations/ClearArtifactTransformCacheMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/ClearArtifactTransformCacheMutator.java
@@ -4,10 +4,6 @@ import com.typesafe.config.Config;
 import org.gradle.profiler.BuildMutator;
 
 import java.io.File;
-import java.io.FilenameFilter;
-import java.util.Arrays;
-import java.util.Objects;
-import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


adammurdoch (approved):

src/main/java/org/gradle/profiler/mutations/ClearArtifactTransformCacheMutator.java
```diff
@@ -17,15 +13,7 @@ public ClearArtifactTransformCacheMutator(File gradleUserHome, CleanupSchedule s
...
-        filesAsStream(
-            cacheDir,
-            (dir, name) -> name.startsWith("files-")
-        ).flatMap(file -> filesAsStream(file, (dir, name) -> !name.endsWith(".lock")))
-            .forEach(AbstractCleanupMutator::delete);
-    }
-
-    private Stream<File> filesAsStream(File dir, FilenameFilter filter) {
-        return Arrays.stream(Objects.requireNonNull(dir.listFiles(filter)));
+        delete(cacheDir);
```
wolfs, line 16: Can we delete the cache dir when the daemon is running? I suppose we won't be able to remove the `.lock` file, right?

PR #274, opened 2020-09-29
Do not allow cleanup actions to run between builds with warm daemon

lptr:
Warm daemons have a lot of in-memory state that could get confused if these directories are deleted while they are alive.


src/main/java/org/gradle/profiler/GradleBuildInvoker.java
```diff
@@ -19,6 +19,11 @@ private GradleBuildInvoker(GradleClientSpec client, GradleDaemonReuse daemonReus
         this.daemonReuse = daemonReuse;
     }
 
+    @Override
+    public boolean allowsCleanupBetweenBuilds() {
+        return daemonReuse == GradleDaemonReuse.ColdDaemonOnly;
```
wolfs, line 24:
```suggestion
        return !isReuseDaemon();
```

wolfs (changes requested):
I wonder if we are going to have a problem when integrating the changes about instantiating the BuildMutators directly. I suppose we shouldn't, right?

src/main/java/org/gradle/profiler/GradleBuildInvoker.java
lptr: Added with a comment.

lptr:
There will be some changes necessary, but I think it's going to be fine on the Gradle side (and considerably simpler, too).

wolfs (approved):
LGTM!

PR #275, opened 2020-10-01
Fix issue with system properties not being applied on config reading

maxim-pandra:
Previously, when you run profiler with some scenario and add system property with -Dkey=value. This property would not be passed to a configurationReader. So in our case, nexus credentials were not provided and :help task in the configuration reader couldn't be executed.

I've added an additional parameter to a BuildConfigurationReader, so it taking into account system properties.


ljacomet:
Sorry for the late reply

PR #276, opened 2020-10-01
Release automatically to SDKMAN

wolfs:
Publishes releases automatically to SDKMAN.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -199,6 +199,14 @@ tasks.withType<SdkAnnounceVersionTask>().configureEach {
 
 tasks.register("releaseToSdkMan") {
     val versionString = project.version.toString()
+    // We only announce and set the default version for final releases
+    // A release is not final if it contains things different to numbers and dots.
+    // For example:
+    //   - 1.3: final
+    //   - 1.3.25: final
+    //   - 1.3-rc-4: not final
+    //   - 1.3.RC5: not final
+    //   - 1.3-alpha5: not final
     val isFinalRelease = Regex("""[0-9\.]*""").matchEntire(versionString) != null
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -164,9 +169,40 @@ tasks.register<Exec>("gitTag") {
...
+    mustRunAfter(tasks.withType<SdkReleaseVersionTask>())
+}
+
+tasks.withType<SdkAnnounceVersionTask>().configureEach {
+    mustRunAfter(tasks.withType<SdkReleaseVersionTask>())
+}
+
+tasks.register("releaseToSdkMan") {
+    val versionString = project.version.toString()
+    val isFinalRelease = Regex("""[0-9\.]*""").matchEntire(versionString) != null
```
lptr, line 202: This could use a bit of commenting. And it would be great to mention the release logic and the fact that we are publishing to SDKMAN in the `RELEASING.md`. That file is in dire need of updating anyway.

lptr (changes requested):
LGTM, I think we need to better document this though.

build.gradle.kts
wolfs: I added a comment and updated the release instructions.

wolfs:
@lptr PTAL!

build.gradle.kts
lptr: Thanks. I think the way the version number format controls the SDKMAN publishing behavior should also be mentioned in `RELEASING.md`.

wolfs: I added it the releasing instructions as well.

lptr (approved):
LGTM.

PR #277, opened 2020-10-02
Measure garbage collection time

lptr:
What it says on the tin. Enable with `--measure-gc`.

image


lptr:
I've tried two approaches to capturing the time, see the two commits.

wolfs (commented):
LGTM! Can you add the new flag to the README as well?

PR #280, opened 2020-10-06
Add ToolingApiGradleClient.runOperation

wolfs:
So we don't have to use reflection on the Gradle side
to invoke these methods.

For reference, we use those fields here:
https://github.com/gradle/gradle/blob/b3f0f8d19e6a1f82d1f2169ff79b77f970de71ef/subprojects/internal-performance-testing/src/main/groovy/org/gradle/performance/fixture/CrossVersionPerformanceTestRunner.groovy#L350-L357


lptr:
LGTM.

PR #290, opened 2020-10-28
Support passing JVM arguments with spaces to the Gradle daemon

wolfs:
When passing JVM arguments with spaces to the Gradle daemon, Gradle profiler failed on Windows when using the CLI invoker. This PR fixes the problem by passing in `org.gradle.jvmargs` via `GRADLE_OPTS` to the client.

### DIFF
--- a/src/main/java/org/gradle/profiler/CliGradleClient.java
+++ b/src/main/java/org/gradle/profiler/CliGradleClient.java
@@ -44,10 +44,7 @@ public class CliGradleClient implements GradleInvoker, GradleClient {
 
     @Override
     public void runTasks(List<String> tasks, List<String> gradleArgs, List<String> jvmArgs) {
-        String gradleOpts = jvmArgs.stream()
-            .peek(arg -> { if (arg.contains("\"") || arg.contains("'")) { throw new IllegalArgumentException("jvmArgs must not contain quotes, but this argument does: " + arg); } })
-            .map(arg -> "'" + arg + "'")
-            .collect(Collectors.joining(" "));
+        String gradleOpts = quoteJvmArguments(daemon, jvmArgs);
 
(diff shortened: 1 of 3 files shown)


blindpirate (approved):
LGTM, I have only a question.

src/main/java/org/gradle/profiler/CliGradleClient.java
```diff
@@ -44,23 +44,24 @@ public void loadToolingModel(List<String> tasks, List<String> gradleArgs, List<S
...
 
         List<String> commandLine = new ArrayList<>();
         gradleBuildConfiguration.addGradleCommand(commandLine);
         commandLine.addAll(gradleArgs);
         commandLine.addAll(tasks);
         commandLine.add("-Dorg.gradle.daemon=" + daemon);
-        if (daemon) {
-            commandLine.add("-Dorg.gradle.jvmargs=" + gradleOpts);
-        } else {
+        if (!daemon) {
```
blindpirate, line 54: Unrelated question, why do we add empty `jvmargs` here?

wolfs: I think this is because we want to make sure we don't start a daemon after all. Though I am not sure why exactly, I left it in place just in case.

Safran1022:
wolfs/pass-jvmOpts-via-gradle-opts

src/main/java/org/gradle/profiler/CliGradleClient.java
Safran1022: wolfs/pass-jvmOpts-via-gradle-opts[](url)

Safran1022:
invokerString = runUsing == "cli" ? "`gradle` command" : "tooling API"

PR #292, opened 2020-11-23
Fix README: Android incremental build examples had used invalid path

carstenhag:
The values folder has an `s` at the end, the example was missing.  
I copied over most of that example and it could not find the strings.xml file, as the path was wrong.


carstenhag:
Pretty weird with that DCO bot, as I had proposed the change via the Github Website.

wolfs (approved):
LGTM!

@carstenhag Thank you for fixing the documentation!

PR #295, opened 2020-11-26
Allow configuration cache state to be cleaned up between builds

bamboo:


eskatos (approved):

src/main/java/org/gradle/profiler/mutations/ClearConfigurationCacheStateMutator.java
```diff
@@ -17,6 +18,11 @@ public ClearConfigurationCacheStateMutator(File projectDir, CleanupSchedule sche
         this.projectDir = projectDir;
     }
 
+    @Override
+    public void validate(BuildInvoker invoker) {
```
wolfs, line 22: Hmm... Are there any tests for this mutator? I would be great if we could add an integration test which runs a build with the daemon.

wolfs (approved):
Looks good! Can you add a test, please?

PR #296, opened 2020-11-26
Add a profiler that takes a heap dump at the end of the build

adammurdoch:


subprojects/heap-dump/src/main/java/org/gradle/trace/heapdump/HeapDump.java
```diff
@@ -0,0 +1,54 @@
...
+
+        @Override
+        public void onFinish(FinishEvent event) {
+            // Ignore
+        }
+
+        @Override
+        public void close() {
+            try {
+                File dumpFile = new File(getParameters().getBaseFile().get() + "-heap-" + (++counter) + ".hprof");
```
wolfs, line 44: Is there any reason to use a build service instead of a `JCmd` invocation? The `jCmd` invocation would also have the advantage that it wouldn't affect the build time to take a heap dump.

wolfs (commented):
I wonder why you didn't use JCmd to generate the heap dumps. That would also not create heap dumps from warmups.

subprojects/heap-dump/src/main/java/org/gradle/trace/heapdump/HeapDump.java
adammurdoch: If we use jcmd (which is what I did initially), then the heap dump happens after all of the build state has been released, which isn't useful for looking at what state is retained during the build. Using a build service allows heap dumps to happen at various points in the build lifecycle. For now, it happens during "build finished" which is good enough for my purposes, but could happen at any number of places.

We can skip heap dumps for warmups via some system property if we want to.

PR #299, opened 2020-12-11
Run some tests on Java 11

wolfs:


.teamcity/settings/extensions.kt
```diff
@@ -21,6 +21,17 @@ fun ParametrizedWithType.java8Home(os: Os) {
     param("env.JAVA_HOME", "%${os.name}.java8.oracle.64bit%")
 }
```
lptr, line 22: Do we still need `java8Home()`? If so, I think it call through to `javaHome()`.

.teamcity/settings/root-project.kt
```diff
@@ -7,6 +7,7 @@ fun Project.configureGradleProfilerProject() {
     buildType(MacOSJava18)
     buildType(WindowsJava18)
     buildType(LinuxJava18)
+    buildType(LinuxJava11)
```
lptr, line 10: I think this would be a good opportunity to rename `Java18` to `Java8`.

src/test/groovy/org/gradle/profiler/AbstractProfilerIntegrationTest.groovy
```diff
@@ -8,18 +10,40 @@ abstract class AbstractProfilerIntegrationTest extends AbstractIntegrationTest {
 
     private static int NUMBER_OF_HEADERS = 4
 
+    static List<String> supportedOnCurrentJvm(List<String> gradleVersions) {
```
lptr, line 13: How about calling this `gradleVersionsSupportedOnCurrentJvm()`?

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1349,7 +1354,7 @@ buildGoal {
         def scenarios = file("performance.scenario")
         scenarios.text = """
 buildTarget {
-    versions = ["4.5"]
+    versions = ["5.2"]
```
lptr, line 1357: Why the upgrade here?

wolfs: Because 4.5 doesn't work with Java 11.

.teamcity/settings/extensions.kt
wolfs: I removed `java8Home`.

.teamcity/settings/root-project.kt
wolfs: I can do this post-merge, so I still run the tests on my branch. Renaming the object causes a change in the ID in Teamcity, so I can't do this change on a branch.

src/test/groovy/org/gradle/profiler/AbstractProfilerIntegrationTest.groovy
wolfs: Renamed.

wolfs:
@lptr PTAL!

lptr (approved):
LGTM.

.teamcity/settings/root-project.kt
wolfs: Done: https://github.com/gradle/gradle-profiler/commit/53c13e43cab0935424adb65fd4ce0cf2bc199792

PR #309, opened 2021-02-04
Update README.md

quanturium:
Fixing name of --warmups command parameter


lptr:
Thanks! Could you please sign your commit so it complies with the DCO?

carstenhag (changes requested):
At line 234 there's still a `warm-ups` usage.

blundell:
In my scenario file I have `warm-ups` (running version 16.0.1)

If I change it to `warmups` I get 

> java.lang.IllegalArgumentException: Unrecognized key 'my_scenario.warmups' defined in scenario file tools/gradle-profiler/my.scenarios

Is this readme change a mistake or for something else?

ljacomet:
Sorry for the late reply

The current version of the README clarifies that the CLI uses `--warmups` while scenario files use `warm-ups`

PR #310, opened 2021-02-07
Add execute-command-before BuildMutator

### LINKED ISSUE #335: Feature request: a `run` mutation
While there are a number of mutations today for defining scenarios, we've found them unhelpful in some cases due to either being too trivial or too limiting. I'd like to propose a new `run` mutation that would act similar to [`git bisect run`](https://git-scm.com/docs/git-bisect#_bisect_run), where an arbitrary shell script + any arguments is given. This would be invoked before each iteration to apply a set of changes handled by the script.

One example where we would want to use this is to apply changes that would specifically change the dependency graph in a build profiling Dagger, such as adding a parameter to an injected class's constructor. This would give flexibility for scenario authors to profile specific cases unique to their build.

AlexBeggs:
- BuildMutator allows commands to be executed prior to the SCENARIO or BUILD phase. Enabled for all types of builds, Gradle, Bazel, Buck, and Maven.

Dependent on https://github.com/gradle/gradle-profiler/pull/315/

### DIFF
--- a/src/main/java/org/gradle/profiler/ScenarioLoader.java
+++ b/src/main/java/org/gradle/profiler/ScenarioLoader.java
@@ -15,6 +15,7 @@ import org.gradle.profiler.mutations.ApplyChangeToPropertyResourceFileMutator;
 import org.gradle.profiler.mutations.ApplyNonAbiChangeToSourceFileMutator;
 import org.gradle.profiler.mutations.ApplyValueChangeToAndroidResourceFileMutator;
 import org.gradle.profiler.mutations.BuildMutatorConfigurator;
+import org.gradle.profiler.mutations.ExecuteCommandBuildMutator;
 import org.gradle.profiler.mutations.ClearArtifactTransformCacheMutator;
 import org.gradle.profiler.mutations.ClearBuildCacheMutator;
 import org.gradle.profiler.mutations.ClearConfigurationCacheStateMutator;
@@ -37,6 +38,10 @@ import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 class ScenarioLoader {
+
+    static final String BAZEL = "bazel";
+    static final String BUCK = "buck";
+    static final String MAVEN = "maven";
     private static final String TITLE = "title";
     private static final String VERSIONS = "versions";
     private static final String TASKS = "tasks";
@@ -45,9 +50,6 @@ class ScenarioLoader {
     private static final String RUN_USING = "run-using";
     private static final String DAEMON = "daemon";
     private static final String SYSTEM_PROPERTIES = "system-properties";
-    private static final String BAZEL = "bazel";
-    private static final String BUCK = "buck";
-    private static final String MAVEN = "maven";
     private static final String TOOL_HOME = "home";
     private static final String WARM_UP_COUNT = "warm-ups";
     private static final String ITERATIONS = "iterations";
@@ -61,6 +63,7 @@ class ScenarioLoader {
     private static final String APPLY_ANDROID_MANIFEST_CHANGE_TO = "apply-android-manifest-change-to";
     private static final String APPLY_CPP_SOURCE_CHANGE_TO = "apply-cpp-change-to";
     private static final String APPLY_H_SOURCE_CHANGE_TO = "apply-h-change-to";
+    private static final String EXECUTE_COMMAND_BEFORE = "execute-command-before";
     private static final String CLEAR_BUILD_CACHE_BEFORE = "clear-build-cache-before";
     private static final String CLEAR_GRADLE_USER_HOME_BEFORE = "clear-gradle-user-home-before";
     private static final String CLEAR_INSTANT_EXECUTION_STATE_BEFORE = "clear-instant-execution-state-before";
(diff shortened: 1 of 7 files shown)


src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -212,80 +217,48 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
...
-                .map(entry -> entry.getValue().configure(scenario, scenarioName, settings, entry.getKey()))
+                .filter(entry -> getBuildConfig(scenario, settings).hasPath(entry.getKey()))
+                .map(entry -> entry.getValue().configure(getBuildConfig(scenario, settings), scenarioName, settings, entry.getKey()))
                 .filter(mutator -> mutator != BuildMutator.NOOP)
                 .collect(Collectors.toList());
 
             int buildCount = getBuildCount(settings, scenario);
             File scenarioBaseDir = new File(settings.getOutputDir(), GradleScenarioDefinition.safeFileName(scenarioName));
 
-            if (scenario.hasPath(BAZEL) && settings.isBazel()) {
```
sunyal, line 223: Refactored while I was here to make it more readable.  Extracted this into separate methods instead of a single large method

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -212,80 +217,48 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
...
+            scenario = rootScenario.getConfig(BUCK);
+        } else if (settings.isMaven()) {
+            scenario = rootScenario.getConfig(MAVEN);
+        } else {
+            scenario = rootScenario;
+        }
+        return scenario;
+    }
+    
+    private static Config getExecutionInstructions(File scenarioFile, InvocationSettings settings, String scenarioName, Config scenario, String toolName, List<String> toolKeys) {
```
sunyal, line 261: renamed method name to be clearer

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -298,6 +271,128 @@ private static Config getConfig(File scenarioFile, InvocationSettings settings,
         return executionInstructions;
     }
 
+    private static List<ScenarioDefinition> getScenarioDefinitions(File scenarioFile,
```
sunyal, line 274: this is the section of extracted methods

src/main/java/org/gradle/profiler/mutations/ExecuteCommandBuildMutator.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + schedule + ")";
+    }
+
+    public enum ExecuteCommandSchedule {
+        SCENARIO, BUILD
```
sunyal, line 102: doesn't handle CLEANUP, Bazel, Buck and Maven don't implement that

src/main/java/org/gradle/profiler/mutations/ProcessBuilderCommandInvoker.java
```diff
@@ -0,0 +1,24 @@
...
+        try {
+            if (command == null || command.isEmpty()) {
+                throw new IllegalArgumentException(
+                    String.format("command cannot be null or empty, was %s", command));
+            }
+            ProcessBuilder processBuilder = new ProcessBuilder(command)
+                .redirectOutput(ProcessBuilder.Redirect.INHERIT)
+                .redirectError(ProcessBuilder.Redirect.INHERIT);
+            Process process = processBuilder.start();
+            return process.waitFor();
```
sunyal, line 19: could potentially add a timeout but for now I left it

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -212,80 +218,26 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
             String title = scenario.hasPath(TITLE) ? scenario.getString(TITLE) : null;
 
             List<BuildMutator> mutators = BUILD_MUTATOR_CONFIGURATORS.entrySet().stream()
-                .filter(entry -> scenario.hasPath(entry.getKey()))
```
sunyal, line 215: filtering isn't done here. The BuildMutator's should be responsible for determining if they support the configuration or not.

ZacSweers:
Does this run commands in the root directory of the project? It would be helpful I think to have an example that runs a script that is local to the project being profiled since I think that would be a common case.

AlexBeggs:
If I recall correctly, this runs the command in the working directory or `user.dir`

h0tk3y:
Sorry for the late reply.

This pull request needs a decision from the team responsible for that area. They have been informed. Response time may vary.

alllex:
Sorry for the late reply

This looks outdated and will thus be closed. If you are still interested please open an issue first to discuss the change. The issue should focus on the use cases that should be solved, rather than just a capability to be added.

PR #314, opened 2021-02-08
Refactor ScenarioLoader

sunyal:
- separate large section of code into multiple methods for clarity


src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -298,6 +246,128 @@ private static Config getConfig(File scenarioFile, InvocationSettings settings,
...
+                scenarioBaseDir);
+        } else if (!settings.isBazel() && !settings.isBuck() && !settings.isMaven()) {
+            scenarioDefinitions = getGradleScenarioDefinitions(
+                scenarioFile, settings, scenarioName, scenario, title, mutators, buildCount,
+                scenarioBaseDir, inspector
+            );
+        } else {
+            scenarioDefinitions = Collections.emptyList();
+        }
+        return Collections.unmodifiableList(scenarioDefinitions);
```
lptr, line 274: Can we use an `ImmutableList.Builder` here? Another option would be to return early in each clause of the `if`, and make sure that the methods called would return `ImmutableList`'s.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -298,6 +246,128 @@ private static Config getConfig(File scenarioFile, InvocationSettings settings,
...
+
+        Config executionInstructions = getExecutionInstructions(scenarioFile, settings, scenarioName,
+            scenario, MAVEN, MAVEN_KEYS);
+        List<String> targets = ConfigUtil.strings(executionInstructions, TARGETS);
+        File mavenHome = getToolHome(executionInstructions);
+        File outputDir = new File(scenarioBaseDir, "maven");
+        int warmUpCount = getWarmUpCount(settings, scenario);
+        return Arrays.asList(new MavenScenarioDefinition(scenarioName,
+            title, targets, mutators, warmUpCount,
+            buildCount, outputDir, mavenHome));
```
lptr, line 338: We could use `ImmutableList.of()` here.

lptr (commented):
Thanks for the improvements, I left a couple comments.

PR #315, opened 2021-02-08
Buildmutators filter removal

sunyal:
- remove the key filtering from the ScenarioLoader outside of the BuildMutators
- update existing BuildMutators to return NOOP when key doesn't exist in config, allowing the BuildMutators to be responsible for enabling/disabling functionality. This will allow BuildMutators to work with specific Build types (ie Bazel) or all of them depending on the configuration.


src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -212,7 +212,6 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
             String title = scenario.hasPath(TITLE) ? scenario.getString(TITLE) : null;
 
             List<BuildMutator> mutators = BUILD_MUTATOR_CONFIGURATORS.entrySet().stream()
-                .filter(entry -> scenario.hasPath(entry.getKey()))
```
sunyal, line 215: BuildMutators are responsible for returning BuildMutator.NOOP if they don't support this key during configuration. This centralizes the logic to the BuildMutator instead of in 2 places.

AlexBeggs:
@lptr any word on this?

ljacomet:
Sorry for the late reply

This looks outdated and will thus be closed. If you are still interested please open an issue first to discuss the change.

PR #320, opened 2021-02-26
Normalize stack frames with Lambda class names

big-guy:
Lambda names are auto-generated and change from build to build. This
makes it difficult to compare flamegraphs since the stack may look very
different but ultimately be exactly the same

### DIFF
--- a/src/main/java/org/gradle/profiler/flamegraph/FlameGraphSanitizer.java
+++ b/src/main/java/org/gradle/profiler/flamegraph/FlameGraphSanitizer.java
@@ -52,6 +52,8 @@ public class FlameGraphSanitizer {
 
     public static final SanitizeFunction SIMPLE_NAMES = new ToSimpleName();
 
+    public static final SanitizeFunction NORMALIZE_LAMBDA_NAMES = new NormalizeLambda();
+
     private final SanitizeFunction sanitizeFunction;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/gradle/profiler/flamegraph/FlameGraphSanitizer.java
```diff
@@ -213,6 +215,15 @@ private ChopPrefix(String stopToken) {
         }
     }
 
+    private static class NormalizeLambda extends FrameWiseSanitizeFunction {
+        @Override
+        protected String mapFrame(String frame) {
+            // Lambdas contain a name that's based on an index + timestamp at runtime and changes build-to-build.
+            // This makes comparing two builds very difficult when a lambda is in the stack
+            return frame.replaceFirst(Pattern.quote("$$Lambda$")+"[0-9/]+", "\\$\\$Lambda\\$");
```
wolfs, line 223:
```suggestion
            return frame.replaceFirst(Pattern.quote("$$Lambda$")+"[0-9./]+", "\\$\\$Lambda\\$");
```

Can you leave the `.` and the `/`? That way we also normalize Lambdas for different Java versions correctly.

wolfs (commented):
I left a comment.

wolfs (approved):
LGTM!

PR #322, opened 2021-03-01
Do not publish snapshots or alphas to SDKman

wolfs:
so we can do some internal release before we do the actual release.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -200,19 +200,21 @@ tasks.withType<SdkAnnounceVersionTask>().configureEach {
 
 tasks.register("releaseToSdkMan") {
     val versionString = project.version.toString()
-    // We only announce and set the default version for final releases
-    // A release is not final if it contains things different to numbers and dots.
-    // For example:
-    //   - 1.3: final
-    //   - 1.3.25: final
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -208,9 +209,13 @@ tasks.register("releaseToSdkMan") {
     //   - 1.3.RC5: not final
     //   - 1.3-alpha5: not final
     val isFinalRelease = Regex("""[0-9\.]*""").matchEntire(versionString) != null
-    dependsOn(tasks.withType<SdkReleaseVersionTask>())
-    if (isFinalRelease) {
-        dependsOn(tasks.withType<SdkDefaultVersionTask>())
-        dependsOn(tasks.withType<SdkAnnounceVersionTask>())
+    // We don't publish snapshots and alphas at all to SDKman.
```
blindpirate, line 212: LGTM, I think we also need to change the comment, esp. the `1.3-alpha5`.

wolfs: How about now?

blindpirate (approved):
LGTM, thanks for doing this!

felipe2i16 (approved):

PR #328, opened 2021-03-19
Support capturing multiple events with async profiler

wolfs:
Async profiler 2 allows capturing multiple events at once. It only support JFR files for multiple events. The change allows now capturing multiple events and adds a profiler `async-profiler-all` which captures cpu, alloc and lock samples at the same time.

When profiling multiple processes, we now use the infrastructure we use for the JFR profiler to capture multiple JFR files and then merge them into one flame graph for async profiler as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ You can use async profiler to profile a Gradle build using `--profile async-prof
 
 Alternatively, you can also use `--profile async-profiler-heap` to profile heap allocations, with some reasonable default settings.
 
+Finally, you can also use `--profile async-profiler-all` to profile cpu, heap allocations, and locks with some reasonable default settings.
+
 By default, an Async profiler release will be downloaded from [Github](https://github.com/jvm-profiling-tools/async-profiler/releases) and installed, if not already available.
 
 The output are flame and icicle graphs which show you the call tree and hotspots of your code.
@@ -105,8 +107,9 @@ The following options are supported and closely mimic the options of Async profi
 - `--async-profiler-event`: The event to sample, e.g. `cpu` or `alloc`. Defaults to `cpu`
 - `--async-profiler-count`: The count to use when aggregating event data. Either `samples` or `total`. `total` is especially useful for allocation profiling. Defaults to `samples`
 - `--async-profiler-interval`: The sampling interval in ns, defaults to 10_000_000 (10ms)
+- `--async-profiler-alloc-interval`: The sampling interval in bytes for allocation profiling, defaults to 10 bytes. Corresponds to the `--alloc` command line option for Async profiler.
+- `--async-profiler-lock-threshold`: lock profiling threshold in nanoseconds, defaults to 1ns.
 - `--async-profiler-stackdepth`: The maximum stack depth. Lower this if profiles with deep recursion get too large. Defaults to 2048.
-- `--async-profiler-framebuffer`: The size of the frame buffer in bytes. Defaults to 10_000_000 (~10MB)
 - `--async-profiler-system-threads`: Whether to show system threads like GC and JIT compilation in the profile. Usually makes them harder to read, but can be useful if you suspect problems in that area. Defaults to `false` 
 
 You can also use either the `ASYNC_PROFILER_HOME` environment variable or the `--async-profiler-home` command line option to point to the Async profiler install directory.
--- a/README.md
+++ b/README.md
@@ -104,13 +104,13 @@ The output are flame and icicle graphs which show you the call tree and hotspots
 
 The following options are supported and closely mimic the options of Async profiler. Have a look at its readme to find out more about each option:
 
-- `--async-profiler-event`: The event to sample, e.g. `cpu` or `alloc`. Defaults to `cpu`
-- `--async-profiler-count`: The count to use when aggregating event data. Either `samples` or `total`. `total` is especially useful for allocation profiling. Defaults to `samples`
-- `--async-profiler-interval`: The sampling interval in ns, defaults to 10_000_000 (10ms)
+- `--async-profiler-event`: The event to sample, e.g. `cpu`, `wall`, `lock` or `alloc`. Defaults to `cpu`. Multiple events can be profiled by using this parameter multiple times.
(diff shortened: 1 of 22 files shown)


src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java
```diff
@@ -71,10 +74,42 @@ public void captureSnapshot(String pid) {
...
+    }
+
+    private JfrFlameGraphGenerator.Stacks sanitizeStacks(File outputDir, String outputBaseName, File stacksFileToConvert, JfrToStacksConverter.EventType jfrEventType, JfrFlameGraphGenerator.DetailLevel level, FlameGraphSanitizer flamegraphSanitizer) {
+        String eventFileBaseName = getEventFileBaseName(outputBaseName, jfrEventType, level);
+        File sanitizedStacksFile = new File(outputDir, eventFileBaseName + "-stacks.txt");
+        flamegraphSanitizer.sanitize(stacksFileToConvert, sanitizedStacksFile);
+        return new JfrFlameGraphGenerator.Stacks(sanitizedStacksFile, jfrEventType, level, eventFileBaseName);
+    }
+
+    private String getEventFileBaseName(String outputBaseName, JfrToStacksConverter.EventType type, JfrFlameGraphGenerator.DetailLevel level) {
```
melix, line 120: This is a personal style but I tend to dislike `getXXX` methods which are not getters. What about `eventFileBaseNameFor(...)`

src/main/java/org/gradle/profiler/jfr/JfrFlameGraphGenerator.java
```diff
@@ -26,15 +26,50 @@
...
  */
-class JfrFlameGraphGenerator {
+public class JfrFlameGraphGenerator {
     private final JfrToStacksConverter stacksConverter = new JfrToStacksConverter();
     private final FlameGraphTool flameGraphGenerator = new FlameGraphTool();
 
-    public void generateGraphs(File jfrFile, String outputBaseName) {
-        if (!flameGraphGenerator.checkInstallation()) {
-            return;
+    public static class Stacks {
```
melix, line 34: This would benefit from Java 16 records :)

src/main/java/org/gradle/profiler/ScenarioSettings.java
```diff
@@ -16,4 +21,27 @@ public InvocationSettings getInvocationSettings() {
     public GradleScenarioDefinition getScenario() {
         return scenario;
     }
+
+    public File getJfrProfilerOutputLocation() {
```
melix, line 25: Getters which are not getters :/ `computeJfrProfilerOutputLocation()`?

melix (approved):
I didn't do a _functional_ review, only a code review checking for obvious mistakes and didn't see any. Good from my POV!

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java
wolfs: Renamed.

src/main/java/org/gradle/profiler/ScenarioSettings.java
wolfs: Renamed.

README.md
```diff
@@ -105,8 +107,9 @@ The following options are supported and closely mimic the options of Async profi
 - `--async-profiler-event`: The event to sample, e.g. `cpu` or `alloc`. Defaults to `cpu`
 - `--async-profiler-count`: The count to use when aggregating event data. Either `samples` or `total`. `total` is especially useful for allocation profiling. Defaults to `samples`
 - `--async-profiler-interval`: The sampling interval in ns, defaults to 10_000_000 (10ms)
+- `--async-profiler-alloc-interval`: The sampling interval in bytes for allocation profiling, defaults to 10 bytes. Corresponds to the `--alloc` command line option for Async profiler.
+- `--async-profiler-lock-threshold`: lock profiling threshold in nanoseconds, defaults to 1ns.
```
lptr, line 111: Does this correspond to some command-line argument for Async profiler?

Also, `1 ns` is more readable for me than `1ns`.

src/main/java/org/gradle/profiler/ProfilerFactory.java
```diff
@@ -37,6 +42,7 @@ public Profiler createFromOptions(OptionSet parsedOptions) {
             put("yourkit-heap", new YourKitHeapAllocationProfilerFactory());
             put("async-profiler", AsyncProfilerFactory.INSTANCE);
             put("async-profiler-heap", AsyncProfilerHeapAllocationProfilerFactory.INSTANCE);
+            put("async-profiler-all", AsyncProfilerAllEventsProfilerFactory.INSTANCE);
```
lptr, line 45: Would it make more sense to have `async-profiler` be the "all" option, and introduce `async-profiler-cpu` as the CPU-only variant?

lptr: We could potentially have a `-lock` option as well, no? I think it would also make sense to rename `-heap` to `-alloc` to match the naming in Async profiler.

lptr: Another option would be to ditch the different profilers, and add something like `--async-profiler-events=cpu,lock` with a default to `all`.

wolfs: You already can use `--async-profiler-event` multiple times and that will work. And if you use now `--async-profiler-event alloc` it will use the correct "interval" setting, maybe we should get rid of `async-profiler-heap` and `async-profiler-all` completely.

README.md
wolfs: It corresponds to `--lock`. I used `1ns` instead of `1 ns`, since we use `10ms` above. I can change that so both use the space.

wolfs: I improved the readme. Please take a look.

src/main/java/org/gradle/profiler/ProfilerFactory.java
wolfs: I added a test to show how to specify multiple events from the command line.

PR #329, opened 2021-03-22
Create differential flame graphs for async-profiler or JFR

wolfs:
When using async-profiler or JFR we can generate differential flame graphs for the stack files generated during "regular" profiling. This changes creates the differential flame graphs at the end of the profile runs, if
- Either more than one version or more than one scenario has been used,
- and JFR or async-profiler are used.


src/main/java/org/gradle/profiler/CompositeProfiler.java
```diff
@@ -122,4 +122,9 @@ public void calculateGradleArgs(List<String> gradleArgs) {
             }
         };
     }
+
+    @Override
+    public boolean isCreatesStacksFiles() {
+        return delegates.stream().anyMatch(Profiler::isCreatesStacksFiles);
```
blindpirate, line 128: What will happen if a composite profiler of profiler A and B, but profiler A supports stacks file, profiler B doesn't support? Will differential tool still work in this case?

blindpirate (commented):
Nice work! I left a few comments, it looks great overall.

src/main/resources/org/gradle/profiler/flamegraph/flamegraph.pl
```diff
@@ -107,8 +107,7 @@
 my $nametype = "Function:";     # what are the names in the data?
 my $countname = "samples";      # what are the counts in the data?
 my $colors = "hot";             # color theme
-my $bgcolor1 = "#eeeeee";       # background color gradient start
-my $bgcolor2 = "#eeeeb0";       # background color gradient stop
+my $bgcolors = "";              # background color theme
```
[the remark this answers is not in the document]
wolfs: Please check out the commit comment. I documented there where the files come from.

src/main/java/org/gradle/profiler/CompositeProfiler.java
wolfs: The differential flame graph generation works on the stack files only. So if any of the profiler generates the stacks, then there will be differential flame graphs as well. I don't think using multiple "actual" profilers does make much sense, and they would overwrite each others stack files currently.

wolfs: Though using something like a chrome trace profiler and async profiler is fine, and you'll get differential flames for the async profiler stacks.

src/main/resources/org/gradle/profiler/flamegraph/flamegraph.pl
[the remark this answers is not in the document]
blindpirate: Thanks for the clarification!

src/main/java/org/gradle/profiler/CompositeProfiler.java
blindpirate: Got it.

src/test/groovy/org/gradle/profiler/DifferentialFlameGraphIntegrationTest.groovy
```diff
@@ -0,0 +1,92 @@
...
+        and:
+        assertGraphsGenerated('upToDate')
+        assertDifferentialGraphsGenerated(['upToDate'], [latestSupportedGradleVersion, minimalSupportedGradleVersion])
+
+
+        where:
+        profiler << ["async-profiler", "jfr"]
+    }
+
+    @Requires({ !OperatingSystem.isWindows() })
```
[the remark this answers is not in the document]
wolfs: Done.

PR #333, opened 2021-06-23
Generate differential diff in Java

wolfs:
The generation is simple enough, so we can implement it in Java and don't need to rely on it for the stacks generation.


lptr (approved):
LGTM.

PR #336, opened 2021-07-07
Produce json file for Chrome trace

### LINKED ISSUE #303: chrome-trace generates file that can't be loaded into browser
After running gradle-profiler with chrome-trace I get an html-file which I cannot load into browser (blank page)

If I strip away all but the json (via grep or editor), I can go to chrome://tracing and load it.

I suspect other people do not have this problem or there would be an issue for it and that it is related to the browser (tried with Chromium & Brave), but having the json file would help me.

wolfs:
The HTML file is broken and it seems to be easier to open the trace file with Google Chrome directly.

Fixes #303


lptr (approved):
LGTM.

PR #337, opened 2021-07-13
Support configuration client Gradle VM in CliGradleClient

blindpirate:
We found sometimes we have to configure Gradle client VM, which is unsupported now. For example, build scan performance generates a huge amount of output events, when they're not flushed in time, the buffer caused an OOM in client VM:

image

image

image

This PR adds the extra `gradleOpts` configuration to make it possible.

### DIFF
--- a/src/main/java/org/gradle/profiler/CliGradleClient.java
+++ b/src/main/java/org/gradle/profiler/CliGradleClient.java
@@ -63,7 +63,7 @@ public class CliGradleClient implements GradleInvoker, GradleClient {
             String orgGradleJvmArgs = jvmArgs.isEmpty()
                 ? ""
                 : " \"-Dorg.gradle.jvmargs=" + daemonJvmArgs + "\"";
-            LinkedHashSet<String> gradleOptsSet = new LinkedHashSet<>(gradleBuildConfiguration.getGradleOpts());
+            LinkedHashSet<String> gradleOptsSet = new LinkedHashSet<>(gradleBuildConfiguration.getClientJvmArguments());
             gradleOptsSet.add("-XX:+HeapDumpOnOutOfMemoryError");
             builder.environment().put("GRADLE_OPTS", quoteJvmArguments(false, gradleOptsSet) + orgGradleJvmArgs);
         } else {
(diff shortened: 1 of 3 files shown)


src/main/java/org/gradle/profiler/GradleBuildConfiguration.java
```diff
@@ -13,13 +14,38 @@
     private final File javaHome;
     private final List<String> jvmArguments;
     private final boolean usesScanPlugin;
+    /**
+     * The GRADLE_OPTS environment variable passed to configure the client VM.
+     * See https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables
+     * <p>
+     * This is used only for --daemon. Use jvmArguments for --no-daemon.
+     */
+    private final List<String> gradleOpts;
```
wolfs, line 23: Let's call this `clientJvmArgs`. I think the Javadoc should go to the getter. We should make it clear here that:
- This is only for the CLI invoker,
- and only when the daemon is used.

It would be great to only have this on the CLI-invoker, though I don't see a good way how to do so.

src/main/java/org/gradle/profiler/CliGradleClient.java
```diff
@@ -60,11 +62,13 @@ public void runTasks(List<String> tasks, List<String> gradleArgs, List<String> j
         if (daemon) {
             String orgGradleJvmArgs = jvmArgs.isEmpty()
                 ? ""
-                : " \"-Dorg.gradle.jvmargs=" + gradleOpts + "\"";
-            builder.environment().put("GRADLE_OPTS", "-Xmx128m -Xms128m -XX:+HeapDumpOnOutOfMemoryError" + orgGradleJvmArgs);
+                : " \"-Dorg.gradle.jvmargs=" + daemonJvmArgs + "\"";
+            LinkedHashSet<String> gradleOptsSet = new LinkedHashSet<>(gradleBuildConfiguration.getGradleOpts());
+            gradleOptsSet.add("-XX:+HeapDumpOnOutOfMemoryError");
```
wolfs, line 67: I think we should add `-Xmx128m -Xms128m` if the set is empty, WDYT?

src/main/java/org/gradle/profiler/GradleBuildConfiguration.java
blindpirate: Yeah... my first attempt also goes into Cli invoker only but failed...

wolfs (approved):
LGTM!

PR #339, opened 2021-07-26
Upgradle to 7.1.1

wolfs:
This also drops support for Chrome Trace profiling on Gradle < 3.5.

Further more we upgrade the Groovy and Spock versions to the latest.


blindpirate (approved):
LGTM!

PR #340, opened 2021-07-27
Sign artifacts for publishing

wolfs:


buildSrc/src/main/kotlin/profiler.publication.gradle.kts
```diff
@@ -37,3 +38,18 @@ fun Project.gradleInternalRepositoryUrl(): URI {
     val repositoryQualifier = if (isSnapshot) "snapshots" else "releases"
     return uri("https://repo.gradle.org/gradle/ext-$repositoryQualifier-local")
 }
+
+val signArtifacts: Boolean = !System.getenv("PGP_SIGNING_KEY").isNullOrEmpty()
```
blindpirate, line 42: Shall we use `providers.environmentVariable("...").forUseAtConfigurationTime().orNull`?

blindpirate, line 42: Shall we use `providers.environmentVariable("...").forUseAtConfigurationTime().orNull`?

.teamcity/settings/GradleProfilerPublishing.kt
```diff
@@ -22,6 +24,8 @@ object GradleProfilerPublishing : BuildType({
         param("env.ORG_GRADLE_PROJECT_sdkmanToken", "%gradleprofiler.sdkman.token%")
         param("env.GRADLE_CACHE_REMOTE_USERNAME", "%gradle.cache.remote.username%")
         param("env.GRADLE_CACHE_REMOTE_PASSWORD", "%gradle.cache.remote.password%")
+        param("env.PGP_SIGNING_KEY", "%pgpSigningKey%")
```
blindpirate, line 27: Personally I like storing these credentials in TC parameters so that they can be updated without changing code, wdyt?

buildSrc/src/main/kotlin/profiler.publication.gradle.kts
```diff
@@ -37,3 +38,18 @@ fun Project.gradleInternalRepositoryUrl(): URI {
...
+tasks.withType<Sign>().configureEach { isEnabled = signArtifacts }
```
blindpirate, line 43: Nit: can we add some extra lines between these code?

blindpirate (commented):
I left some comments.

.teamcity/settings/GradleProfilerPublishing.kt
wolfs: That is a good point. I stored it for now like this, since I don't want to define the pgpSigning key in the root project, which would essentially allow all builds to access it. I did the same for native platform. Let's revisit this later.

buildSrc/src/main/kotlin/profiler.publication.gradle.kts
wolfs: Done.

buildSrc/src/main/kotlin/profiler.publication.gradle.kts
wolfs: Done.

wolfs:
@blindpirate PTAL!

.teamcity/settings/GradleProfilerPublishing.kt
blindpirate: If pgp signing key is used only in gradle-profiler, why don't define it in gradle-profiler TC project?

blindpirate (approved):

PR #341, opened 2021-07-29
Use version catalog instead of homegrown solution

wolfs:


blindpirate (approved):
LGTM! Let's see how it works.

PR #350, opened 2021-08-13
Replace reference to repo.gradle.org

britter:
Instead of depending on /repo we now use /public with a content filter
that makes sure we only query the repository for content that's
available there.


pbielicki (approved):
👍

big-guy (approved):

PR #365, opened 2021-11-03
Implement automatic Android studio sync

asodja:
Usage:
```
./gradlew installDist 
echo -e 'androidStudioSync { title = "Android Studio Sync" \n android-studio-sync { } }' > ./build/performance.scenarios

./build/install/gradle-profiler/bin/gradle-profiler --benchmark --scenario-file ./build/performance.scenarios --studio-install-dir  <Android studio installation> --project-dir <Project dir to test>
```

Output:
[profile-out.zip](https://github.com/gradle/gradle-profiler/files/7469064/profile-out.zip)

Proper integration tests are here: https://github.com/gradle/gradle-profiler/pull/366.

Right now I wait for all Android Studio background processes to finish after Gradle sync is done.

Fixes https://github.com/gradle/gradle-private/issues/3467


lptr (commented):
We are off to a good start! I have one major issue, and it's how the new steps of the sync process are measured, and how the measurements are exposed. Let's discuss these face-to-face, but here's the gist:

1) We are currently adding more measurements "around" execution time. I think "execution time" should mean the total time spent running the scenario, just as with other kinds of scenarios, and we should provide a breakdown of "Gradle execution time" + "Tooling API agent time" (whatever this is) + "Android Studio doing its thing".

2) I don't like how we now have a generic `BuildActionResult` type that contains very Studio-specific things. Instead, we should allow `BuildAction` to return a generic result, and extend `GradleBuildInvocationResult` to have a `StudioSyncInvociationResult` with the necessary extra measurements.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Client.java
```diff
@@ -1,16 +1,25 @@
...
+import java.util.concurrent.Executors;
+import java.util.function.Consumer;
 
 /**
  * A singleton that runs inside a client process to communicate with the controller process.
  */
-public class Client {
-    public static final Client INSTANCE = new Client();
+public enum Client {
+    INSTANCE;
```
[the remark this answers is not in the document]
asodja: It's the best singleton in Java :). On the other hand I don't think this should be a singleton anyway, I will check if it's simple to remove singleton nature.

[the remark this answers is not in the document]
lptr: Alright. Let's do that in a separate PR, and focus on delivering this one first.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/messages/Message.java
```diff
@@ -0,0 +1,4 @@
+package org.gradle.profiler.client.protocol.messages;
+
+public class Message {
+}
```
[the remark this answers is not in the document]
asodja: Yes most likely, I just moved it to messages. I will convert it to interface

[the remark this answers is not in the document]
asodja: done

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/ConnectionParameters.java
```diff
@@ -1,5 +1,7 @@
 package org.gradle.profiler.client.protocol;
 
+import org.gradle.profiler.client.protocol.messages.Message;
```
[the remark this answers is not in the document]
asodja: done

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/SyncParameters.java
```diff
@@ -1,5 +1,7 @@
 package org.gradle.profiler.client.protocol;
 
+import org.gradle.profiler.client.protocol.messages.Message;
```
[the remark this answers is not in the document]
asodja: do

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/messages/Message.java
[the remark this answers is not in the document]
lptr: Thanks

buildSrc/build.gradle.kts
```diff
@@ -4,4 +4,5 @@ plugins {
 
 repositories {
     mavenCentral()
+    gradlePluginPortal()
```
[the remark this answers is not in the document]
asodja: We need it for intellij-gradle-plugin

settings.gradle.kts
```diff
@@ -11,6 +11,7 @@ include("heap-dump")
 include("client-protocol")
 include("instrumentation-support")
 include("studio-agent")
+include("studio-plugin")
```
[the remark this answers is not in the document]
asodja: studio-plugin is an Android Studio plugin, and studio-agent is a java agent that intruments `DefaultGradleConnector` so measure Gradle time. So it's a different thing and I think it's good that they are separated. 

Maybe we could just name studio-agent differently

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -38,9 +50,9 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
         }
         System.out.println("* Main class: " + launchConfiguration.getMainClass());
 
-        server = new Server("agent");
-        studioProcess = startStudio(launchConfiguration, studioInstallDir, invocationSettings, server);
-        agentConnection = server.waitForIncoming(Duration.ofMinutes(1));
+        studioProcess = startStudio(launchConfiguration, studioInstallDir, invocationSettings, agentServer);
+        pluginAgentConnection = pluginServer.waitForIncoming(Duration.ofMinutes(1));
+        agentConnection = agentServer.waitForIncoming(Duration.ofMinutes(1));
```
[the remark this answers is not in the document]
asodja: good point, renamed

src/main/java/org/gradle/profiler/studio/LauncherConfigurationParser.java
```diff
@@ -17,7 +23,7 @@
 import java.util.stream.Collectors;
 
 public class LauncherConfigurationParser {
-    public LaunchConfiguration calculate(Path studioInstallDir) {
+    public LaunchConfiguration calculate(Path studioInstallDir, String pluginPort) {
```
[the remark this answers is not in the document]
asodja: Changed to int

src/main/java/org/gradle/profiler/BuildAction.java
```diff
@@ -44,5 +45,39 @@ public Duration run(GradleClient gradleClient, List<String> gradleArgs, List<Str
...
+        public Duration getGradleToolingAgentExecutionTime() {
+            return gradleToolingAgentExecutionTime;
+        }
+
+        public Duration getStudioExecutionTime() {
+            return studioExecutionTime;
+        }
+
+        public static BuildActionResult of(Duration executionTime) {
+            return new BuildActionResult(executionTime, null, null);
```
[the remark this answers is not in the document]
asodja: I went with Duration.ZERo for now and `withExecutionTimeOnly() and withIdeTimings()`. I will try to refactor this later as discussed in person.

src/main/java/org/gradle/profiler/BuildAction.java
```diff
@@ -44,5 +45,39 @@ public Duration run(GradleClient gradleClient, List<String> gradleArgs, List<Str
...
+            return executionTime;
+        }
+
+        public Duration getGradleToolingAgentExecutionTime() {
+            return gradleToolingAgentExecutionTime;
+        }
+
+        public Duration getStudioExecutionTime() {
+            return studioExecutionTime;
+        }
```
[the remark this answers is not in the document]
asodja: Renames to `getIdeExecutionTime()`

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -58,34 +70,40 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
...
         System.out.println("* Using command line: " + commandLine);
 
+        new StudioPluginInstaller().installPlugin(launchConfiguration);
         return new CommandExec().inDir(studioInstallDir.toFile()).start(commandLine);
     }
 
     @Override
     public void close() throws IOException {
-        System.out.println("* PLEASE STOP ANDROID STUDIO....");
+        System.out.println("* STOPPING ANDROID STUDIO....");
```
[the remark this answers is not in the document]
asodja: done

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -58,34 +70,40 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
...
-            System.out.println("* PLEASE RUN SYNC IN ANDROID STUDIO (once it has finished starting up)....");
-            hasRun = true;
-        } else {
-            System.out.println("* PLEASE RUN SYNC IN ANDROID STUDIO....");
-        }
-
-        // Use a long time out because it can take quite some time between the tapi action completing and studio finishing the sync
-        SyncStarted started = agentConnection.receiveSyncStarted(Duration.ofMinutes(10));
+    public BuildActionResult sync(List<String> gradleArgs, List<String> jvmArgs) {
+        System.out.println("* PLEASE WAIT TO RUN SYNC IN ANDROID STUDIO...");
```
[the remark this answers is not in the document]
asodja: done

src/main/java/org/gradle/profiler/studio/LauncherConfigurationParser.java
```diff
@@ -30,12 +36,27 @@ public LaunchConfiguration calculate(Path studioInstallDir) {
...
         Path asmJar = GradleInstrumentation.unpackPlugin("asm").toPath();
         Path supportJar = GradleInstrumentation.unpackPlugin("instrumentation-support").toPath();
         Path protocolJar = GradleInstrumentation.unpackPlugin("client-protocol").toPath();
-        return new LaunchConfiguration(javaCommand, classPath, systemProperties, mainClass, agentJar, supportJar, Arrays.asList(asmJar, protocolJar));
+        Path studioPlugin = GradleInstrumentation.unpackPlugin("studio-plugin").toPath();
+        Path studioPluginsDir = newPluginTempDir();
+
+        systemProperties.put("idea.plugins.path", studioPluginsDir.toAbsolutePath().toString());
+        systemProperties.put("idea.log.path", studioPluginsDir.toAbsolutePath().toString());
+        return new LaunchConfiguration(javaCommand, classPath, systemProperties, mainClass, agentJar, supportJar, Arrays.asList(asmJar, protocolJar), Arrays.asList(studioPlugin, protocolJar), studioPluginsDir);
```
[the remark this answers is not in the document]
asodja: done

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Installs the Android Studio plugin into the plugins directory.
+ */
+public class StudioPluginInstaller {
+
+    /**
+     * Note: if you change that, and you load plugins always from the same location,
+     * then you will have also "old plugin" on the classpath, and it might override your new classes.
+     */
+    private static final String PLUGIN_INSTALL_FOLDER_NAME = "gradle-profiler-studio-plugin";
```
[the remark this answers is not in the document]
asodja: Changed "that" to "this constant" and "old plugin" to "a plugin installed before"

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/messages/StudioSyncRequestCompleted.java
```diff
@@ -0,0 +1,32 @@
+package org.gradle.profiler.client.protocol.messages;
+
+public class StudioSyncRequestCompleted extends Message {
```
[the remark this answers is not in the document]
asodja: Good point. Renamed

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Serializer.java
```diff
@@ -61,6 +65,29 @@ public void send(ConnectionParameters message) {
         }
     }
 
+    public void send(StudioRequest message) {
+        try {
+            connection.writeByte((byte) 5);
```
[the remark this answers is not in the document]
asodja: I did this a bit differently. I now have MessageRegistry with all info about messages and `MessageReaderWriter` that have a code to read or write messages.

Let me know what you think

lptr:
One thing unrelated to the code changes but related to the functionality is that I think we should rename `execution` to `total execution time` in the `Sample`s used in the HTML report for consistentcy:

image

settings.gradle.kts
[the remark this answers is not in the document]
lptr: That's not what I meant, sorry. If they have to be different things (an agent vs an IntelliJ plugin), then it's fine to keep them in the projects they are now, separated.

My question was about whether whatever the agent was doing today could also be implemented by an IntelliJ plugin. If so, then I'd favor dropping the agent and moving the code to the plugin to keep things simple. If not, then we need to keep the agent.

src/main/java/org/gradle/profiler/BuildUnderTestInvoker.java
```diff
@@ -80,10 +80,10 @@ public GradleBuildInvocationResult run(BuildContext buildContext, BuildStep buil
             return new GradleBuildInvocationResult(
                 buildContext,
                 buildActionResult.getExecutionTime(),
+                buildActionResult.getGradleToolingAgentExecutionTime(),
+                buildActionResult.getIdeExecutionTime(),
```
lptr, line 84: Actually, I think these should be `null`s here (and hence we should go for `Optional` in `BuildActionResult`). Otherwise even non-IDE benchmarks will have these measurements, no?

src/main/java/org/gradle/profiler/RunTasksAction.java
```diff
@@ -1,5 +1,6 @@
 package org.gradle.profiler;
 
+import java.time.Duration;
```
lptr, line 3: This looks strange, why was this added?

src/main/java/org/gradle/profiler/studio/LauncherConfigurationParser.java
```diff
@@ -46,9 +46,12 @@ public LaunchConfiguration calculate(Path studioInstallDir, String pluginPort) {
...
         Path studioPluginsDir = newPluginTempDir();
 
-        systemProperties.put("idea.plugins.path", studioPluginsDir.toAbsolutePath().toString());
-        systemProperties.put("idea.log.path", studioPluginsDir.toAbsolutePath().toString());
-        return new LaunchConfiguration(javaCommand, classPath, systemProperties, mainClass, agentJar, supportJar, Arrays.asList(asmJar, protocolJar), Arrays.asList(studioPlugin, protocolJar), studioPluginsDir);
+        String studioPluginsDirPath = studioPluginsDir.toAbsolutePath().toString();
+        List<Path> sharedJars = Arrays.asList(asmJar, protocolJar);
+        List<Path> studioPluginJars = Arrays.asList(studioPlugin, protocolJar);
+        systemProperties.put("idea.plugins.path", studioPluginsDirPath);
+        systemProperties.put("idea.log.path", studioPluginsDirPath);
```
lptr, line 53: Readability: is there a reason not to swap line 50-51 with line 52-53?

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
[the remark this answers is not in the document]
lptr: I wont pretend that I understand what's going on here :) Why is it important to document what happens if somebody changes this constant? Why would anyone want to touch this?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Client.java
```diff
@@ -57,15 +55,15 @@ public void send(StudioSyncRequestCompleted message) {
         }
     }
```
lptr, line 56: Why do we have these separate overloads? Couldn't we just have a `send(Message)` method here? Or have a common super-interface for `Message`s that can be passed in here?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageReaderWriter.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.gradle.profiler.client.protocol.Connection;
+import org.gradle.profiler.client.protocol.messages.Message;
+
+import java.io.IOException;
+
+/**
+ * Message reader writer that reads and writes messages from and to a {@link Connection}.
+ */
+public interface MessageReaderWriter {
```
lptr, line 11: I think this is what is called a `MessageSerializer`. The current `MessageSerializer` is more like a `ProtocolHandler` or something. I might even rename the current `MessageRegistry` enum to `MessageSerializer`, and add the `readFrom()` and `writeTo()` methods there.

I'd also add maps for both lookups instead of streaming through `values()` to make the lookup more efficient.

buildSrc/build.gradle.kts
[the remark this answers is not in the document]
wolfs: I don't understand how the `repositories` declaration in `buildSrc` has an effect the plugins available in the plugins block in the `studio-plugin` subproject. Shouldn't that go into `pluginManagement` in the top-level `settings.gradle.kts`?

README.md
```diff
@@ -33,6 +33,8 @@ First, build and install the `gradle-profiler` app using:
 
 This will install the executable into `./build/install/gradle-profiler/bin`. The examples below assume that you add this location to your PATH or create a `gradle-profiler` alias for it.
 
+NOTE: You have to use Java 11 or later to build this project.
```
wolfs, line 36: We could add a check in `settings.gradle.kts` to print out a nice error message, WDYT?

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
```diff
@@ -12,9 +12,48 @@
...
+        @Nullable Duration garbageCollectionTime,
+        @Nullable Duration timeToTaskExecution,
+        Map<String, Duration> cumulativeBuildOperationTimes,
+        String daemonPid
+    ) {
+        super(buildContext, executionTime);
+        this.garbageCollectionTime = garbageCollectionTime;
+        this.timeToTaskExecution = timeToTaskExecution;
+        this.gradleToolingAgentExecutionTime = gradleToolingAgentExecutionTime;
+        this.studioExecutionTime = ideExecutionTime;
```
wolfs, line 34:
```suggestion
        this.ideExecutionTime = ideExecutionTime;
```

This field should probably be renamed as well.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -38,10 +48,10 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
         }
         System.out.println("* Main class: " + launchConfiguration.getMainClass());
 
-        server = new Server("agent");
-        studioProcess = startStudio(launchConfiguration, studioInstallDir, invocationSettings, server);
-        agentConnection = server.waitForIncoming(Duration.ofMinutes(1));
-        agentConnection.send(new ConnectionParameters(buildConfiguration.getGradleHome()));
+        studioProcess = startStudio(launchConfiguration, studioInstallDir, invocationSettings, studioAgentServer);
+        studioPluginConnection = studioPluginServer.waitForIncoming(Duration.ofMinutes(1));
+        studioAgentConnection = studioAgentServer.waitForIncoming(Duration.ofMinutes(1));
```
wolfs, line 53: Should we extract the `Duration.ofMinutes(1)` into some constant in this class? Do you think this needs to be parameterized?

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -58,34 +68,40 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
...
     @Override
     public void close() throws IOException {
-        System.out.println("* PLEASE STOP ANDROID STUDIO....");
+        System.out.println("* Stopping Android Studio....");
+        studioPluginConnection.send(new StudioRequest(EXIT));
         studioProcess.waitForSuccess();
         System.out.println("* Android Studio stopped.");
-
-        server.close();
+        studioPluginServer.close();
```
wolfs, line 84: Should the closing be done in a `finally` here or via some composite closeable to make sure all the servers have been stopped?

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -58,34 +68,40 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
...
-        // Use a long time out because it can take quite some time between the tapi action completing and studio finishing the sync
-        SyncStarted started = agentConnection.receiveSyncStarted(Duration.ofMinutes(10));
-        agentConnection.send(new SyncParameters(gradleArgs, jvmArgs));
+    public BuildActionResult sync(List<String> gradleArgs, List<String> jvmArgs) {
+        System.out.println("* Running sync in Android Studio...");
+        studioPluginConnection.send(new StudioRequest(SYNC));
+        System.out.println("* Sent sync request");
+        // Use a long time out because it can take quite some time
+        // between the tapi action completing and studio finishing the sync
+        studioAgentConnection.receiveSyncStarted(Duration.ofMinutes(10));
```
wolfs, line 94: I think it would be good to extract all the wait times into constants in this class, to have an overview what we are waiting for and how long.

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
```diff
@@ -0,0 +1,54 @@
...
+        installPluginToDirectory(
+            configuration.getStudioPluginsDir(),
+            configuration.getStudioPluginJars()
+        );
+    }
+
+    private static void installPluginToDirectory(Path directory, List<Path> jars) {
+        try {
+            for (Path jar : jars) {
+                String jarName = jar.getFileName().toString().endsWith(".jar")
```
wolfs, line 36: Why do the jars we pass in here sometimes not end with `.jar`?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Client.java
```diff
@@ -1,19 +1,26 @@
 package org.gradle.profiler.client.protocol;
 
+import org.gradle.profiler.client.protocol.messages.*;
```
wolfs, line 3: Can we not do the `*` import here but individual imports instead?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageSerializer.java
```diff
@@ -0,0 +1,82 @@
...
+            MessageReaderWriter readerWriter = MessageRegistry.findMessageReaderWriter(message.getClass());
+            readerWriter.writeTo(connection, message);
+            connection.flush();
+        } catch (IOException e) {
+            throw new RuntimeException(String.format("Could not write to %s.", peerName), e);
+        }
+    }
+
+    public <T extends Message> T receive(Class<T> type, Duration timeout) {
+        BlockingQueue<Object> queue = new LinkedBlockingQueue<>();
```
wolfs, line 36: Wouldn't it be simpler to submit a `Callable` to an `ExecuterService` here instead of starting a new thread and using a blocking queue?

subprojects/instrumentation-support/src/main/java/org/gradle/profiler/studio/instrumented/Interceptor.java
```diff
@@ -38,11 +38,11 @@ public static void onConnect(DefaultGradleConnector connector) {
...
         int id = COUNTER.incrementAndGet();
 
-        SyncParameters syncParameters;
+        GradleInvocationParameters syncParameters;
         try {
             Client client = Client.INSTANCE;
-            client.send(new SyncStarted(id));
-            syncParameters = client.receiveSyncParameters(Duration.ofSeconds(60));
+            client.send(new GradleInvocationStarted(id));
+            syncParameters = client.receiveSyncParameters(Duration.ofSeconds(300));
```
wolfs, line 45: Another wait time which could be extracted to somewhere.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
```diff
@@ -0,0 +1,19 @@
...
+import org.gradle.profiler.studio.plugin.client.GradleProfilerClient;
+
+public class GradleProfilerProjectManagerListener implements ProjectManagerListener {
+
+    private static final Logger LOG = Logger.getInstance(GradleProfilerProjectManagerListener.class);
+
+    @Override
+    public void projectOpened(com.intellij.openapi.project.Project project) {
+        LOG.info("Project opened");
+        TrustedProjects.setTrusted(project, true);
```
wolfs, line 15: You are setting trusted here, and you also pass the system property to trust all projects. Can we remove one of the two things?

subprojects/studio-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,22 @@
+<idea-plugin>
```
wolfs, line 1: Does this file need to be checked in? It seems like the build file also adds the plugin dependencies.

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+import java.time.Duration
+
+import static org.gradle.profiler.client.protocol.messages.StudioRequest.StudioRequestType.EXIT
+import static org.gradle.profiler.client.protocol.messages.StudioRequest.StudioRequestType.SYNC
+import static org.gradle.profiler.client.protocol.messages.StudioSyncRequestCompleted.StudioSyncRequestResult.SUCCEEDED
+import static org.gradle.profiler.studio.plugin.client.GradleProfilerClient.PROFILER_PORT_PROPERTY
+
+/**
+ * Note: This class uses Junit3 so every test should start with "test"
```
wolfs, line 17: Why JUnit 3? I wonder if `BasePlatformTestCase` also works with JUnit 4, the Javadoc on `UsefulTestCase` seems to claim that. Though I suppose it is not so important.

src/main/java/org/gradle/profiler/studio/LauncherConfigurationParser.java
```diff
@@ -30,12 +36,30 @@ public LaunchConfiguration calculate(Path studioInstallDir) {
         List<Path> classPath = Arrays.stream(jvmOptions.string("ClassPath").split(":")).map(s -> FileSystems.getDefault().getPath(s.replace("$APP_PACKAGE", actualInstallDir.toString()))).collect(Collectors.toList());
         String mainClass = jvmOptions.string("MainClass");
         Map<String, String> systemProperties = mapValues(jvmOptions.dict("Properties").toMap(), v -> v.replace("$APP_PACKAGE", actualInstallDir.toString()));
-        Path javaCommand = actualInstallDir.resolve("Contents/jre/jdk/Contents/Home/bin/java");
+        systemProperties.put("gradle.profiler.port", String.valueOf(studioPluginPort));
+        systemProperties.put("idea.trust.all.projects", "true");
+        Path javaCommand = actualInstallDir.resolve("Contents/jre/Contents/Home/bin/java");
```
wolfs, line 41: This path has already been fixed on master, so there is a conflict now.

wolfs (approved):
Hey, that looks pretty good! I left some comments. I am fine with merging as is now and addressing my comments afterwards.

settings.gradle.kts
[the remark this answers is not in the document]
asodja: So I think it's good that we keep the agent since the agent attaches directly to Gradle and measures Gradle time more accurately. The plugin calls Android Studio API that does some extra work, so the time we can measure there is not exactly just the time of the Gradle.

README.md
asodja: good idea, will add

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
asodja: You are right, it works with Junit4. I fixed that in: https://github.com/gradle/gradle-profiler/pull/366

subprojects/studio-plugin/src/main/resources/META-INF/plugin.xml
asodja: This file is required since this is a plugin descriptor and is packed into the jar and later read by IntelliJ. Dependencies are unfortunately duplicated. Added a comment that dependency should be added to intellij extension

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
asodja: Hawk eyes. Yes, I will remove the property. The thing is that property was added in one of the latest IntelliJ, but is not supported for example in 2021.1 what is used by Android Studio Bumblebee atm.

wolfs: It is also fine to leave both things, as long as there is a comment somewhere.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
asodja: renamed

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageSerializer.java
asodja: This is here in diff just to the renaming. But I refactored it anyway, it's much nicer! Thanks

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Client.java
asodja: Fixed. Also fixed my Intellij settings

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
asodja: GradleInstrumentation.unpackPlugin creates a file that doesn't end with `.jar`. I think this is a bug. I fixed it and add precondition here

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
asodja: Done

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
asodja: Yeah, it makes sense

subprojects/instrumentation-support/src/main/java/org/gradle/profiler/studio/instrumented/Interceptor.java
asodja: done

asodja: Also I don't know why I changed that to 300 so I reverted it back to 60.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
asodja: done

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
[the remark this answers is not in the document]
asodja: Removed comment and added shutdown hook.

src/main/java/org/gradle/profiler/BuildUnderTestInvoker.java
asodja: Makes sense for the consistency, fixed.  Note that for non-IDE benchmarks we don't read these values so these values are not there.

src/main/java/org/gradle/profiler/RunTasksAction.java
asodja: no checkstyle :)

src/main/java/org/gradle/profiler/studio/LauncherConfigurationParser.java
asodja: fixed

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Client.java
asodja: I believe so you can't send wrong messages the other side. But I wil remove that.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageReaderWriter.java
asodja: Naming is hard...  Renamed, and added changes

src/main/java/org/gradle/profiler/studio/LauncherConfigurationParser.java
asodja: rebased

settings.gradle.kts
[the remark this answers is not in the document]
lptr: 👍

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/Client.java
lptr: We should commit those settings to Git.

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
```diff
@@ -33,9 +34,8 @@ public void installPlugin(LaunchConfiguration configuration) {
     private static void installPluginToDirectory(Path directory, List<Path> jars) {
         try {
             for (Path jar : jars) {
-                String jarName = jar.getFileName().toString().endsWith(".jar")
-                    ? jar.getFileName().toString()
-                    : jar.getFileName() + ".jar";
+                String jarName = jar.getFileName().toString();
+                checkArgument(jarName.endsWith(".jar"), "Expected jar file: %s to end with .jar", jar);
```
lptr, line 38: Do we really need to check this? Seems like defensive programming.

subprojects/studio-plugin/build.gradle.kts
```diff
@@ -19,6 +19,9 @@ java {
 intellij {
     pluginName.set("gradle-profiler-studio-plugin")
     version.set("2021.1.1")
+    // Don't override "since-build" and "until-build" properties in plugin.xml,
+    // so we can don't need to update plugin to use it also on future IntelliJ versions.
```
lptr, line 23: Typo.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -76,13 +95,27 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
...
-        studioAgentServer.close();
+    public void close() {
+        try {
+            System.out.println("* Stopping Android Studio....");
+            studioPluginConnection.send(new StudioRequest(EXIT));
+            studioProcess.waitForSuccess(STUDIO_EXIT_TIMEOUT_SECONDS, SECONDS);
+            System.out.println("* Android Studio stopped.");
+        } catch (Exception e) {
+            e.printStackTrace();
+            System.out.println("* Android Studio did not stopped successfully, you will have to close it manually.");
```
lptr, line 106: Typo: "did not stop", or better: "did not finish successfully".

src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
```diff
@@ -30,17 +30,19 @@ public GradleScenarioInvoker(DaemonControl daemonControl, PidInstrumentation pid
     @Override
     public List<Sample<? super GradleBuildInvocationResult>> samplesFor(InvocationSettings settings, GradleScenarioDefinition scenario) {
         ImmutableList.Builder<Sample<? super GradleBuildInvocationResult>> builder = ImmutableList.builder();
-        builder.add(BuildInvocationResult.EXECUTION_TIME);
+        if (scenario.isAndroidStudioSync()) {
+            builder.add(GradleBuildInvocationResult.TOTAL_EXECUTION_TIME);
+            builder.add(GradleBuildInvocationResult.GRADLE_TOOLING_AGENT_EXECUTION_TIME);
+            builder.add(GradleBuildInvocationResult.STUDIO_EXECUTION_TIME);
+        } else {
+            builder.add(BuildInvocationResult.EXECUTION_TIME);
```
lptr, line 38: Why do we now have `EXECUTION_TIME` and `TOTAL_EXECUTION_TIME` as well? I think we should only have `TOTAL_EXECUTION_TIME` now.

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
```diff
@@ -16,22 +17,21 @@
...
         // Delete previous directory in case the plugin directory is reused
-        deletePluginDirectory(configuration.getStudioPluginsDir());
+        File pluginsDir = configuration.getStudioPluginsDir().toFile();
+        File pluginInstallDir = new File(pluginsDir, PLUGIN_INSTALL_FOLDER_NAME);
+        deletePluginDirectory(pluginInstallDir);
         installPluginToDirectory(
             configuration.getStudioPluginsDir(),
             configuration.getStudioPluginJars()
         );
+        registerDeletePluginShutdownHook(pluginInstallDir);
```
lptr, line 31: A shutdown hook seems like a heavy-handed approach. Suppose you want to run the profiler embedded in a JVM that never exits, then this won't be called at all. Can we remove the directory more explicitly?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageSerializer.java
```diff
@@ -1,57 +1,164 @@
...
-            throw new RuntimeException(e);
+    /**
+     * Writes a message to the given connection.
+     */
+    public void writeTo(Connection connection, Message message) throws IOException {
+        connection.writeByte(messageId);
+        doWriteTo(connection, message);
+    }
+
+    public static MessageSerializer findMessageSerializer(Class<? extends Message> messageClass) {
```
lptr, line 148: I'd say these methods should be called `get...()` — we typically use `find...()` for getters that are not expected to always find a single instance.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageSerializer.java
```diff
@@ -1,57 +1,164 @@
...
+        MessageSerializer serializer = SERIALIZERS_BY_CLASS.get(messageClass);
+        if (serializer == null) {
+            throw new RuntimeException("No message reader writer is declared in the " + MessageSerializer.class + " for message class: " + messageClass);
```
lptr, line 151: It would be better to throw `IllegalArgumentException` here.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageSerializer.java
asodja: fixed

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageSerializer.java
asodja: fixed

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
asodja: Good point. Moved to the close method of the StudioGradleClient

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
asodja: Changed to "did not finish successfully"

subprojects/studio-plugin/build.gradle.kts
asodja: fixed typo

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
asodja: I would leave it since Android Studio doesn't detect the plugins if it doesn't end with .jar. And it's really annoying to debug this. I was there, debugging this once :)

src/main/java/org/gradle/profiler/GradleScenarioInvoker.java
asodja: Changin execution time also affects others. I remove `TOTAL_EXECUTION_TIME`, 🤞  that we don't break any automation that reads csv.

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
lptr: Yes, but weren't you there because of that bug causing the file to have no `.jar` extension? Now that it's fixed it's unlikely anyone will have a similar problem. The parameter is called `jar` and everything, it's not like it's unexpected that it has to be called `something.jar` :)

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -113,7 +116,7 @@ public void close() {
     private void tryClose(Closeable closeable) {
         try {
             closeable.close();
-        } catch (IOException e) {
+        } catch (UncheckedIOException | IOException e) {
             e.printStackTrace();
         }
```
lptr, line 121: This is getting way too complicated. Why do we need to catch an unchecked exception here? And if we want to do it properly, we should print its cause anyway. But why is this so complicated, I wonder?

Not that I know a better solution off the top of my head...

lptr: BTW, I'm ok with merging this like it is, but it irks me that I have no better solution.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -103,8 +105,9 @@ public void close() {
             System.out.println("* Android Studio stopped.");
         } catch (Exception e) {
             e.printStackTrace();
-            System.out.println("* Android Studio did not stopped successfully, you will have to close it manually.");
+            System.out.println("* Android Studio did not finish successfully, you will have to close it manually.");
         } finally {
+            tryClose(studioPluginInstaller::uninstallPlugin);
             tryClose(studioPluginServer);
             tryClose(studioAgentServer);
```
lptr, line 112: BTW, we should use try-with-resources (what's a better name for this construct?) here:

https://github.com/gradle/gradle-profiler/blob/c81fc08884fe24a1403d04c390a5ed524ebb0df7/src/main/java/org/gradle/profiler/GradleScenarioInvoker.java#L79

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -103,8 +105,9 @@ public void close() {
             System.out.println("* Android Studio stopped.");
         } catch (Exception e) {
             e.printStackTrace();
-            System.out.println("* Android Studio did not stopped successfully, you will have to close it manually.");
+            System.out.println("* Android Studio did not finish successfully, you will have to close it manually.");
         } finally {
+            tryClose(studioPluginInstaller::uninstallPlugin);
```
lptr, line 110: What arcane magic is this? You are "auto-boxing" that method reference in a `Closable` and executing it as via `tryClose()` to..? Why?! :)

I never knew you could do this... I'm not sure how I feel. Conflicted..?!

lptr (approved):
I left a few more comments, but in general I think this is GTM.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
asodja: fixed that magic with try-with-resources, thank you for the idea

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
asodja: Fixed that with try-with-resources. That solution was not nice, I agree

src/main/java/org/gradle/profiler/studio/plugin/StudioPluginInstaller.java
asodja: Makes sense, removed

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
```diff
@@ -39,36 +78,28 @@ public Duration extractFrom(GradleBuildInvocationResult result) {
...
-        Map<String, Duration> cumulativeBuildOperationTimes,
-        String daemonPid
-    ) {
-        super(buildContext, executionTime);
-        this.garbageCollectionTime = garbageCollectionTime;
-        this.timeToTaskExecution = timeToTaskExecution;
-        this.cumulativeBuildOperationTimes = cumulativeBuildOperationTimes;
-        this.daemonPid = daemonPid;
-    }
+    public static final Sample<GradleBuildInvocationResult> STUDIO_EXECUTION_TIME = new Sample<GradleBuildInvocationResult>() {
```
wolfs, line 93:
```suggestion
    public static final Sample<GradleBuildInvocationResult> IDE_EXECUTION_TIME = new Sample<GradleBuildInvocationResult>() {
```

Should this also be renamed? As well as the `getName()` method?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageProtocolHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class MessageProtocolHandler {
+    private final String peerName;
+    private final Connection connection;
+    private final ExecutorService executorService;
+
+    public MessageProtocolHandler(String peerName, Connection connection) {
+        this.peerName = peerName;
+        this.connection = connection;
+        this.executorService = Executors.newCachedThreadPool();
```
wolfs, line 19: I suppose this class should be closable as well and shutdown the `executorService` on `close`.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageProtocolHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public <T extends Message> T receive(Class<T> type, Duration timeout) {
+        try {
+            Future<Object> future = executorService.submit((Callable<Object>) this::receive);
+            Object result = future.get(timeout.toMillis(), TimeUnit.MILLISECONDS);
+            return type.cast(result);
+        } catch (TimeoutException e) {
+            throw new IllegalStateException(String.format("Timeout waiting to receive message from %s.", peerName));
+        } catch (ExecutionException | InterruptedException e) {
+            throw new RuntimeException(e);
```
wolfs, line 40: Not so important, though you could get the cause of the `ExecutionException` here instead of using the execution exception directly.

subprojects/studio-plugin/src/main/resources/META-INF/plugin.xml
wolfs: OK.

wolfs (commented):
I left some more smaller comments.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageProtocolHandler.java
asodja: Good point, but I would like to avoid Closeables, since it's easy to forget to close it. I will use `CompletableFuture.supplyAsync` that uses common thread pool. Should do the trick in that case.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
asodja: Renamed.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -98,26 +96,16 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
 
     @Override
     public void close() {
-        try {
+        try (Server studioPluginServer = this.studioPluginServer;
+             Server studioAgentServer = this.studioAgentServer;
+             Closeable uninstallPlugin = studioPluginInstaller::uninstallPlugin) {
```
lptr, line 101: This is a lot better, though I feel this cast to `Closable` is still weird. It's at least a bit less magical than when you passed it directly as a parameter. Anyway, I think it's good now.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageProtocolHandler.java
```diff
@@ -31,7 +32,7 @@ public void send(Message message) {
 
     public <T extends Message> T receive(Class<T> type, Duration timeout) {
         try {
-            Future<Object> future = executorService.submit((Callable<Object>) this::receive);
+            CompletableFuture<Object> future = CompletableFuture.supplyAsync(this::receive);
```
lptr, line 35: Why change this if we don't use the more specific type?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/serialization/MessageProtocolHandler.java
lptr: Is this a good idea, I wonder? I never used the common thread pool, but my expectation would be is that we have no control over it, and if someone else fills it up with requests, our measurements might suffer. For this reason a private executor service sounds better for me.

lptr (approved):
I'm a bit wary about the common thread pool, otherwise LGTM.

PR #367, opened 2021-11-08
CI build now always uses Java11 as JAVA_HOME

asodja:
… so it works with studio-plugin project

### DIFF
--- a/.teamcity/settings/GradleProfilerTest.kt
+++ b/.teamcity/settings/GradleProfilerTest.kt
@@ -10,14 +10,20 @@ open class GradleProfilerTest(os: Os, javaVersion: JavaVersion) : BuildType({
     gradleProfilerVcs()
 
     params {
-        javaHome(os, javaVersion)
+        // Java home must always use Java11
+        // since intellij-gradle-plugin is not compatible with Java8
+        javaHome(os, JavaVersion.OPENJDK_11)
     }
 
     steps {
         gradle {
             tasks = "clean :test"
             buildFile = ""
-            gradleParams = "-s --build-cache ${toolchainConfiguration(os)}"
+            gradleParams = "-s" +
+                " --build-cache ${toolchainConfiguration(os)}" +
+                " -PtestJavaVersion=${javaVersion.majorVersion}" +
+                " -PtestJavaVendor=${javaVersion.vendor}"
+            param("org.jfrog.artifactory.selectedDeployableServer.defaultModuleVersionConfiguration", "GLOBAL")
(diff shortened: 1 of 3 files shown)


wolfs (approved):
The changes look good. I suppose we set the toolchain for compilation in each of the subprojects, right? So we don't depend on the Java home there and we only depend on the Java home for selecting the Gradle version for running the build.

asodja:
Yes, for a compilation we have it on every subproject.

.teamcity/settings/GradleProfilerTest.kt
```diff
@@ -10,14 +10,20 @@ open class GradleProfilerTest(os: Os, javaVersion: JavaVersion) : BuildType({
...
     steps {
         gradle {
             tasks = "clean :test"
             buildFile = ""
-            gradleParams = "-s --build-cache ${toolchainConfiguration(os)}"
+            gradleParams = "-s" +
+                " --build-cache ${toolchainConfiguration(os)}" +
+                " -PtestJavaVersion=${javaVersion.majorVersion}" +
+                " -PtestJavaVendor=${javaVersion.vendor}"
+            param("org.jfrog.artifactory.selectedDeployableServer.defaultModuleVersionConfiguration", "GLOBAL")
```
wolfs, line 26: Why is this parameter here twice now?

build.gradle.kts
```diff
@@ -99,10 +99,24 @@ tasks.processResources {
...
+        providers.gradleProperty("testJavaVendor").map {
+            when (it.toLowerCase()) {
+                "oracle" -> vendor.set(JvmVendorSpec.ORACLE)
+                "openjdk" -> vendor.set(JvmVendorSpec.ADOPTOPENJDK)
+            }
+        }.getOrNull()
+    }
+    javaLauncher.set(launcher)
+    // So processes started from test also use same Java version
+    environment = mapOf("JAVA_HOME" to javaLauncher.get().metadata.installationPath)
```
wolfs, line 119: Would it be better here to only add the one extra environment variable instead of setting all of them?

wolfs (commented):
I left some comments.

build.gradle.kts
asodja: Fixed

.teamcity/settings/GradleProfilerTest.kt
asodja: Good catch, removed

wolfs (approved):
LGTM!

PR #368, opened 2021-11-11
Add dependency mutation

asodja:
Currently projects are generated. And then combinations of projects are applied to the build


README.md
```diff
@@ -276,6 +276,7 @@ Values are optional and default to the values provided on the command-line or de
 A scenario can define changes that should be applied to the source before each build. You can use this to benchmark or profile an incremental build. The following mutations are available:
 
 - `apply-build-script-change-to`: Add a statement to a Groovy or Kotlin DSL build script, init script or settings script. Each iteration adds a new statement and removes the statement added by the previous iteration.
+- `apply-dependency-change-to`: Add dependencies as projects to a Groovy or a Kotlin DSL build script. Each iteration adds a new combinations of projects as dependencies and removes the projects added by the previous iteration.
```
lptr, line 279: I think the name should be reflective to the fact that we are modifying _project_ dependencies. Something like `--apply-project-dependency-change-to`?

asodja: Agreed. I would be happy to get a suggestions also for:
```
            projects-set-size = 10
            applied-projects-size = 5
```
since it feels like they don't tell exactly what they do (I already renamed that 2 times but I am still not happy)

asodja: Renamed, also renamed `applied-projects-size` to `applied-projects-set-size`

src/test/groovy/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorIntegrationTest.groovy
```diff
@@ -0,0 +1,101 @@
...
+            s1 {
+                tasks = assemble
+                apply-project-dependency-change-to {
+                    files = ["build.gradle"]
+                }
+            }
+        """
+
+        when:
+        new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--gradle-version", latestSupportedGradleVersion, "--scenario-file", scenarioFile.absolutePath, "--benchmark", "s1")
```
lptr, line 22: Can we extract this into some sort of utility method?

src/test/groovy/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorIntegrationTest.groovy
```diff
@@ -0,0 +1,101 @@
...
+import org.gradle.profiler.Main
+
+class ApplyProjectDependencyChangeMutatorIntegrationTest extends AbstractProfilerIntegrationTest {
+
+    def "successfully benchmarks gradle projects with dependency mutations with default values"() {
+        given:
+        instrumentedBuildScript()
+        println buildFile
+        def scenarioFile = file("performance.scenarios") << """
+            s1 {
```
lptr, line 13: Let's not use single-letter names.

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeSetupMutator.java
```diff
@@ -0,0 +1,66 @@
...
+            File buildGradle = new File(projectDir, "build.gradle");
+            FileSupport.writeUnchecked(buildGradle.toPath(), "plugins { id 'java' }");
+        }
+    }
+
+    @Override
+    public void afterScenario(ScenarioContext context) {
+        FileUtils.deleteQuietly(generatedProjectsDir);
+        FileSupport.writeUnchecked(settingsFile.toPath(), originalText);
+    }
```
lptr, line 65: Does this approach work when combining multiple build-script mutators? If it does, cool. If not, then should it? If not, then it should fail in a visible way (i.e. error message preventing multiple build script mutators to be applied).

I think it should work, but for that we need to save the original text in `beforeScenario()`, not in the constructor, no?

lptr: OK, I see we don't do this in `AbstractFileChangeMutator` either. And if we want to do it, we should call `afterScenario()` in reverse order in `CompositeBuildMutator`... I think we should do this, but not necessarily in this PR.

src/main/java/org/gradle/profiler/mutations/FileChangeMutatorConfigurator.java
```diff
@@ -60,6 +60,7 @@ private static File openFile(String fileName, File projectDir, String scenarioNa
 			return null;
 		} else {
 			File file = new File(projectDir, fileName);
+            System.out.println(file.toPath().toAbsolutePath());
```
lptr, line 63: Debug leftover?

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.gradle.profiler.mutations.support.ProjectCombinations;
+
+import java.io.File;
+import java.util.Set;
+
+public class ApplyProjectDependencyChangeMutator extends AbstractFileChangeMutator {
+
+    private final ProjectCombinations projectCombinations;
+
+    protected ApplyProjectDependencyChangeMutator(File sourceFile, ProjectCombinations projectCombinations) {
```
lptr, line 15: Why `protected` here?

src/main/java/org/gradle/profiler/mutations/FileChangeMutatorConfigurator.java
```diff
@@ -24,30 +24,30 @@ public FileChangeMutatorConfigurator(Class<? extends AbstractFileChangeMutator>
...
-					throw e.getCause();
-				}
-			} catch (Throwable e) {
-				throw new RuntimeException("Could not create instance of mutator " + mutatorClass.getSimpleName(), e);
-			}
-
-		} else {
-			return null;
-		}
+	protected BuildMutator getBuildMutatorForFile(Config scenario, InvocationSettings settings, File sourceFileToChange) {
```
lptr, line 33: This should instead be package private, no? Or just `public`, given that we don't typically use package private in Gradle elsewhere.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinations.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public ProjectCombinations(String salt, List<String> projectNames, Set<Set<String>> combinations) {
+        this.salt = salt;
+        this.projectNames = Collections.unmodifiableList(projectNames);
+        this.combinations = Iterables.cycle(combinations).iterator();
+    }
+
+    public String getSalt() {
+        return salt;
+    }
```
lptr, line 24: This method is not called from production code. Why is it here?

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private ProjectCombinationsSupport() {
+    }
+
+    public static ProjectCombinations createProjectCombinations(int projectsToGenerate, int appliedProjectDependencies) {
+        checkArgument(projectsToGenerate > 0 && appliedProjectDependencies > 0, String.format("Values '%s' and '%s' should be greater than 0.", PROJECTS_SET_SIZE, APPLIED_PROJECTS_SET_SIZE));
+        checkArgument(projectsToGenerate >= appliedProjectDependencies, String.format("Value '%s' should be at least equal to '%s'.", PROJECTS_SET_SIZE, APPLIED_PROJECTS_SET_SIZE));
+        String salt = UUID.randomUUID().toString().substring(0, 5);
+        List<String> projectNames = IntStream.range(0, projectsToGenerate)
+            .mapToObj(it -> String.format("project-%s-%s", salt, it))
```
lptr, line 26:
```suggestion
            .mapToObj(index -> String.format("project-%s-%s", salt, index))
```

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class ProjectCombinationsSupport {
+
+    private ProjectCombinationsSupport() {
+    }
+
+    public static ProjectCombinations createProjectCombinations(int projectsToGenerate, int appliedProjectDependencies) {
+        checkArgument(projectsToGenerate > 0 && appliedProjectDependencies > 0, String.format("Values '%s' and '%s' should be greater than 0.", PROJECTS_SET_SIZE, APPLIED_PROJECTS_SET_SIZE));
+        checkArgument(projectsToGenerate >= appliedProjectDependencies, String.format("Value '%s' should be at least equal to '%s'.", PROJECTS_SET_SIZE, APPLIED_PROJECTS_SET_SIZE));
+        String salt = UUID.randomUUID().toString().substring(0, 5);
```
lptr, line 24: Why do we need a salt here? Couldn't we simply name our projects something like `__generated-dependency-project` or something to make sure there are no collisions?

src/test/groovy/org/gradle/profiler/mutations/ApplyProjectDependencyChangeSetupMutatorTest.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+    def "generates projects to gradle-profiler-generated-projects folder"() {
+        tmpDir.newFile("settings.gradle")
+        def mutator = new ApplyProjectDependencyChangeSetupMutator(tmpDir.root, combinations)
+
+        when:
+        mutator.beforeScenario(scenarioContext)
+
+        then:
+        File project1BuildScript = file(tmpDir.root, "gradle-profiler-generated-projects", "project-$salt-0", "build.gradle")
```
lptr, line 36: I think in this case it'd be fine to simply use `/`, it works on Windows as well.

src/main/java/org/gradle/profiler/mutations/support/FileSupport.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.OpenOption;
+import java.nio.file.Path;
+
+public class FileSupport {
+
+    private FileSupport() {
+    }
```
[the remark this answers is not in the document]
asodja: removed

src/test/groovy/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorIntegrationTest.groovy
asodja: renamed

src/main/java/org/gradle/profiler/mutations/FileChangeMutatorConfigurator.java
asodja: looks like it

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutator.java
asodja: Changed to public

src/main/java/org/gradle/profiler/mutations/FileChangeMutatorConfigurator.java
asodja: Changed to private, it stayed protected since I was using it at one point. But reverted it later.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
asodja: My idea was. that we don't have collisions between our own projects in case something goes wrong. But yes I can change that behavior. I will add some random string instead of `__` though.

src/test/groovy/org/gradle/profiler/mutations/ApplyProjectDependencyChangeSetupMutatorTest.groovy
asodja: I did not know that, will change, thank you

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinations.java
asodja: removed

src/main/java/org/gradle/profiler/mutations/support/FileSupport.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static void writeUnchecked(Path path, String text, OpenOption... options) {
+        try {
+            Files.write(path, text.getBytes(StandardCharsets.UTF_8), options);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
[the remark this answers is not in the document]
asodja: Made it more symetric

src/test/groovy/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorIntegrationTest.groovy
asodja: moved to runBenchmark method

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
asodja: fixed

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeSetupMutator.java
asodja: Makes senses, moved logic to the `beforeScenario` for that class and FileChangeMutator.

README.md
```diff
@@ -307,6 +308,16 @@ They can be added to a scenario file like this:
         tasks = ["assemble"]
 
         apply-build-script-change-to = "build.gradle.kts"
+        apply-project-dependency-change-to {
+            files = ["build.gradle"]
+            // Example: This will apply a unique combination of 5 projects from a set of 10 projects.
```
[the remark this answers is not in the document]
asodja: Changed to 2.

README.md
```diff
@@ -307,6 +308,16 @@ They can be added to a scenario file like this:
...
+            // Number of combinations can be calculated by typing "10 choosen 5" in your favorite search engine.
+            // When both values are equal there is just one combination possible, so all
+            // runs will have the same combination of the projects applied as dependency.
+            // Default values are 10 and 5. 
+            projects-set-size = 10
+            applied-projects-set-size = 5
```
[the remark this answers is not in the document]
asodja: Number of projects is autogenerated now. User can specify `applied-projects-count`, default is 3. Also renamed that set size to `applied-projects-count`.  Explanation is still complicated but nevertheless

README.md
```diff
@@ -310,13 +310,11 @@ They can be added to a scenario file like this:
...
             files = ["build.gradle"]
-            // Example: This will apply a unique combination of 5 projects from a set of 10 projects.
-            // Number of combinations can be calculated by typing "10 choosen 5" in your favorite search engine.
-            // When both values are equal there is just one combination possible, so all
-            // runs will have the same combination of the projects applied as dependency.
-            // Default values are 10 and 5. 
-            projects-set-size = 10
-            applied-projects-set-size = 5
+            # Default number of applied-projects-count is 3.
+            # Gradle-profiler will generate minimal number of projects to have a unique combination of project dependencies every run.
```
asodja, line 314: I think this is not easy to understand...

lptr: How about:
```suggestion
            # Gradle Profiler will simulate changes to project dependencies by generate some additional projects and then add a combination of project dependencies to every non-generated subprojects before each iteration.
            # The profiler will generate the minimal number of subprojects to allow for a unique combination of dependencies to be used for each iteration.
```

README.md
```diff
@@ -307,6 +308,14 @@ They can be added to a scenario file like this:
         tasks = ["assemble"]
 
         apply-build-script-change-to = "build.gradle.kts"
+        apply-project-dependency-change-to {
+            files = ["build.gradle"]
+            # Default number of applied-projects-count is 3.
+            # Gradle-profiler will generate minimal number of projects to have a unique combination of project dependencies every run.
+            # Note: Number of projects is calculated as binomial coffiecient: "X choose applied-projects-count = iterations * files",
+            # due to that, higher the applied-projects-count is, less projects will be generated.
```
lptr, line 316:
```suggestion
            # Because of this the higher the applied-projects-count is the fewer projects will be generated.
```

README.md
```diff
@@ -307,6 +308,14 @@ They can be added to a scenario file like this:
         tasks = ["assemble"]
 
         apply-build-script-change-to = "build.gradle.kts"
+        apply-project-dependency-change-to {
+            files = ["build.gradle"]
+            # Default number of applied-projects-count is 3.
+            # Gradle-profiler will generate minimal number of projects to have a unique combination of project dependencies every run.
+            # Note: Number of projects is calculated as binomial coffiecient: "X choose applied-projects-count = iterations * files",
```
lptr, line 315: Could you clarify this a bit?

README.md
```diff
@@ -307,6 +308,14 @@ They can be added to a scenario file like this:
...
+            # due to that, higher the applied-projects-count is, less projects will be generated.
+            applied-projects-count = 3
```
lptr, line 317: This term is still not meaningful for me. What is "applied" is the change itself, not the projects. What we want here is to say how many subproject dependencies the profiler is supposed to add. How about:

`project-dependency-count = 3`

or:

`dependency-count = 3`

or even simply:

`count = 3`

?

I think in the context even `count` makes perfect sense.

src/main/java/org/gradle/profiler/CompositeBuildMutator.java
```diff
@@ -47,7 +49,7 @@ public void afterBuild(BuildContext context, Throwable error) {
 
 	@Override
 	public void afterScenario(ScenarioContext context) {
-		for (BuildMutator mutator : mutators) {
+		for (BuildMutator mutator : Lists.reverse(mutators)) {
```
lptr, line 52: I guess we should also invoke `afterBuild` in reverse, no?

lptr: Probably not very important, but creating a new list just to iterate through the original backwards seems... unnecessary here.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -210,6 +207,7 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
                 File bazelHome = getToolHome(executionInstructions);
                 File outputDir = new File(scenarioBaseDir, "bazel");
                 int warmUpCount = getWarmUpCount(settings, scenario);
+                List<BuildMutator> mutators = getMutators(scenario, scenarioName, settings, warmUpCount, buildCount);
```
lptr, line 210: Why do these individual calls here now? Seems redundant.

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorConfigurator.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.gradle.profiler.mutations.support.ProjectCombinationsSupport.createProjectCombinations;
+import static org.gradle.profiler.mutations.support.ScenarioSupport.sourceFiles;
+
+public class ApplyProjectDependencyChangeMutatorConfigurator implements BuildMutatorConfigurator {
```
lptr, line 19: This should be a nested `Configurator` class inside `ApplyProjectDependencyChangeMutator`.

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutator.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.gradle.profiler.BuildContext;
+import org.gradle.profiler.BuildInvoker;
+import org.gradle.profiler.GradleBuildInvoker;
+import org.gradle.profiler.mutations.support.ProjectCombinations;
+
+import java.io.File;
+import java.util.Set;
+
+public class ApplyProjectDependencyChangeMutator extends AbstractFileChangeMutator {
```
lptr, line 11: It would be good to add some Javadoc here describing how this relates to `ApplyProjectDependencyChangeSetupMutator`. I wonder if the two could be a single mutator, or one delegating to the other or something, to capture the relationship?

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeSetupMutator.java
```diff
@@ -0,0 +1,67 @@
...
+            File projectDir = new File(generatedProjectsDir, projectName);
+            projectDir.mkdirs();
+            File buildGradle = new File(projectDir, "build.gradle");
+            FileSupport.writeUnchecked(buildGradle.toPath(), "plugins { id 'java' }");
+        }
+    }
+
+    @Override
+    public void afterScenario(ScenarioContext context) {
+        FileUtils.deleteQuietly(generatedProjectsDir);
```
lptr, line 64: Shouldn't we `forceDelete()` here instead? I'd prefer the profiling failing to running the next benchmark with some leftover generated files.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static org.gradle.profiler.mutations.ApplyProjectDependencyChangeMutatorConfigurator.APPLIED_PROJECTS_COUNT_KEY;
+
+public class ProjectCombinationsSupport {
+
+    /**
+     * Just to prevent some collisions with user's projects
+     */
+    public static final String PROJECT_HASH = "4b038a";
```
lptr, line 20: I think instead of something like this we should use a prefix like `generated-dependency-`. If you have a subproject called `generated-depednency-2` yourself, then... well, you can submit a PR and then we can make the prefix configurable. Let's not make this too defensive.

src/main/java/org/gradle/profiler/mutations/support/ScenarioSupport.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public static List<File> sourceFiles(Config config, String scenarioName, File projectDir, String key) {
+        return ConfigUtil.strings(config, key)
+            .stream()
+            .map(fileName -> openFile(fileName, projectDir, scenarioName))
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
+    }
+
+    private static File openFile(String fileName, File projectDir, String scenarioName) {
```
lptr, line 21: This doesn't really open files, but instead converts a `String` to a `File` while making sure it exists.

src/main/java/org/gradle/profiler/mutations/support/ScenarioSupport.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class ScenarioSupport {
+
+    public static List<File> sourceFiles(Config config, String scenarioName, File projectDir, String key) {
+        return ConfigUtil.strings(config, key)
+            .stream()
+            .map(fileName -> openFile(fileName, projectDir, scenarioName))
+            .filter(Objects::nonNull)
```
lptr, line 17: You could move filtering for `null`s earlier, and then you can simplify `openFile()`. But why are we filtering for `null`? Does `ConfiguUtil.strings()` ever return a `null` entry?

src/main/java/org/gradle/profiler/mutations/support/ScenarioSupport.java
```diff
@@ -0,0 +1,33 @@
...
+            return null;
+        } else {
+            File file = new File(projectDir, fileName);
+            if (!file.isFile()) {
+                throw new IllegalArgumentException("Source file " + file.getName() + " specified for scenario " + scenarioName + " does not exist.");
+            }
+            return file;
+        }
+    }
+
```
lptr, line 32: How about something like this?

```java
    public static List<File> sourceFiles(Config config, String scenarioName, File projectDir, String key) {
        return ConfigUtil.strings(config, key)
            .stream()
            .map(fileName -> new File(projectDir, fileName))
            .peek(file -> {
                if (!file.isFile()) {
                    throw new IllegalArgumentException("Source file " + file.getName() + " specified for scenario " + scenarioName + " does not exist.");
                }
            })
            .collect(Collectors.toList());
    }
```

I understand that you didn't write this code, but I need to write my self-assessment and needed distraction. Sorry. :)

lptr: We could probably also return the stream instead of converting to a `List`, too.

src/main/java/org/gradle/profiler/mutations/support/FileSupport.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public static String readUnchecked(Path path) {
+        try {
+            return new String(Files.readAllBytes(path), StandardCharsets.UTF_8.toString());
+        } catch (IOException e) {
+            throw new UncheckedIOException("Could not read contents of a file: " + path, e);
+        }
+    }
+}
```
lptr, line 27: I'm wondering if these `*Support` types are worth it vs. inlining the trivial code, or using something already existing. Let's leave it like this for now, but let's talk about this during the retro or something.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static org.gradle.profiler.mutations.ApplyProjectDependencyChangeMutatorConfigurator.APPLIED_PROJECTS_COUNT_KEY;
+
+public class ProjectCombinationsSupport {
```
lptr, line 15: Why not have these methods on `ProjectCombinations`?

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinations.java
```diff
@@ -0,0 +1,25 @@
+package org.gradle.profiler.mutations.support;
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+public class ProjectCombinations {
```
lptr, line 8: Why does this belong in a package called `support`? This is part of the mutators.

If we could merge the two project-dependency-change-related mutators, we could move this to a nested subclass, and that'd look a lot better I think.

src/test/groovy/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutatorTest.groovy
```diff
@@ -28,9 +29,11 @@ class ApplyChangeToAndroidLayoutFileMutatorTest extends AbstractMutatorTest {
         def mutator = new ApplyChangeToAndroidLayoutFileMutator(sourceFile)
 
         when:
+        mutator.beforeScenario(scenarioContext)
         mutator.beforeBuild(buildContext)
 
         then:
+        println sourceFile.text
```
lptr, line 36: Debug.

src/test/groovy/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutatorTest.groovy
```diff
@@ -47,6 +50,7 @@ class ApplyChangeToAndroidLayoutFileMutatorTest extends AbstractMutatorTest {
         def mutator = new ApplyChangeToAndroidLayoutFileMutator(sourceFile)
 
         when:
+        mutator.beforeScenario(scenarioContext)
         mutator.afterScenario(scenarioContext)
```
lptr, line 54: I think these tests don't make much sense now. We could make some random change between `beforeScenario()` and `afterScenario()` to make it meaningful.

But since this logic of "saving and restoring the original text" happens for every file change mutator, it's not really useful to test it so many times.

lptr: Maybe a better option would be to have a method to create the mutator in every test that also calls `mutator.beforeScenario()` and `mutator.beforeBuild()`?

src/main/java/org/gradle/profiler/mutations/support/FileSupport.java
asodja: They are mostly here since it makes code much cleaner and easier to read. For that particular example, always handling these exceptions to throw unchecked exceptions makes a lot of noise.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
asodja: I really like "data only classes". So classes that don't have any logic in there (except maybe trivial creator methods), since that makes other code easier to test also soon that data class can get bloated with similar static methods.

This particular logic is already a bit complex so I decided to move out of it. Since this is a creater logic, I can move it into the class.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinations.java
asodja: Mutations package is really huge, and it's hard to find classes there. I created this package so these support classes, that are not mutators, are easy to find. They are still under mutations package.

I will check how easy it is to merge two classes into one.

asodja: I merged mutator classes into one. Code wise it is probably easier to understand. I also moved ProjectCombinations as a subclass of now one mutator class. 

But I kept the support class, that way since it's a bit complex logic and it's much nicer to test that if it is separate. And since it's easier to test that when separate unit, that is a good flag for me that it might deserve a separate class. WDYT?

src/main/java/org/gradle/profiler/ScenarioLoader.java
asodja: Warmup count and build count are calculated differently in some cases. And since we these two counts to the configurators, we have to handle it that way

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorConfigurator.java
asodja: What is the reason for that? There is for example FileChangeMutatorConfigurator that is not nested. And nesting this class will make ApplyProjectDependencyChangeMutator bigger and harder to read.

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutator.java
asodja: Merged these two mutators into one as suggested. Also added some javadoc what it does

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeSetupMutator.java
asodja: Makes sense, changed to force delete

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
asodja: removed hash

src/main/java/org/gradle/profiler/mutations/support/ScenarioSupport.java
asodja: I moved nonull higher and removed open method

src/main/java/org/gradle/profiler/mutations/support/ScenarioSupport.java
asodja: removed

src/main/java/org/gradle/profiler/mutations/support/ScenarioSupport.java
asodja: Used that code and also add `.filter(Objects::nonNull)` since I am not sure if ConfigUtil can return null. Better safe than sorry

src/test/groovy/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutatorTest.groovy
asodja: I see your point., but there are actually tests with: ```reverts changes when nothing has been applied``` :D. So they already tested before that there is no action.

I left them for now, since they are there, do you think I should delete them to not continue with that pattern in future?

src/main/java/org/gradle/profiler/CompositeBuildMutator.java
asodja: Added Lists.reverse() also to `afterBuild` and `afterCleanup`. Yeah I did not wanted to reverse original list, since I don't want to mutate original list. 

I believe Lists.reverse() does not do an actual copy of a list, but it just creates a wrapper around the list. And looping with indexes is just so 2010 :P

README.md
asodja: Changed to dependency-count. For me it seems `count` does not hold enough information. And `dependency-count` is nice in case we will have also `apply-external-dependency-change-to` or smth similar in future

README.md
asodja: I tried, let me know what you think

README.md
asodja: I removed that "because of this..." line, it seems confusing. If someone will want use some different non-default value, he/she can calculate values with provided formula.

src/test/groovy/org/gradle/profiler/mutations/ApplyChangeToAndroidLayoutFileMutatorTest.groovy
asodja: Thank you, removed

README.md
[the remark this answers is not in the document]
lptr: 👍

src/main/java/org/gradle/profiler/CompositeBuildMutator.java
lptr: Wait, 2010 means old? I thought you were supposed to say "XY is so 1970!" or something. Must be getting old. :)

You're right on `reverse()` being cheap, let's keep it!

lptr (approved):
LGTM, the remaining comments are about code style that we can discuss separately if we want.

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinations.java
lptr: I like that `ProjectCombinations` is now a nested class. I think that makes this support class even more of an orphan that should be reunited with its folks.

asodja: removed support class

src/main/java/org/gradle/profiler/mutations/support/ProjectCombinationsSupport.java
asodja: removed support class

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutatorConfigurator.java
asodja: Ok, I moved it to the `ApplyProjectDependencyChangeMutator`

src/test/groovy/org/gradle/profiler/mutations/support/ProjectCombinationsTest.groovy
```diff
@@ -1,12 +1,15 @@
 package org.gradle.profiler.mutations.support
```
lptr, line 1: Shall we move this to a level higher? Or merge the test into `ApplyProjectDependencyChangeMutatorTest`?

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutator.java
```diff
@@ -123,5 +139,78 @@ public ProjectCombinations(List<String> projectNames, Set<Set<String>> combinati
         public Set<String> getNextCombination() {
             return combinations.next();
         }
+
+        public static ProjectCombinations createProjectCombinations(int numberOfRequiredCombinations, int appliedProjectsCount) {
```
lptr, line 143: No need for this to be `public` (we typically use default visibility + `@VisibleForTesing` for methods that need to be available from tests). The methods can also be part of `ApplyProjectDependencyChangeMutator` itself, though that might be a question of taste.

lptr (approved):

src/test/groovy/org/gradle/profiler/mutations/support/ProjectCombinationsTest.groovy
asodja: Done

src/main/java/org/gradle/profiler/mutations/ApplyProjectDependencyChangeMutator.java
asodja: Removed public and add annotation

PR #369, opened 2021-11-12
Detect if studio was not started sooner

asodja:
This PR changes the behavior of the profiler so we detect faster if Android Studio was not started. Previously we waited 60s, now we wait for Android Studio start just 15s. This is possible since we now connect to the profiler from the plugin as soon as Android Studio starts and we don't wait anymore for project to open. Normally Studio start happens after 2s on my Mac, but to be on the safe side I set it to 15s. 

Also this change removes singleton nature from Client and cleans up a bit StudioGradleClient.


src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -57,60 +56,32 @@ public StudioGradleClient(GradleBuildConfiguration buildConfiguration, Invocatio
...
-        for (Path entry : launchConfiguration.getClassPath()) {
-            System.out.println("  " + entry);
-        }
-        System.out.println("* System properties:");
-        for (Map.Entry<String, String> entry : launchConfiguration.getSystemProperties().entrySet()) {
-            System.out.println("  " + entry.getKey() + " -> " + entry.getValue());
-        }
-        System.out.println("* Main class: " + launchConfiguration.getMainClass());
+        LaunchConfiguration launchConfiguration = new LauncherConfigurationParser(studioInstallDir, sandbox)
+            .installStudioPlugin(studioStartDetectorServer.getPort(), studioPluginServer.getPort())
```
wolfs, line 62: Since we are always calling these two methods, why do we have the boolean flags in `LauncherConfigurationParser`?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/agent/AgentClientContainer.java
```diff
@@ -0,0 +1,23 @@
+package org.gradle.profiler.client.protocol.agent;
+
+import org.gradle.profiler.client.protocol.Client;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public enum AgentClientContainer {
+    INSTANCE;
+
+    private final AtomicReference<Client> client = new AtomicReference<>();
```
wolfs, line 10: I don't think an `AtomicReference` is the right thing here. Why not use a `volatile Client client` field?

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/agent/AgentClientContainer.java
```diff
@@ -0,0 +1,23 @@
+package org.gradle.profiler.client.protocol.agent;
+
+import org.gradle.profiler.client.protocol.Client;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public enum AgentClientContainer {
```
wolfs, line 7: Can we maybe get rid of this client "holder" altogether and pass around the client after it has been created?

wolfs (commented):
I left some comments.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -43,9 +42,9 @@
 
     private final Server studioAgentServer;
     private final Server studioPluginServer;
-    private final CommandExec.RunHandle studioProcess;
-    private final ServerConnection studioAgentConnection;
-    private final ServerConnection studioPluginConnection;
+    private final RunHandle studioProcess;
+    private ServerConnection studioAgentConnection;
```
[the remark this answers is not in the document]
asodja: Probably there was an "upsi" during refactoring. They are back to final in latest revisions

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
```diff
@@ -133,4 +104,20 @@ public BuildActionResult sync(List<String> gradleArgs, List<String> jvmArgs) {
             Duration.ofMillis(syncRequestCompleted.getDurationMillis() - agentCompleted.getDurationMillis())
         );
     }
+
+
```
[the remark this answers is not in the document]
asodja: removed

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
asodja: Actually, in tests we start a studio without a plugin or an agent. So there it comes handy that we have option to handle this

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/agent/AgentClientContainer.java
asodja: Hmm, I think there is no big difference, except that AtomicReference might take more memory. I set it to volatile though.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/agent/AgentClientContainer.java
asodja: This container is used so the client is shared between agent and interceptor. I think it is possible, but it would require some extra effort to modify instrumentation. I think it's easier with a static reference. WDYT?

src/main/java/org/gradle/profiler/studio/tools/StudioLauncher.java
```diff
@@ -0,0 +1,36 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+public class StudioLauncher {
+
+    private StudioLauncher() {
+    }
+
+    public static CommandExec.RunHandle launchStudio(LaunchConfiguration launchConfiguration, File projectDir) {
```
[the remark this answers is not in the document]
asodja: We can also do that yes. Moved to `LaunchConfiguration`

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/agent/AgentClientContainer.java
wolfs: Ok, then let's leave it as is.

src/main/java/org/gradle/profiler/studio/StudioGradleClient.java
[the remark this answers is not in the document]
wolfs: Right, I saw that and forgot to delete my comment.

subprojects/client-protocol/src/main/java/org/gradle/profiler/client/protocol/agent/AgentClientContainer.java
wolfs: In your usage it doesn't make a difference, right. When looking at this I was wondering why we did the extra locking when updating the atomic reference, and then I thought it would be better to use `volatile` to remove the confusion.

wolfs (approved):
LGTM!

PR #372, opened 2021-11-22
Add support for running Android Studio profiler with Linux and Windows and running tests on CI/CD 

asodja:


src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class LauncherConfigurationParser {
```
asodja, line 15: This class was just moved and some parts of it were moved to StudioConfigurationProvider

.teamcity/settings/extensions.kt
```diff
@@ -17,6 +17,15 @@ fun ParametrizedWithType.javaHome(os: Os, javaVersion: JavaVersion) {
     param("env.JAVA_HOME", javaVersion.javaHome(os))
 }
 
+fun ParametrizedWithType.androidHome(os: Os) {
+    val androidHome = when (os) {
+        Os.linux, Os.macos -> "/opt/android/sdk"
+        Os.windows -> """C:\Program Files\android\sdk"""
+    }
+    param("env.ANDROID_HOME", androidHome)
```
wolfs, line 25: Let's only set `ANDROID_SDK_ROOT` and not `ANDROID_HOME`. `ANDROID_HOME` is only for very old AGP versions.

build.gradle.kts
```diff
@@ -1,3 +1,4 @@
+
```
wolfs, line 1: Nit: Unexpected newline

build.gradle.kts
```diff
@@ -128,6 +130,13 @@ tasks.test {
     setForkEvery(1)
 }
 
+androidStudioTests {
+    val autoDownloadAndRunInHeadless = providers.gradleProperty("autoDownloadAndRunInHeadless").orNull == "true"
```
wolfs, line 134: How about you use `map` here, so you can convert to a Property<Boolean>` and use that directly. That would mean you don't need to evaluate the property at configuration time.

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+fun isMacOS(): Boolean = os.startsWith("mac")
+fun isLinux(): Boolean = os.startsWith("linux")
+fun isIntel(): Boolean = architecture == "x86_64" || architecture == "x86"
+fun File.isTar(): Boolean = name.endsWith(".tar.gz")
+
+val extension = extensions.create<AndroidStudioTestExtension>("androidStudioTests")
+
+val androidStudioRuntime by configurations.creating
+val autoDownloadAndroidStudio by lazy { extension.autoDownloadAndroidStudio.getOrElse(false) }
+afterEvaluate {
```
wolfs, line 26: Can we remove this `afterEvaluate` somehow? You could always add the dependency, with a lazy version I suppose. And then skip the task if you shouldn't download it.

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+                isLinux() -> androidStudioRuntime("android-studio:android-studio:$androidStudioVersion@linux.tar.gz")
+            }
+        }
+    }
+}
+
+val androidStudioPath = "$buildDir/android-studio"
+val unpackAndroidStudio = tasks.register<Copy>("unpackAndroidStudio") {
+    if (autoDownloadAndroidStudio) {
+        val file = androidStudioRuntime.files.first()
```
wolfs, line 43: This happens at configuration time, so should happen at runtime. I think you can make this lazy.

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+        inputs.file(file)
+        outputs.dir(androidStudioPath)
```
wolfs, line 45: The Copy task already registers these inputs and outputs, no need to add them manually.

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+        from(fileTree)
+        into(androidStudioPath)
+    }
+}
+
+val macOsAndroidStudioPath = "$androidStudioPath/Android Studio.app"
+val macOsAndroidStudioPathPreview = "$androidStudioPath/Android Studio Preview.app"
+val windowsAndLinuxPath = "$androidStudioPath/android-studio"
+tasks.withType<Test>().configureEach {
+    dependsOn(unpackAndroidStudio)
```
wolfs, line 59: Instead of adding this `dependsOn`, you probably should wire up the task by using the output property.

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+    if (autoDownloadAndroidStudio) {
+        val studioHome = when {
+            isMacOS() && file(macOsAndroidStudioPath).exists() -> macOsAndroidStudioPath
+            isMacOS() -> macOsAndroidStudioPathPreview
+            else -> windowsAndLinuxPath
+        }
+        systemProperty("studio.home", studioHome)
```
wolfs, line 66: This makes the task non-relocatable, let's use a command line argument provider instead. That also allows declaring the input location better.

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    if (extension.runAndroidStudioInHeadlessMode.getOrElse(false)) {
```
wolfs, line 68: I suppose this should also go into the command line argument provider, so we don't need to evaluate this eagerly here.

src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
```diff
@@ -0,0 +1,108 @@
...
+        if (OperatingSystem.isMacOS()) {
+            return getMacOSConfiguration(studioInstallDir);
+        } else {
+            return getWindowsOrLinuxConfiguration(studioInstallDir);
+        }
+    }
+
+    private static StudioConfiguration getWindowsOrLinuxConfiguration(Path studioInstallDir) {
+        String mainClass = "com.intellij.idea.Main";
+        List<Path> classPath = Stream.of("lib/bootstrap.jar", "lib/util.jar", "lib/jdom.jar", "lib/log4j.jar", "lib/jna.jar")
```
wolfs, line 33: Can we parse the classpath on Windows/Linux from a file as well? It feels wrong to hardcode this here.

wolfs (commented):
I left some comments regarding lazy wiring of tasks. Apart from that this looks good!

src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
wolfs: We discussed this and it is too complicated to parse it from the start script. Let's see if the need arises to change it and then let's re-visit.

.teamcity/settings/extensions.kt
asodja: done

build.gradle.kts
asodja: removed

wolfs (approved):
LGTM!

Issue #371, opened 2021-11-18
Move IDE related code out of GradleScenarioInvoker


asodja:
After introducing Android Studio syncs we now some code related to IDE in GradleScenarioInvoker and other places. It would be nice that IDE classes are separated and they only enhance the results somehow.

Let's check if we can separate that code to Gradle specific and IDE specific code.

PR #374, opened 2021-11-22
Move Android Studio related code out of GradleScenarioInvoker and rel…

asodja:
…ated classes

So yeah, I did not introduce generics to BuildAction at the moment since this can complicate things. So right now I just wrap Gradle invoker logic and enhance samples, here is some casting involved that might not be that nice (in `StudioSample` class), but it makes a solution simpler.

Fixes: #371


src/main/java/org/gradle/profiler/studio/invoker/StudioSample.java
```diff
@@ -0,0 +1,34 @@
...
+            return studioResult.getGradleExecutionTime();
+        }
+    },
+    IDE_EXECUTION_TIME("IDE execution time") {
+        @Override
+        public Duration extractFrom(BuildInvocationResult result) {
+            StudioBuildActionResult studioResult = (StudioBuildActionResult) result.getActionResult();
+            return studioResult.getIdeExecutionTime();
+        }
+    };
```
lptr, line 22: I'd move these to `StudioBuildActionResult`, and likewise move `EXECUTION_TIME` to `BuildActionResult`. WDYT?

lptr: I guess I'm just not very fond of the use of these `enum`s.

BTW, if we do want to use `enum`s, we could change `getName()` to `name()` and that would simplify things significantly.

src/main/java/org/gradle/profiler/studio/invoker/StudioSample.java
```diff
@@ -0,0 +1,34 @@
...
+import org.gradle.profiler.result.BuildInvocationResult;
+import org.gradle.profiler.result.Sample;
+
+import java.time.Duration;
+
+public enum StudioSample implements Sample<BuildInvocationResult> {
+    GRADLE_EXECUTION_TIME("Gradle execution time") {
+        @Override
+        public Duration extractFrom(BuildInvocationResult result) {
+            StudioBuildActionResult studioResult = (StudioBuildActionResult) result.getActionResult();
```
lptr, line 12: We don't want to introduce a `StudioBuildInvocationResult` to avoid this cast?

asodja: That would be ideal, but it requires some changes, I will check what I can do.

asodja: What I did. I created a StudioBuildInvocationResult, and I do a cast there. It's not totally perfect, but I think that right now that is the simplest and cleanest solution. 

Otherwise, we would have to inject (or override some method that will return) a custom `BuildUnderTestInvoker` into `GradleScenarioInvoker` and also create a new `InvokeAndMeasureAction` that would return  `StudioBuildInvocationResult`. Additionally to that, `BuildAction` would have to return generic result from `run` method and that would require passing generics everywhere and it will get a bit messy.

WDYT?

src/main/java/org/gradle/profiler/studio/invoker/StudioSample.java
asodja: I moved code out of enums. I also created `StudioBuildInvocationResult` and moved samples there. I think it makes sense that they are there since they implement `Sample<StudioBuildInvocationResult>`.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -288,7 +289,11 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
                         outputDir,
                         jvmArgs,
                         measuredBuildOperations
-                    ));
+                    );
+                    ScenarioDefinition scenarioDefinition = (buildAction instanceof AndroidStudioSyncAction)
+                        ? new StudioGradleScenarioDefinition(gradleScenarioDefinition)
+                        : gradleScenarioDefinition;
```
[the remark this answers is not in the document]
asodja: I changed to path, it might be a tiny bit better

src/main/java/org/gradle/profiler/studio/invoker/StudioSample.java
lptr: Alright, one step at a time.

lptr (commented):
LGTM.

PR #376, opened 2021-11-24
Enable parallel builds for gradle-profiler

liutikas:
Speed up the builder of gradle-profiler that currently was all running on a single core.


wolfs:
Thank you @liutikas!

wolfs (approved):
LGTM!

PR #383, opened 2021-12-14
Use absolute paths for async-profiler output files

asodja:
When testing Android Studio sync with gradle-profiler I noticed after a long debugging session that profiler does not work if absolute --output-dir is not set. This fixes it so absolute paths are passed in JVM args.

### DIFF
--- a/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
+++ b/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
@@ -9,13 +9,13 @@ public enum AsyncProfilerOutputType {
     STACKS("collapsed") {
         @Override
         File outputFileFor(ScenarioSettings settings) {
-            return settings.profilerOutputLocationFor(".stacks.txt");
+            return settings.profilerOutputLocationFor(".stacks.txt").getAbsoluteFile();
         }
     },
     JFR("jfr") {
         @Override
         File outputFileFor(ScenarioSettings settings) {
-            return settings.computeJfrProfilerOutputLocation();
+            return settings.computeJfrProfilerOutputLocation().getAbsoluteFile();
         }
     };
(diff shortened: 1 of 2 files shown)


src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
```diff
@@ -34,7 +34,7 @@ public static AsyncProfilerOutputType from(AsyncProfilerConfig config, GradleSce
     abstract File outputFileFor(ScenarioSettings settings);
 
     File individualOutputFileFor(ScenarioSettings settings) {
-        File outputFile = outputFileFor(settings);
+        File outputFile = outputFileFor(settings).getAbsoluteFile();
```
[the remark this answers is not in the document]
asodja: Done.

I also tested jfr, YourKit, JProfiler and they don't have this issue.

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
```diff
@@ -9,13 +9,13 @@
     STACKS("collapsed") {
         @Override
         File outputFileFor(ScenarioSettings settings) {
-            return settings.profilerOutputLocationFor(".stacks.txt");
+            return settings.profilerOutputLocationFor(".stacks.txt").getAbsoluteFile();
```
wolfs, line 12: I wonder if we should go even deeper: It seems like `ScenarioDefinition.getOutputDir()` or `ScenarioDefinition.outputDir` should always be an absolute file. Then the rest would follow automatically I think.

asodja: We can do that

wolfs (approved):
LGTM!

PR #384, opened 2021-12-15
Add option to override Android Studio jvm args

asodja:


src/main/java/org/gradle/profiler/GradleClientSpec.java
```diff
@@ -42,19 +43,19 @@ public GradleClient create(GradleBuildConfiguration buildConfiguration, Invocati
     AndroidStudio("Android Studio") {
         @Override
         public GradleClient create(GradleBuildConfiguration buildConfiguration, InvocationSettings invocationSettings) {
-            return new StudioGradleClient(buildConfiguration, invocationSettings, NEVER);
+            return new StudioGradleClient((StudioGradleBuildConfiguration) buildConfiguration, invocationSettings, NEVER);
```
wolfs, line 46: I wonder if we can avoid the cast here by introducing new types. I am not sure if it is worth it.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -305,6 +307,12 @@ public ScenarioLoader(GradleBuildConfigurationReader gradleBuildConfigurationRea
         return definitions;
     }
 
+    private static StudioGradleScenarioDefinition newStudioGradleScenarioDefinition(GradleScenarioDefinition gradleScenarioDefinition, Config scenario) {
+        Config androidStudioSync = scenario.getConfig(ANDROID_STUDIO_SYNC);
+        List<String> studioJvmArgs = ConfigUtil.strings(androidStudioSync, ANDROID_STUDIO_JVM_ARGS, Collections.emptyList());
```
wolfs, line 312: Should the default be `["-Xms256m", "-Xmx4096m"]` and not the empty list?

asodja: I set it to `["-Xms256m", "-Xmx2048m"]` first as we had before, but then I removed them. I am sure, maybe I can add them back, not sure what is better here

src/main/java/org/gradle/profiler/GradleClientSpec.java
asodja: It would require some extra refactoring, for example we would most likely have to totally separate StudioGradleScenarioInvoker and GradleScenarioInvoker.

wolfs: I wondered if you could share a common interface for the two enums, so at least `StudioGradleScenarioInvoker` would be a `GradleScenarioInvoker` as well. Anyway, it is probably not worth it at this time to separate the two.

src/main/java/org/gradle/profiler/ScenarioLoader.java
wolfs: For me, the basic question is if Android studio works without any jvmArgs. If it does, then I suppose leaving them out is fine. If it doesn't, then we should use some sensible default.

wolfs (approved):
LGTM!

src/main/java/org/gradle/profiler/ScenarioLoader.java
asodja: Without any jvm args it works. But it sets Xmx to default Java xmx. On my machine that is 16GB (1/4 of all available memory).

I set default to  `["-Xms256m", "-Xmx4096m"]` I think it's sensible and if anyone has issue with it, it can override it.

PR #385, opened 2021-12-17
Output chrome trace file for every invocation

asodja:
So now chrome trace outputs file for every gradle invocation separately.

### DIFF
--- a/src/test/groovy/org/gradle/profiler/ChromeTraceIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/ChromeTraceIntegrationTest.groovy
@@ -1,9 +1,10 @@
 package org.gradle.profiler
 
+import org.gradle.profiler.studio.tools.StudioFinder
 import org.gradle.util.GradleVersion
+import spock.lang.Requires
 import spock.lang.Shared
 import spock.lang.Unroll
-
 
 class ChromeTraceIntegrationTest extends AbstractProfilerIntegrationTest {
     static final MINIMAL_SUPPORTED_VERSION_FOR_CHROME_TRACE = GradleVersion.version("3.5")
@@ -23,7 +24,9 @@ class ChromeTraceIntegrationTest extends AbstractProfilerIntegrationTest {
             run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath, "--gradle-version", versionUnderTest, "--profile", "chrome-trace", "assemble")
 
         then:
-        new File(outputDir, "${versionUnderTest}-trace.json").isFile()
+        new File(outputDir, "${versionUnderTest}-trace/${versionUnderTest}-warm-up-build-1-invocation-1-trace.json").isFile()
+        new File(outputDir, "${versionUnderTest}-trace/${versionUnderTest}-warm-up-build-2-invocation-1-trace.json").isFile()
+        new File(outputDir, "${versionUnderTest}-trace/${versionUnderTest}-measured-build-1-invocation-1-trace.json").isFile()
 
         where:
         versionUnderTest << supportedGradleVersionsForChromeTrace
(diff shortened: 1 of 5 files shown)


src/test/groovy/org/gradle/profiler/ChromeTraceIntegrationTest.groovy
```diff
@@ -55,9 +59,38 @@ class ChromeTraceIntegrationTest extends AbstractProfilerIntegrationTest {
...
+        new Main().run("--project-dir", projectDir.absolutePath, "--output-dir", outputDir.absolutePath,
+            "--gradle-version", latestSupportedGradleVersion,
+            "--profile", "chrome-trace",
+            "--scenario-file", scenarioFile.absolutePath,
+            "--studio-install-dir", studioHome.absolutePath,
+            "--warmups", "1",
+            "--iterations", "1")
+
+        then:
+        new File(outputDir, "scenario-${latestSupportedGradleVersion}-trace/scenario-${latestSupportedGradleVersion}-warm-up-build-1-invocation-1-trace.json").isFile()
```
wolfs, line 91: Have we also been generating trace files for warmup builds? I somewhat was expecting that we'd only produce the traces for the actual iterations. Note that I am not sure what is better here.

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
```diff
@@ -25,18 +25,36 @@
...
         this.buildRequestMetaData = gradle.getServices().get(BuildRequestMetaData.class);
-        traceResult = new TraceResult(traceFile);
+        traceResult = new TraceResult(getTraceFile(traceFolder, traceFilePattern));
         systemMonitoring.start(traceResult);
         gcMonitoring.start(traceResult);
         buildOperationListener = BuildOperationListenerAdapter.create(gradle, traceResult);
         gradle.addListener(new TraceFinalizerAdapter(gradle));
     }
 
+    private File getTraceFile(File traceFolder, String traceFilePatter) {
```
wolfs, line 37:
```suggestion
    private File getTraceFile(File traceFolder, String traceFilePattern) {
```

Nit.

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
```diff
@@ -25,18 +25,36 @@
...
+        String phase = System.getProperty("org.gradle.profiler.phase.display.name");
+        String build = System.getProperty("org.gradle.profiler.number");
+        int invocation = 1;
+        File traceFile;
+        do {
+            String traceFileName = traceFilePatter.replace("{phase}", phase)
```
wolfs, line 43: I think it would make sense that the `start` method only takes the file base name, and then would add this postfix. This way we can avoid passing in a pattern altogether. WDYT?

wolfs (commented):
Looks good! I left some comments.

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
asodja: Good point, will fix it

src/test/groovy/org/gradle/profiler/ChromeTraceIntegrationTest.groovy
asodja: It looks like we do yes. Since it's really low overhead I think we can leave it as it is

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
asodja: Fixed

subprojects/chrome-trace/src/main/java/org/gradle/trace/GradleTracingPlugin.java
asodja: Fixed

wolfs (approved):
LGTM!

PR #386, opened 2021-12-21
Automatically install Android SDK on CI

asodja:
To avoid test failures on fresh agents without Android SDK.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+
 import com.moowork.gradle.node.npm.NpxTask
 import io.sdkman.vendors.tasks.SdkAnnounceVersionTask
 import io.sdkman.vendors.tasks.SdkDefaultVersionTask
@@ -134,6 +135,10 @@ androidStudioTests {
     runAndroidStudioInHeadlessMode.set(autoDownloadAndRunInHeadless)
     autoDownloadAndroidStudio.set(autoDownloadAndRunInHeadless)
     testAndroidStudioVersion.set("2021.1.1.16")
+    testAndroidSdkVersion.set("7.0.0")
+    // For local development it's easier to setup Android SDK with Android Studio, since auto download needs ANDROID_SDK_ROOT to be set with
+    // a accepted license in it. See https://developer.android.com/studio/intro/update.html#download-with-gradle.
+    autoDownloadAndroidSdk.set(autoDownloadAndRunInHeadless)
 }
 
 val testReports = mapOf(
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,7 +4,7 @@ import io.sdkman.vendors.tasks.SdkAnnounceVersionTask
 import io.sdkman.vendors.tasks.SdkDefaultVersionTask
 import io.sdkman.vendors.tasks.SdkReleaseVersionTask
 import io.sdkman.vendors.tasks.SdkmanVendorBaseTask
-import java.util.Locale
+import java.util.*
 
 plugins {
     id("profiler.java-library")
@@ -135,9 +135,9 @@ androidStudioTests {
     runAndroidStudioInHeadlessMode.set(autoDownloadAndRunInHeadless)
     autoDownloadAndroidStudio.set(autoDownloadAndRunInHeadless)
     testAndroidStudioVersion.set("2021.1.1.16")
-    testAndroidSdkVersion.set("7.0.0")
+    testAndroidSdkVersion.set("7.1.0-beta04")
     // For local development it's easier to setup Android SDK with Android Studio, since auto download needs ANDROID_SDK_ROOT to be set with
-    // a accepted license in it. See https://developer.android.com/studio/intro/update.html#download-with-gradle.
+    // an accepted license in it. See https://developer.android.com/studio/intro/update.html#download-with-gradle.
     autoDownloadAndroidSdk.set(autoDownloadAndRunInHeadless)
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,10 +1,9 @@
-
 import com.moowork.gradle.node.npm.NpxTask
 import io.sdkman.vendors.tasks.SdkAnnounceVersionTask
 import io.sdkman.vendors.tasks.SdkDefaultVersionTask
 import io.sdkman.vendors.tasks.SdkReleaseVersionTask
 import io.sdkman.vendors.tasks.SdkmanVendorBaseTask
-import java.util.*
+import java.util.Locale
 
 plugins {
     id("profiler.java-library")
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -76,26 +94,46 @@ abstract class AndroidStudioInstallation {
     abstract val studioInstallLocation: DirectoryProperty
 }
 
+abstract class AndroidSdkInstallation {
+    @get:InputFiles
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val sdkInstallLocation: DirectoryProperty
```
wolfs, line 100: That means the SDK directory would be an input. I am not sure snapshotting this directory makes much sense.

buildSrc/src/main/kotlin/tasks/InstallAndroidSdkTask.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+import org.gradle.tooling.GradleConnector
+import java.io.File
+import javax.inject.Inject
+
+/**
+ * Installs Android SDK. This task requires the ANDROID_SDK_ROOT env variable to be set.
+ */
+abstract class InstallAndroidSdkTask : DefaultTask() {
```
wolfs, line 19: I think this task should be untracked, since we don't really want to do up-to-date checks for it, right? Especially the files in the output directory would probably change when we run other builds on it.

build.gradle.kts
```diff
@@ -1,3 +1,4 @@
+
```
wolfs, line 1: Nit: extra new line.

wolfs (commented):
I think this looks good. I am a little bit worried about the input declarations, since I think we'd end up snapshotting a lot.

buildSrc/src/main/kotlin/tasks/InstallAndroidSdkTask.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        createAndroidProject(projectDir, androidSdkVersion)
+        buildAndroidProject(projectDir)
+    }
+
+    private fun buildAndroidProject(projectDir: File) {
+        val connector = GradleConnector.newConnector()
+            .forProjectDirectory(projectDir)
+        connector.connect().use {
+            it.newBuild().forTasks("build").run()
```
wolfs, line 51: I wonder if you'd need to set the environment variable `ANDROID_SDK_ROOT` to `androidSdkRootEnvVariable` for this build, so it actually populates the right directory. I think the other option is to use `System.getenv()` in the task action directly, so you don't depend on anybody configuring the right location.

asodja: This is not needed, will use System.getenv()

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -57,15 +59,31 @@ val unpackAndroidStudio = tasks.register<Copy>("unpackAndroidStudio") {
...
 }
 
+val installAndroidSdk = tasks.register<InstallAndroidSdkTask>("installAndroidSdk") {
+    androidSdkVersion.set(extension.testAndroidSdkVersion)
+    androidProjectDir.set(layout.buildDirectory.dir("android-sdk-project"))
+    androidSdkInstallationDir.set(providers.environmentVariable("ANDROID_SDK_ROOT").flatMap {
+        val installationDirectory = objects.directoryProperty()
+        installationDirectory.set(File(it))
+        installationDirectory
+    })
```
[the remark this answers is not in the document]
asodja: removed this entirely

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -110,6 +148,9 @@ class AndroidStudioSystemProperties(
             }
             systemProperties.add("-Dstudio.home=$studioHome")
         }
+        if (autoDownloadAndroidSdk.get()) {
+            systemProperties.add("-Dsdk.instalation.path=${sdkInstallation.sdkInstallLocation.get().asFile.absolutePath}")
```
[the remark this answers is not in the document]
asodja: removed

buildSrc/src/main/kotlin/tasks/InstallAndroidSdkTask.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Optional since we use manual error message
+     */
+    @get:Optional
+    @get:OutputDirectory
+    abstract val androidSdkInstallationDir: DirectoryProperty
+
+    @get:Inject
+    abstract val projectLayout: ProjectLayout
```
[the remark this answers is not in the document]
asodja: good catch, removed

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
asodja: removed

build.gradle.kts
asodja: removed

wolfs (approved):
LGTM!

buildSrc/src/main/kotlin/tasks/InstallAndroidSdkTask.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        createAndroidProject(projectDir, androidSdkVersion)
+        buildAndroidProject(projectDir)
+    }
+
+    private fun buildAndroidProject(projectDir: File) {
+        val connector = GradleConnector.newConnector()
+            .forProjectDirectory(projectDir)
+        connector.connect().use {
+            it.newBuild().forTasks("build").run()
```
wolfs, line 46: Ah, one more thing: I think it would be good to print out the output of the build when it fails, WDYT?

wolfs (approved):

buildSrc/src/main/kotlin/tasks/InstallAndroidSdkTask.kt
asodja: Added. I had to handle it manually by catching exception. Previously it worked if you used --stacktrace, but now it's always visible

buildSrc/src/main/kotlin/tasks/InstallAndroidSdkTask.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    private fun buildAndroidProject(projectDir: File) {
+        try {
+            val connector = GradleConnector.newConnector()
+                .forProjectDirectory(projectDir)
+            connector.connect().use {
+                it.newBuild().forTasks("build").run()
+            }
+        } catch (e: Exception) {
+            e.printStackTrace()
```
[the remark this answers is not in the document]
asodja: Something like that?
```
            it.newBuild().forTasks("build")
                .addArguments("--info")
                .setStandardError(System.err)
                .setStandardOutput(System.out)
                .run()```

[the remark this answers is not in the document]
wolfs: That would always print the output, right? I am not sure how useful this would be. I'd like to print the output only when the task fails.

I don't think `--info` is necessary.

wolfs (approved):
LGTM!

PR #389, opened 2021-12-23
Use a separate job to publish the released version to SDKman

wolfs:


.teamcity/settings/GradleProfilerPublishing.kt
```diff
@@ -34,7 +37,7 @@ object GradleProfilerPublishing : BuildType({
 
     steps {
         gradle {
-            tasks = "clean publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag releaseToSdkMan %additional.gradle.parameters%"
+            tasks = "clean createBuildReceipt publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag releaseToSdkMan %additional.gradle.parameters%"
```
asodja, line 40: Here we still have releaseToSdkMan, is that ok?

wolfs: No, it is not. I removed it.

buildSrc/src/main/kotlin/extensions.kt
```diff
@@ -0,0 +1 @@
+const val buildReceiptName = "build-receipt.properties"
```
asodja, line 1: This value is duplicated in  buildSrc/src/main/kotlin/profiler.versioning.gradle.kts. Can it be reused?

buildSrc/src/main/kotlin/plugins/RootProjectVersionPlugin.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.*
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.util.Properties
+
+const val incomingBuildReceiptLocation = "incoming-distributions/$buildReceiptName"
+
+open class RootProjectVersionPlugin : Plugin<Project> {
```
asodja, line 14: Could that be a precompiled script plugin?

buildSrc/src/main/kotlin/plugins/RootProjectVersionPlugin.kt
```diff
@@ -0,0 +1,42 @@
...
+        target.tasks.register("createBuildReceipt") {
+            outputs.file(createdBuildReceipt).withPropertyName("buildReceipt")
+            inputs.property("version", versionInfo.version)
+            doLast {
+                createdBuildReceipt.get().asFile.writeText("version=${versionInfo.version.get()}", StandardCharsets.UTF_8)
+            }
+        }
+
+        target.gradle.taskGraph.whenReady {
+            if (hasTask(":publishToSonatype") || hasTask(":releaseToSdkMan"))
```
asodja, line 38: `if` without `{` is easy to miss. Or I am just not used to it?

asodja (approved):
Added some comments, feel free to apply them if you think they make sense.

buildSrc/src/main/kotlin/extensions.kt
wolfs: Well spotted! I removed it in `profiler.versioning.gradle.kts`.

buildSrc/src/main/kotlin/plugins/RootProjectVersionPlugin.kt
wolfs: Yeah. I converted it.

buildSrc/src/main/kotlin/plugins/RootProjectVersionPlugin.kt
wolfs: Good point. I added the `{`.

asodja (approved):

PR #406, opened 2022-02-28
Add support to run profiler with IntelliJ

asodja:
Unofficial support for IntelliJ, tested on MacOS with 2021.3.2 and 2022.1 EAP locally.

### DIFF
--- a/src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
+++ b/src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
@@ -34,6 +34,10 @@ public class StudioConfigurationProvider {
             .map(studioInstallDir::resolve)
             .collect(Collectors.toList());
         Path javaCommand = studioInstallDir.resolve("jre/bin/java");
+        if (!javaCommand.toFile().exists()) {
+            // Try IntelliJ location
+            javaCommand = studioInstallDir.resolve("jbr/bin/java");
+        }
         Map<String, String> systemProperties = ImmutableMap.<String, String>builder()
             .put("idea.vendor.name", "Google")
             .put("idea.executable", "studio")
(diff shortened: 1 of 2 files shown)


src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
```diff
@@ -59,6 +63,10 @@ private static StudioConfiguration getMacOSConfiguration(Path studioInstallDir)
         Map<String, String> systemProperties = mapValues(jvmOptions.dict("Properties").toMap(), v -> v.replace("$APP_PACKAGE", actualInstallDir.toString()));
         String mainClass = jvmOptions.string("MainClass");
         Path javaCommand = actualInstallDir.resolve("Contents/jre/Contents/Home/bin/java");
```
wolfs, line 65: Can you add the possible paths to a list and then find the first one which exits? And if none of the paths exists, does that mean that the build should fail?

wolfs (commented):
Looks good! I left a comment.

src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
asodja: Yes, I can do that. I think we can fail in that case. I don't see a way that it would work without the valid path

src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
```diff
@@ -20,6 +21,18 @@
 
 public class StudioConfigurationProvider {
 
+    private static final List<String> DEFAULT_JAVA_PATHS = ImmutableList.<String>builder()
+        // MacOS Studio and IntelliJ paths
+        .add("Contents/jre/Contents/Home/bin/java")
```
wolfs, line 26: I am not sure if it makes sense to have all Java paths in one list or to have multiple lists, on per OS. WDYT?

wolfs (approved):

src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
asodja: I split the paths. Even though finding the wrong path for OS is very unlikely, it still might save some weird "gotcha" in the future.

wolfs (approved):
LGTM!

PR #407, opened 2022-02-28
Upgrade versions to support building and running on M1 macs

adammurdoch:
Upgrade nodejs and async-profiler versions.

### DIFF
--- a/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDownload.java
+++ b/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDownload.java
@@ -1,7 +1,10 @@
 package org.gradle.profiler.asyncprofiler;
 
 import com.google.common.collect.ImmutableSet;
+import org.apache.commons.compress.archivers.ArchiveEntry;
+import org.apache.commons.compress.archivers.ArchiveInputStream;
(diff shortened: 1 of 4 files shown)


src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDownload.java
```diff
@@ -69,28 +79,50 @@ static File defaultHome() {
...
+
+    private static void extract(ArchiveInputStream archiveStream, Path destDir) throws IOException {
+        ArchiveEntry entry;
+        while ((entry = archiveStream.getNextEntry()) != null) {
+            if (entry.isDirectory()) {
+                continue;
+            }
+            String name = entry.getName();
+            Path file = destDir.resolve(name);
+            if (!file.startsWith(destDir)) {
```
asodja, line 102: I think this might fix it:
```suggestion
            if (!file.normalize().startsWith(destDir)) {
```

I tried it in https://github.com/gradle/gradle-profiler/pull/408 and no issue is reported.

ljacomet (commented):
I confirmed that this fixes running the build on M1, so all good on that front!

adammurdoch:
Thanks for reviewing

PR #409, opened 2022-03-03
Add support for running YourKit on Windows

sambsnyd:


src/main/java/org/gradle/profiler/yourkit/YourKit.java
```diff
@@ -35,6 +37,12 @@ public static File findControllerJar() {
 
     public static File findJniLib() {
         File yourKitHome = findYourKitHome();
+        if(OperatingSystem.isWindows()) {
+            if("32".equals(System.getProperty("sun.arch.data.model"))) {
```
asodja, line 41: Can we use "os.arch" instead of "sun.arch.data.model"? "sun.arch.data.model" seems to be tied to sun so I am not sure if all JVM report that property.

I think having a method like `OperatingSystem.isLinuxX86()` but for windows would make sense.

asodja: On second thought, I don't think we need to support 32bit.

asodja: @sambsnyd Or you maybe need support for 32bit?

asodja:
Verified on Windows 64bit.

asodja:
Thank you for your contribution!

PR #410, opened 2022-03-15
Simplify Android Studio download

wolfs:
By adding a dependency notation from a provider instead of using `withDependencies`.

### DIFF
--- a/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
@@ -31,21 +31,14 @@ val extension = extensions.create<AndroidStudioTestExtension>("androidStudioTest
 val androidStudioRuntime by configurations.creating
 
 dependencies {
-    val extension = when {
+    val fileExtension = when {
         isWindows() -> "windows.zip"
         isMacOS() && isIntel() -> "mac.zip"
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -31,21 +31,14 @@ val extension = extensions.create<AndroidStudioTestExtension>("androidStudioTest
...
     }
-    androidStudioRuntime("android-studio:android-studio@$extension")
-}
-androidStudioRuntime.withDependencies {
-    this.forEach { dependency ->
-        if (dependency is ExternalDependency && dependency.version == null) {
-            dependency.version { require(extension.testAndroidStudioVersion.get()) }
-        }
-    }
+    androidStudioRuntime(extension.testAndroidStudioVersion.map { version -> "android-studio:android-studio:${version}@$fileExtension" })
```
asodja, line 41: Can ${version} be withour {}, just $version?

asodja (approved):

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
wolfs: Good point, done.

PR #418, opened 2022-03-23
Chrome trace data with backslash fix #293

### LINKED ISSUE #293: Chrome trace data with backslash cannot be opened on macOS
I'm assuming that Chrome just won't open those chrome traces on other operating systems as well. The key point here is chrome trace generated on Windows most likely will contain backslash because of the file path separator. This causes an error when opening in the lines of "Unexpected token...".

I was able to "fix" the chrome trace by doubling all backslash in the embedded JSON data. It would be much better if the profiler could generate valid JSON data without needing manual editing.

LoriLori:
Fix for https://github.com/gradle/gradle-profiler/issues/293

### DIFF
--- a/subprojects/chrome-trace/src/main/java/org/gradle/trace/CountEvent.java
+++ b/subprojects/chrome-trace/src/main/java/org/gradle/trace/CountEvent.java
@@ -1,5 +1,7 @@
 package org.gradle.trace;
 
+import org.gradle.trace.util.JsonUtil;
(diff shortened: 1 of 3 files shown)


asodja:
We merged a PR that solves the same issue https://github.com/gradle/gradle-profiler/pull/463, so I will close this PR.

Even though we didn't merge it, thank you for your contribution.

PR #432, opened 2022-06-17
Fix command warm-ups for CLI

carlonzo:
`--warm-ups` is not recognized. `--warmups` is


carlonzo:
hi thanks! I don't quite understand why its a breaking change.
CLI today is not accepting `--warm-ups` and `gradle-profiler --help` suggests to use `--warmups`, thats why I proposed a change to the README. 
here is the current parser code https://github.com/gradle/gradle-profiler/blob/ea32cd9a8d7d32f87cf6e150bf220638a292291f/src/main/java/org/gradle/profiler/CommandLineParser.java#L52

carlonzo:
ok understood what you mean. I was looking at the `warm-up` in scenarios section. the one for CLI is correct. sorry for wasting time

gabrielfeo:
Actually, I was mistaken too. Didn't notice you were only changing the README and not code, that's why I mentioned it'd be a breaking change. Seems I need more coffee before doing this late at night 😅

Issue #437, opened 2022-07-31
IDE sync profiling incompatible with Android Studio Electric Eel Canary 8


mikejuyoon:
We've been using Gradle-profiler to benchmark sync times with different Android Studio versions. Recently when we tried to benchmark Android Studio Electric Eel  (2022.1.1) Canary 8, we found a couple errors due to incompatibility. 
Thank you @rogerhu for the help with this investigation

**Versions:**
- [Android Studio Electric Eel (2022.1.1) Canary 8](https://developer.android.com/studio/archive](https://developer.android.com/studio/archive)
- Gradle Profiler version 0.18.0 or 1.19.0 (master)

**Errors**
1. `java.lang.ClassNotFoundException: com.intellij.openapi.util.SystemInfoRt`
    - Stack Trace: [https://gist.github.com/mikejuyoon/323f7ebf6868e864fb112c4adfe89307](https://gist.github.com/mikejuyoon/323f7ebf6868e864fb112c4adfe89307)
    - A possible fix: [https://github.com/gradle/gradle-profiler/pull/436/commits/d1fcb1bf1cd7839f3d50c620bcef5ae6153fc898](https://github.com/gradle/gradle-profiler/pull/436/commits/d1fcb1bf1cd7839f3d50c620bcef5ae6153fc898)
2.  Bigger, trickier issue: `java.lang.IncompatibleClassChangeError: Method 'com.intellij.util.messages.MessageBusConnection com.android.tools.idea.gradle.project.sync.GradleSyncState.subscribe(com.intellij.openapi.project.Project, com.android.tools.idea.gradle.project.sync.GradleSyncListener)' must be InterfaceMethodref constant`
    - Stack Trace: https://gist.github.com/mikejuyoon/e9ccfd98e4b6d3854ebd90e725c4fd00
    - Looks like the Android plugin source have changed on Android Studio EE but not in the Jetbrains code. So there is API incompatibility. i.e:
[...]

PR #439, opened 2022-08-04
Add support for Android Studio Chipmunk (2021.2), Dolphin (2021.3) and Electric Eel (2022.1)

asodja:
and IntelliJ with Java17 runtime. 

Fixes https://github.com/gradle/gradle-profiler/issues/437

### DIFF
--- a/src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
+++ b/src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
@@ -1,6 +1,7 @@
 package org.gradle.profiler.studio.launcher;
 
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import org.gradle.profiler.instrument.GradleInstrumentation;
 import org.gradle.profiler.studio.tools.StudioSandboxCreator.StudioSandbox;
 
@@ -15,6 +16,46 @@ import java.util.Map;
 public class LauncherConfigurationParser {
 
     private static final boolean SHOULD_RUN_HEADLESS = Boolean.getBoolean("studio.tests.headless");
+    private static final List<String> JAVA17_ADD_OPENS = ImmutableList.of(
+        "--add-opens=java.base/java.io=ALL-UNNAMED",
+        "--add-opens=java.base/java.lang=ALL-UNNAMED",
+        "--add-opens=java.base/java.lang.reflect=ALL-UNNAMED",
+        "--add-opens=java.base/java.net=ALL-UNNAMED",
+        "--add-opens=java.base/java.nio=ALL-UNNAMED",
+        "--add-opens=java.base/java.nio.charset=ALL-UNNAMED",
+        "--add-opens=java.base/java.text=ALL-UNNAMED",
+        "--add-opens=java.base/java.time=ALL-UNNAMED",
+        "--add-opens=java.base/java.util=ALL-UNNAMED",
+        "--add-opens=java.base/java.util.concurrent=ALL-UNNAMED",
+        "--add-opens=java.base/java.util.concurrent.atomic=ALL-UNNAMED",
+        "--add-opens=java.base/jdk.internal.vm=ALL-UNNAMED",
+        "--add-opens=java.base/sun.nio.ch=ALL-UNNAMED",
+        "--add-opens=java.base/sun.security.ssl=ALL-UNNAMED",
+        "--add-opens=java.base/sun.security.util=ALL-UNNAMED",
+        "--add-opens=java.desktop/com.sun.java.swing.plaf.gtk=ALL-UNNAMED",
+        "--add-opens=java.desktop/java.awt=ALL-UNNAMED",
+        "--add-opens=java.desktop/java.awt.dnd.peer=ALL-UNNAMED",
+        "--add-opens=java.desktop/java.awt.event=ALL-UNNAMED",
+        "--add-opens=java.desktop/java.awt.image=ALL-UNNAMED",
+        "--add-opens=java.desktop/java.awt.peer=ALL-UNNAMED",
+        "--add-opens=java.desktop/javax.swing=ALL-UNNAMED",
+        "--add-opens=java.desktop/javax.swing.plaf.basic=ALL-UNNAMED",
+        "--add-opens=java.desktop/javax.swing.text.html=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.awt.X11=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.awt.datatransfer=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.awt.image=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.awt=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.font=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.java2d=ALL-UNNAMED",
+        "--add-opens=java.desktop/sun.swing=ALL-UNNAMED",
+        "--add-opens=jdk.attach/sun.tools.attach=ALL-UNNAMED",
+        "--add-opens=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED",
+        "--add-opens=jdk.internal.jvmstat/sun.jvmstat.monitor=ALL-UNNAMED",
+        "--add-opens=jdk.jdi/com.sun.tools.jdi=ALL-UNNAMED",
+        "--add-exports=java.desktop/com.apple.eawt=ALL-UNNAMED",
+        "--add-exports=java.desktop/com.apple.laf=ALL-UNNAMED",
+        "--add-exports=java.desktop/com.apple.eawt.event=ALL-UNNAMED"
+    );
 
     private final Path studioInstallDir;
     private final StudioSandbox studioSandbox;
@@ -67,6 +108,8 @@ public class LauncherConfigurationParser {
         commandLine.add("-cp");
         commandLine.add(Joiner.on(File.pathSeparator).join(classpath));
         systemProperties.forEach((key, value) -> commandLine.add(String.format("-D%s=%s", key, value)));
+        // Support for IntelliJ with Java17 runtime
+        commandLine.addAll(JAVA17_ADD_OPENS);
         if (enableStudioAgentParameters) {
             commandLine.add(String.format("-javaagent:%s=%s,%s", agentJar, studioAgentPort, supportJar));
             commandLine.add("--add-exports");
@@ -97,6 +140,9 @@ public class LauncherConfigurationParser {
         systemProperties.put("idea.log.path", studioSandbox.getLogsDir().toString());
         // Newer IntelliJ versions require this property to avoid trust project popup
         systemProperties.put("idea.trust.all.projects", "true");
+        // Used so wrapper init script is not run by Android Studio.
+        // We anyway override installation in the org.gradle.profiler.studio.instrumented.Interceptor.
+        systemProperties.put("idea.gradle.distributionType", "BUNDLED");
         if (SHOULD_RUN_HEADLESS) {
             systemProperties.put("java.awt.headless", "true");
         }
--- a/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
+++ b/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
@@ -26,8 +26,10 @@ public class GradleProfilerProjectManagerListener implements ProjectManagerListe
     public void projectOpened(@NotNull com.intellij.openapi.project.Project project) {
         LOG.info("Project opened");
         if (System.getProperty(PROFILER_PORT_PROPERTY) != null) {
-            // Skip Android Studio Gradle automatic import on project open,
-            // interestingly this is an issue only with Android Studio but not IntelliJ
+            // This solves the issue where Android Studio would run the Gradle sync automatically on the first import.
+            // Unfortunately it seems we can't always detect it because it happens very late and due to that there might
+            // a case where two simultaneous syncs would be run: one from automatic sync trigger and one from our trigger.
+            // With this line we disable that automatic sync, but we still trigger our sync later in the code.
             GradleProjectInfo.getInstance(project).setSkipStartupActivity(true);
             TrustedProjects.setTrusted(project, true);
             ApplicationManager.getApplication().executeOnPooledThread(() -> {
--- a/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
+++ b/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
@@ -1,5 +1,6 @@
 package org.gradle.profiler.studio.plugin;
 
+import com.android.tools.idea.gradle.project.GradleProjectInfo;
 import com.intellij.ide.impl.TrustedProjects;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
@@ -25,6 +26,9 @@ public class GradleProfilerProjectManagerListener implements ProjectManagerListe
     public void projectOpened(@NotNull com.intellij.openapi.project.Project project) {
         LOG.info("Project opened");
         if (System.getProperty(PROFILER_PORT_PROPERTY) != null) {
+            // Skip Android Studio Gradle automatic import on project open,
+            // interestingly this is an issue only with Android Studio but not IntelliJ
+            GradleProjectInfo.getInstance(project).setSkipStartupActivity(true);
             TrustedProjects.setTrusted(project, true);
             ApplicationManager.getApplication().executeOnPooledThread(() -> {
                 StudioRequest lastRequest = listenForSyncRequests(project);
@@ -43,5 +47,4 @@ public class GradleProfilerProjectManagerListener implements ProjectManagerListe
             throw new UncheckedIOException(e);
         }
     }
-
 }
--- a/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
+++ b/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
@@ -15,7 +15,6 @@ import org.gradle.profiler.studio.plugin.system.GradleProfilerGradleSyncListener
 
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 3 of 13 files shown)


src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
```diff
@@ -67,6 +108,8 @@ public LaunchConfiguration calculate() {
         commandLine.add("-cp");
         commandLine.add(Joiner.on(File.pathSeparator).join(classpath));
         systemProperties.forEach((key, value) -> commandLine.add(String.format("-D%s=%s", key, value)));
+        // Support for IntelliJ with Java17 runtime
+        commandLine.addAll(JAVA17_ADD_OPENS);
```
lptr, line 112: Should we always add these, or only when Java version is >= 17?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
```diff
@@ -25,6 +26,9 @@ public class GradleProfilerProjectManagerListener implements ProjectManagerListe
     public void projectOpened(@NotNull com.intellij.openapi.project.Project project) {
         LOG.info("Project opened");
         if (System.getProperty(PROFILER_PORT_PROPERTY) != null) {
+            // Skip Android Studio Gradle automatic import on project open,
+            // interestingly this is an issue only with Android Studio but not IntelliJ
+            GradleProjectInfo.getInstance(project).setSkipStartupActivity(true);
```
lptr, line 31: How does this affect benchmarks? Are we doing the import during the first warmup round perhaps, or not at all?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
```diff
@@ -8,29 +10,40 @@
...
+import static org.gradle.profiler.studio.plugin.system.AndroidStudioSystemHelper.waitOnPostStartupActivities;
 import static org.gradle.profiler.studio.plugin.system.AndroidStudioSystemHelper.waitOnPreviousGradleSyncFinish;
 
 public class GradleProfilerClient {
 
     private static final Logger LOG = Logger.getInstance(GradleProfilerClient.class);
 
     private final Client client;
+    private final Stopwatch startupStopwatch;
+    private final AtomicInteger syncCount;
```
lptr, line 36: Using an `AtomicInteger` here suggests some sort of parallelism, but surely we don't run benchmarks in parallel, right? Can we make this a regular `int` instead?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
```diff
@@ -63,19 +76,44 @@ private void handleGradleProfilerRequest(StudioRequest request, Project project)
...
+        }
+        GradleSyncResult result;
+        if (lastSyncResult.isSuccessful() && (lastSyncResult == SKIPPED || lastSyncResult == SKIPPED_OUT_OF_DATE)) {
+            result = new GradleSyncResult(StudioSyncRequestResult.SKIPPED, "");
+        } else if (lastSyncResult.isSuccessful()) {
+            result = new GradleSyncResult(StudioSyncRequestResult.SUCCEEDED, "");
+        } else  {
+            result = new GradleSyncResult(StudioSyncRequestResult.FAILED, "Startup failure");
+        }
+        return result;
```
lptr, line 116: We could simplify this a bit, for example there's no need to check for `isSuccessful()` when dealing with the skipped stated.

```suggestion
        if (lastSyncResult == UNKNOWN) {
            // Sync was not run before, we need to run it manually
            return startManualSync(project);
        } else if (lastSyncResult == SKIPPED || lastSyncResult == SKIPPED_OUT_OF_DATE) {
            return new GradleSyncResult(StudioSyncRequestResult.SKIPPED, "");
        } else if (lastSyncResult.isSuccessful()) {
            return new GradleSyncResult(StudioSyncRequestResult.SUCCEEDED, "");
        } else  {
            return new GradleSyncResult(StudioSyncRequestResult.FAILED, "Startup failure");
        }
```

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
```diff
@@ -1,32 +1,40 @@
...
         try {
-            GradleSyncInvoker.getInstance().requestProjectSync(project, TRIGGER_USER_SYNC_ACTION, syncListener);
-        } catch (Exception e) {
-            e.printStackTrace();
-            syncListener.syncFailed(project, e.getMessage());
+            // We could get sync result from the `Future` returned by the syncProject(),
+            // but it doesn't return error message so we rather listen to GRADLE_SYNC_TOPIC to get the sync result
+            ProjectSystemUtil.getSyncManager(project).syncProject(ProjectSystemSyncManager.SyncReason.USER_REQUEST).get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new RuntimeException(e);
```
lptr, line 37: Should we still call `synclistener.syncFailed()` here, or is it okay to just throw?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
```diff
@@ -35,38 +43,54 @@ public static GradleSyncResult doGradleSync(Project project) {
      * Registers a listener that waits on next gradle sync if it's in progress.
      */
     public static void waitOnPreviousGradleSyncFinish(Project project) {
-        GradleProfilerGradleSyncListener syncListener = new GradleProfilerGradleSyncListener();
-        MessageBusConnection connection = GradleSyncState.subscribe(project, syncListener);
-        if (!GradleSyncState.getInstance(project).isSyncInProgress()) {
+        MessageBusConnection connection = project.getMessageBus().connect();
+        CompletableFuture<String> future = new CompletableFuture<>();
+        connection.subscribe(PROJECT_SYSTEM_SYNC_TOPIC, syncResult -> future.complete("Done"));
```
lptr, line 48: Shouldn't this only happen after checking if a sync is already in progress?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
```diff
@@ -35,38 +43,54 @@ public static GradleSyncResult doGradleSync(Project project) {
...
+    }
+
+    public static void waitOnPostStartupActivities(Project project) {
+        wait(() -> !StartupManager.getInstance(project).postStartupActivityPassed(), WAIT_ON_STARTUP_SLEEP_TIME);
+    }
+
     @SuppressWarnings("BusyWait")
-    private static void waitOn(ProgressIndicator progressIndicator) {
-        while (progressIndicator.isRunning()) {
+    private static void wait(Supplier<Boolean> whileCondition, long sleepMillis) {
```
lptr, line 90: Nit: I think it results in somewhat nicer code at the call site if you put the lambda as the last parameter.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/GradleProfilerGradleSyncListener.java
```diff
@@ -48,6 +53,8 @@ public GradleSyncResult waitAndGetResult() {
             return results.take();
         } catch (InterruptedException e) {
             throw new RuntimeException(e);
+        } finally {
+            connection.disconnect();
```
lptr, line 57: Seems like a code smell to close a connection in a place other than where it was opened. Instead in `AndroidStudioSystemHelper.doGradleSync()` I'd inline `subscribeToGradleSync()` and put a `try` around the `connection` object there.

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
```diff
@@ -38,14 +40,14 @@ class StudioPluginIntegrationTest extends StudioPluginTestCase {
...
-        settingsFile.delete()
+        settingsFile << "garbage"
 
         when:
         connection.send(new StudioRequest(SYNC))
 
         then:
         StudioSyncRequestCompleted requestCompleted = connection.receiveSyncRequestCompleted(Duration.ofSeconds(90))
-        requestCompleted.result == FAILED
+        assert requestCompleted.result == FAILED
```
lptr, line 50: Why do you need an `assert` here?

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
```diff
@@ -38,14 +40,14 @@ class StudioPluginIntegrationTest extends StudioPluginTestCase {
         given:
         ServerConnection connection = server.waitForIncoming(Duration.ofSeconds(10))
         // Fail Gradle build by removing settings file
-        settingsFile.delete()
+        settingsFile << "garbage"
```
lptr, line 43: This now does not match the comment above. Also, instead of appending `garbage` you could do:
```suggestion
        settingsFile.text = "garbage"
```

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
```diff
@@ -57,7 +59,9 @@ class StudioPluginIntegrationTest extends StudioPluginTestCase {
         ServerConnection connection = server.waitForIncoming(Duration.ofSeconds(10))
 
         when:
-        def invalidateFile = Paths.get(PathManager.getSystemPath(), "projectModelCache", ".invalidate").toFile()
+        def invalidateFile = ApplicationInfo.instance.majorVersion == "2021" && ApplicationInfo.instance.minorVersionMainPart <= "1"
```
lptr, line 62: Probably doesn't matter much, but doing `<=` with a string doesn't sound safe. How about:

```suggestion
        def invalidateFile = ApplicationInfo.instance.majorVersion == "2021" && ApplicationInfo.instance.minorVersionMainPart in ["0", "1"]
```

src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
asodja: I tested with different Android Studio versions that have Java11 runtime and it works. I believe these modules are anyway accessible in Java11 by default.

We could open it just for Java17, but it means we need to figure out what JDK version an IDE has bundled (probably for every operating system separately) before we run it. So I would add it always unless we find out there are problems with it.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
asodja: For context: this solves the issue where Android Studio would run the Gradle sync automatically on first import. Unfortunately it seems we can't always detect it. And when we trigger a sync it could happen that two Gradle connections would be started: one for automatic sync and one for our sync. That resulted that two syncs were run simultaneously.

With this line we disable that automatic sync, but we still trigger a sync on first warmup. It's just that this workarounds an issue with possible two simultaneously syncs on that first warmup import.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
asodja: Yeah, it's not parallel. We can also use`int`

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
asodja: thanks

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
asodja: Maybe it's better to call sync failed, to not crash our plugin.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
asodja: We can also do it that way, changed

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
asodja: Unfortunately, Java is not Groovy or Kotlin to have any advantage from that, but changed anyway

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/GradleProfilerGradleSyncListener.java
asodja: We can also do that, done

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
asodja: sure, why not

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
asodja: thanks

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
asodja: It seems Spock is not wired correctly somehow, so it doesn't enhance statement with asserts.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
asodja: Changed to int

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/GradleProfilerProjectManagerListener.java
lptr: Could you add this explanation in the code please?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
```diff
@@ -1,32 +1,44 @@
...
+    private static final long WAIT_ON_STARTUP_SLEEP_TIME = 100;
 
     /**
      * Does a Gradle sync.
      */
     public static GradleSyncResult doGradleSync(Project project) {
+        MessageBusConnection connection = project.getMessageBus().connect();
         GradleProfilerGradleSyncListener syncListener = new GradleProfilerGradleSyncListener();
+        connection.subscribe(GradleSyncState.GRADLE_SYNC_TOPIC, syncListener);
         try {
```
lptr, line 34:
```suggestion
        MessageBusConnection connection = project.getMessageBus().connect();
        try {
            GradleProfilerGradleSyncListener syncListener = new GradleProfilerGradleSyncListener();
            connection.subscribe(GradleSyncState.GRADLE_SYNC_TOPIC, syncListener);
```

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
```diff
@@ -15,6 +17,10 @@ import static org.gradle.profiler.client.protocol.messages.StudioRequest.StudioR
 import static org.gradle.profiler.client.protocol.messages.StudioSyncRequestCompleted.StudioSyncRequestResult.FAILED
 import static org.gradle.profiler.client.protocol.messages.StudioSyncRequestCompleted.StudioSyncRequestResult.SUCCEEDED
 
+/**
+ * Note: This class doesn't extend Spock Specification, but it extends IntelliJ HeavyPlatformTestCase,
+ * so you should always use `assert` when asserting statements
+ */
```
lptr, line 23: This sounds dangerous for future changes. Shall we drop using Spock altogether (including `when:` and `then:`)?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
asodja: We can't move `GradleProfilerGradleSyncListener syncListener = new GradleProfilerGradleSyncListener();` since it's used later, but we can move connection subscribe.

asodja: I changed the code to call listener directly from `try {}`

subprojects/studio-plugin/src/test/groovy/org/gradle/profiler/studio/plugin/StudioPluginIntegrationTest.groovy
asodja: Changed, now we use Spock `Specification`, but we setup Ide via Composition, WDYT?

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
lptr: We are now back to throwing an exception instead of `syncListener.failed()`, is this intentional?

asodja: Hmm, it seems github is showing outdated version here. Current version is here:
https://github.com/gradle/gradle-profiler/pull/439/files#diff-8eb36c0b3730fb2241bb7c46b8367c29778fcf7971216d69e1fb5df5d9938d1e

PR #440, opened 2022-08-09
Upgrade to latest gradle enterprise plugin

liutikas:


asodja:
Thanks for the contribution. Could you just sign-off the commit to make DCO check happy?

liutikas:
Done.

PR #442, opened 2022-08-21
Fix parsing classpath for Bumblebee, Chipmunk and Dolphin on Linux and Windows

asodja:
If this will be still require adjustments with new versions in the future, we could run `start.sh` or `start.bat` directly, but we would need to check what env. variables we need to set for custom AS args.

We don't have AS multi version tests, but I ran a tests with Bumblebee and all pass, here is the build:
https://builds.gradle.org/buildConfiguration/GradleProfiler_GradleProfilerTestTrigger/55190392?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

### DIFF
--- a/src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
+++ b/src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
@@ -31,6 +31,10 @@ public class StudioConfigurationProvider {
     private static final List<String> DEFAULT_MACOS_JAVA_PATHS = ImmutableList.of("Contents/jre/Contents/Home/bin/java", "Contents/jbr/Contents/Home/bin/java");
     private static final List<String> DEFAULT_WINDOWS_JAVA_PATHS = ImmutableList.of("jre/bin/java.exe", "jbr/bin/java.exe");
     private static final List<String> DEFAULT_LINUX_JAVA_PATHS = ImmutableList.of("jre/bin/java", "jbr/bin/java");
+
+    /**
+     * Android Studio Dolphin (2021.3) and older versions use "CLASSPATH", while Electric Eel (2022.1) and newer versions use CLASS_PATH
+     */
     private static final Pattern LINUX_CLASSPATH_LIB_PATTERN = Pattern.compile(".*(CLASS_PATH|CLASSPATH)=.*(?<lib>lib/.+\\.jar).*");
     private static final Pattern WINDOWS_CLASSPATH_LIB_PATTERN = Pattern.compile(".*(CLASS_PATH|CLASSPATH)=.*(?<lib>lib\\\\.+\\.jar).*");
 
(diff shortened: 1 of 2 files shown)


src/test/groovy/org/gradle/profiler/studio/launcher/StudioConfigurationProviderTest.groovy
```diff
@@ -41,16 +41,19 @@ some other text
             studioInstallDir.toPath().resolve("lib/lib2.jar"),
             studioInstallDir.toPath().resolve("lib/lib1/lib1.jar")
         ]
+
+        where:
+        classpathKeyword << ["CLASS_PATH", "CLASSPATH"]
```
lptr, line 46: Let's document somewhere which keyword belongs to the old vs. new versions.

asodja: Good idea, added a comment.

src/main/java/org/gradle/profiler/studio/launcher/StudioConfigurationProvider.java
```diff
@@ -31,8 +31,12 @@ public class StudioConfigurationProvider {
...
     private static final List<String> DEFAULT_WINDOWS_JAVA_PATHS = ImmutableList.of("jre/bin/java.exe", "jbr/bin/java.exe");
     private static final List<String> DEFAULT_LINUX_JAVA_PATHS = ImmutableList.of("jre/bin/java", "jbr/bin/java");
-    private static final Pattern LINUX_CLASSPATH_LIB_PATTERN = Pattern.compile(".*CLASS_PATH=.*(?<lib>lib/.+\\.jar).*");
-    private static final Pattern WINDOWS_CLASSPATH_LIB_PATTERN = Pattern.compile(".*CLASS_PATH=.*(?<lib>lib\\\\.+\\.jar).*");
+
+    /**
+     * Android Studio Dolphin (2021.3) and older versions use "CLASSPATH", while Electric Eel (2022.1) and newer versions use CLASS_PATH
+     */
+    private static final Pattern LINUX_CLASSPATH_LIB_PATTERN = Pattern.compile(".*(CLASS_PATH|CLASSPATH)=.*(?<lib>lib/.+\\.jar).*");
+    private static final Pattern WINDOWS_CLASSPATH_LIB_PATTERN = Pattern.compile(".*(CLASS_PATH|CLASSPATH)=.*(?<lib>lib\\\\.+\\.jar).*");
```
lptr, line 39: Nit: instead of `(CLASS_PATH|CLASSPATH)` you could use `(?:CLASS_PATH|CLASSPATH)` to make it obvious that this group is not captured.

asodja: Fixed

PR #444, opened 2022-08-28
Fix build warnings after Gradle 7.5 update

nevack:
These changes are trivial.

### DIFF
--- a/buildSrc/src/main/kotlin/profiler.publication.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.publication.gradle.kts
@@ -60,7 +60,7 @@ publishing {
     }
 }
 
-val pgpSigningKey: Provider<String> = providers.environmentVariable("PGP_SIGNING_KEY").forUseAtConfigurationTime()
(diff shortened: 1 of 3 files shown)


nevack:
Signed Off all commits

asodja:
Hi @nevack,

thanks for the contribution! It looks good, except that we still have to build for Java8. Due to that could you drop the last commit and force push it again?

JDK11 is required for building a project since `org.jetbrains.intellij` plugin that is used in `studio-plugin` project  is compiled for Java11. But `studio-plugin` is not used directly by the profiler, but it's used by Android Studio when you profile project sync with Android Studio.

nevack:
That's what I expected and this is the reason this change is in the last commit - to drop it if it's not right.
Thanks for clarification.

Maybe it's worth mentioning in README that you need both JDK 8 and JDK 11 to build the project?

asodja:
Thanks! Yeah, if you would like to update that, feel free to do so.

asodja:
Thanks for the contribution!

PR #449, opened 2022-09-13
Include build operation occurrence counts in tracing information

alllex:
TBD

### DIFF
--- a/src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
+++ b/src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
@@ -1,7 +1,9 @@
 package org.gradle.profiler;
 
 import org.gradle.profiler.BuildAction.BuildActionResult;
+import org.gradle.profiler.buildops.BuildOperationDuration;
 import org.gradle.profiler.result.BuildInvocationResult;
+import org.gradle.profiler.result.DurationOnlySample;
 import org.gradle.profiler.result.Sample;
 
 import javax.annotation.Nullable;
@@ -13,7 +15,7 @@ import static org.gradle.profiler.buildops.BuildOperationUtil.getSimpleBuildOper
 public class GradleBuildInvocationResult extends BuildInvocationResult {
     private final Duration garbageCollectionTime;
     private final Duration timeToTaskExecution;
-    private final Map<String, Duration> cumulativeBuildOperationTimes;
+    private final Map<String, BuildOperationDuration> cumulativeBuildOperationDurations;
     private final String daemonPid;
 
     public GradleBuildInvocationResult(
@@ -21,13 +23,13 @@ public class GradleBuildInvocationResult extends BuildInvocationResult {
         BuildActionResult actionResult,
         @Nullable Duration garbageCollectionTime,
         @Nullable Duration timeToTaskExecution,
-        Map<String, Duration> cumulativeBuildOperationTimes,
+        Map<String, BuildOperationDuration> cumulativeBuildOperationDurations,
         String daemonPid
     ) {
         super(buildContext, actionResult);
         this.garbageCollectionTime = garbageCollectionTime;
         this.timeToTaskExecution = timeToTaskExecution;
-        this.cumulativeBuildOperationTimes = cumulativeBuildOperationTimes;
+        this.cumulativeBuildOperationDurations = cumulativeBuildOperationDurations;
         this.daemonPid = daemonPid;
     }
 
@@ -43,8 +45,8 @@ public class GradleBuildInvocationResult extends BuildInvocationResult {
         return timeToTaskExecution;
     }
 
-    public Map<String, Duration> getCumulativeBuildOperationTimes() {
-        return cumulativeBuildOperationTimes;
+    public Map<String, BuildOperationDuration> getCumulativeBuildOperationDurations() {
+        return cumulativeBuildOperationDurations;
     }
 
     public static Sample<GradleBuildInvocationResult> sampleBuildOperation(String buildOperationDetailsClass) {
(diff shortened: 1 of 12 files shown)


src/main/java/org/gradle/profiler/buildops/BuildOperationDuration.java
```diff
@@ -0,0 +1,25 @@
+package org.gradle.profiler.buildops;
+
+import java.time.Duration;
+
+/**
+ * Carries total duration and count of one or more invocations of a build operation.
+ */
+public class BuildOperationDuration {
```
lptr, line 8: How about we call this `BuildOperationExecutionData` or something? "Duration" is only part of the picture.

src/main/java/org/gradle/profiler/result/DurationOnlySample.java
```diff
@@ -0,0 +1,8 @@
+package org.gradle.profiler.result;
+
+public interface DurationOnlySample<T extends BuildInvocationResult> extends Sample<T> {
```
lptr, line 3: How about `SingleInvocationSample`? Might not be worth adding a separate type for this, though.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
```diff
@@ -55,32 +57,39 @@ public String getName() {
             }
 
             @Override
-            public Duration extractFrom(GradleBuildInvocationResult result) {
-                return result.cumulativeBuildOperationTimes.getOrDefault(buildOperationDetailsClass, Duration.ZERO);
+            public Duration extractTotalDurationFrom(GradleBuildInvocationResult result) {
+                BuildOperationDuration duration = result.cumulativeBuildOperationDurations.get(buildOperationDetailsClass);
+                return duration == null ? Duration.ZERO : duration.getTotalDuration();
+            }
```
lptr, line 63: Nit: we could define a `BuildOperationDuration.ZERO` constant, and use it like we did with `Duration.ZERO`.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
```diff
@@ -43,8 +45,8 @@ public Duration getTimeToTaskExecution() {
         return timeToTaskExecution;
     }
 
-    public Map<String, Duration> getCumulativeBuildOperationTimes() {
-        return cumulativeBuildOperationTimes;
+    public Map<String, BuildOperationDuration> getCumulativeBuildOperationDurations() {
+        return cumulativeBuildOperationDurations;
```
lptr, line 49: We use "total" in other places to refer to the same concept we use "cumulative" to describe here. Would be nice to use the same.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
alllex: Good idea! I also realized that `BuildOperationDuration` can be immutable too, given the usages

src/main/java/org/gradle/profiler/buildops/BuildOperationDuration.java
alllex: Done

src/main/java/org/gradle/profiler/result/DurationOnlySample.java
alllex: Done

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
alllex: Done with a slight refactoring to make `BuildOperationExecutionData` immutable and accompanied by a builder.

src/main/java/org/gradle/profiler/GradleBuildInvocationResult.java
alllex: Done

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
```diff
@@ -87,44 +87,54 @@ public Optional<Duration> getTimeToTaskExecution() {
         }
         try {
             try (Stream<String> lines = Files.lines(configurationTimeDataFile.toPath(), StandardCharsets.UTF_8)) {
-                return lines
-                    .reduce((first, second) -> second)
-                    .map(line -> Duration.ofMillis(Long.parseLong(line)));
+                Optional<String> lastLine = lines.reduce((first, second) -> second);
+                return lastLine.map(line -> readExecutionDataLine(line).getTotalDuration());
```
lptr, line 91: Any reason to not use the chaining notation?

lptr (commented):
I don't get why adding the builder is useful for `BuildOperationExecutionData`. Apart from that it LGTM.

src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
alllex: Compared to other places where we parse the files, it's non-trivial that here we decide to ignore the whole file apart from the last line. IMO like this the intent is much more clear, than a sneaky `reduce` call.

alllex:
The builder story for `BuildOperationExecutionData` is a bit complicated.

If we want to have `BuildOperationExecutionData.ZERO` then the instances of `BuildOperationExecutionData` class need to be immutable (similar to why `Duration` is immutable to have a `ZERO`). Otherwise it's unsound design and could to lead to problems later.

Once `BuildOperationExecutionData` is immutable, we need a separate way of collecting total duration and count when reading a file with many samples. This requires a class with mutable fields, because we want to keep the parsing logic in one place... So one thing leads to another, and having a builder is the simplest way around all that.

lptr:
Ah, I missed that it's additive. If it's only used in one place, I think it'd make sense to move it there as a `private` subclass.

lptr (approved):
LGTM.

alllex:
We found out that the files with execution data always contain a single line. Given this assumption, the implementation for parsing has been simplified and reused in additional places. This also allowed to drop the builder for the `BuildOperationExecutionData`.

lptr (approved):
LGTM.

Issue #450, opened 2022-09-21
Incompatibility with Android Studio Electric Eel Canary 10+


mikejuyoon:
When running Gradle Profiler with AS Electric Eel Canary 10+ (including Beta1) we get the following error:
```
2022-08-31 19:59:07,675 [   7587] SEVERE - #c.i.o.a.i.ApplicationImpl - GRADLE_SYNC_TOPIC
java.lang.NoSuchFieldError: GRADLE_SYNC_TOPIC
        at org.gradle.profiler.studio.plugin.system.AndroidStudioSystemHelper.doGradleSync(AndroidStudioSystemHelper.java:35)
        at org.gradle.profiler.studio.plugin.client.GradleProfilerClient.startManualSync(GradleProfilerClient.java:92)
        at org.gradle.profiler.studio.plugin.client.GradleProfilerClient.getStartupSyncResult(GradleProfilerClient.java:104)
        at org.gradle.profiler.studio.plugin.client.GradleProfilerClient.handleSyncRequest(GradleProfilerClient.java:86)
        at org.gradle.profiler.studio.plugin.client.GradleProfilerClient.handleGradleProfilerRequest(GradleProfilerClient.java:63)
        at org.gradle.profiler.studio.plugin.client.GradleProfilerClient.listenForSyncRequests(GradleProfilerClient.java:46)
        at org.gradle.profiler.studio.plugin.GradleProfilerProjectManagerListener.listenForSyncRequests(GradleProfilerProjectManagerListener.java:47)
        at org.gradle.profiler.studio.plugin.GradleProfilerProjectManagerListener.lambda$projectOpened$0(GradleProfilerProjectManagerListener.java:36)
        at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:295)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:668)
        at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:665)
[...]

PR #451, opened 2022-09-27
Fix Android Studio syncs for Electric Eel Canary 10+

asodja:
Removes all usages of `GradleSyncState`. That class is modified in Android Studio code compared to IntelliJ. Hopefully, that would make it more future-proof.

Tested locally also with Bumblebee (2021.1.1) and Flamingo (2022.2.1).

Fixes https://github.com/gradle/gradle-profiler/issues/450.


alllex:
LGTM!

Issue #454, opened 2022-10-12
“--buildscan-version” flag doesn’t override the buildscan plugin version used by profiler


kenyee:
We're using an older version of the GE server and get this error when running gradle profiler when trying to upload buildscans because the gradle profiler is always a bit ahead of our updates:
```
The request was rejected.
Gradle Enterprise plugin version 3.10.2 is newer than the newest version supported by Gradle Enterprise 2022.1.1 which is 3.9. Please update to a newer version of Gradle Enterprise.
```

I found this flag in the integration tests: https://github.com/gradle/gradle-profiler/blob/cd4881ceb74b7c2c14f60c58c18d2239157f172b/src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy

but it doesn't seem to work even though in the profiler.log, it says:
Using build scan plugin 3.6
when I pass it as --buildscan-version 3.6

I also found this old closed issue where it was profiler supposedly doesn't override the buildscan plugin version if it's set already, but it still seems to: https://github.com/gradle/gradle-profiler/issues/58

asodja:
Checking the code, I am not sure where `gradle-profiler` would apply `3.10.2`. Maybe I missed something, but it looks to me gradle-profiler applies lower versions [here](https://github.com/gradle/gradle-profiler/blob/3f688b1eb8041c7068da74d160e6bc88c93bd4ca/src/main/java/org/gradle/profiler/buildscan/BuildScanProfiler.java#L19).

Other than that I think you are right, we probably should check [here](https://github.com/gradle/gradle-profiler/blob/3f688b1eb8041c7068da74d160e6bc88c93bd4ca/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java#L43) also if new plugin `com.gradle.enterprise` is applied.

kenyee:
Thanks..I was a bit puzzled on how 3.10.2 is being applied too...doesn't seem to be anywhere in this repo.
We're using 3.6 in our project but will update when we upgrade our GE server.

The current gradle profiler code should be already using the applied plugin according to that other issue/PR I linked.  And --buildscan-version should allow us to force an override but doesn't 🤔

kenyee:
FYI, what we've been doing to work around this is patch the latest release to do this to src/main/java/org/gradle/profiler/buildscan/BuildScanProfiler.java
```
-            if (buildConfiguration.getGradleVersion().compareTo(GRADLE_5) < 0) {
[...]

PR #455, opened 2022-10-18
Fix detection and handling of Gradle Enterprise plugin

asodja:
Fixes https://github.com/gradle/gradle-profiler/issues/454.

### DIFF
--- a/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java
+++ b/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java
@@ -36,10 +36,16 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
     private void generateInitScript() throws IOException {
         try (PrintWriter writer = new PrintWriter(new FileWriter(initScript))) {
             writer.println(
+                "settingsEvaluated {\n" +
+                    "   def detailsFile = new File(new URI('" + buildDetails.toURI() + "'))\n" +
+                    "   detailsFile.text = \"${it.pluginManager.hasPlugin('com.gradle.enterprise')}\\n\"\n" +
+                    "}\n"
+            );
+            writer.println(
                 "rootProject {\n" +
                     "  afterEvaluate {\n" +
                     "    def detailsFile = new File(new URI('" + buildDetails.toURI() + "'))\n" +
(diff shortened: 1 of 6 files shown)


src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -250,29 +250,111 @@ println "<daemon: " + gradle.services.get(org.gradle.internal.environment.Gradle
         assertBuildScanPublished()
     }
 
-    def "profiles build using Build Scans overridden version specified Gradle version and tasks"() {
+    def "uses Gradle Enterprise plugin version used by the build if present"() {
         given:
+        settingsFile.text = """
+plugins {
```
alllex, line 256: 💄 Indentation of the code code snippets here and in the rest of the tests.

I found that it does make a difference, when one needs to look at this tests at any future point and debug/fix them.

alllex (approved):
Looks good!
I left some comments on potential optional improvements, but it could be merged as it is now.

src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java
```diff
@@ -104,12 +110,13 @@ private GradleBuildConfiguration probe(GradleConnector connector) {
             JavaEnvironment javaEnvironment = buildEnvironment.getJava();
             List<String> allJvmArgs = new ArrayList<>(javaEnvironment.getJvmArguments());
             allJvmArgs.addAll(readSystemPropertiesFromGradleProperties());
+            boolean usesScanPlugin = Boolean.parseBoolean(buildDetails.get(0)) || Boolean.parseBoolean(buildDetails.get(2));
```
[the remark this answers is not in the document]
asodja: Good idea. I now read this file as Properties and I also wrap result in to a class.

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
asodja: I made indentation of ProfilerIntegrationTest consistent, but not for other classes :D.

alllex: Thanks!

PR #461, opened 2022-12-21
Apply default outputDir indexing logic to user-provided output dir

valeryvpetrov-dev:
# Problem
If I pass --output-dir argument with my custom value then I get report in pointed dir. But when I run profiler for the second time it overwrites previous report.

# Solution
If I do not pass --output-dir argument profiler then creates it by itself in profile-out-* dirs. So each run report is placed in separate dir. So why not do the same with custom output dir?

# Results
- [Without --output-dir option — same](https://drive.google.com/file/d/1nJAgM3PH7f0q4p6dmW678lHBz-ao4MqD/view?usp=share_link)
- [With custom --output-dir — new dirs with index are created](https://drive.google.com/file/d/1H9klKKC-KCai6hheJpCNnjvCvuIXH_cU/view?usp=share_link)


gabrielfeo:
Disclaimer: I am not a project member

IMHO, profiler should keep writing results directly to the `--output-dir`, without creating subdirs on its own. This way, output files have a predictable location, which is better for automation, e.g. running profiler and uploading the `outputdir/benchmark.csv` somewhere.

However, I've also overwritten profiler results by mistake as you described. Perhaps an alternative solution would be making profiler fail if `--output-dir` is not empty.

valeryvpetrov-dev:
@gabrielfeo good point about automation. When I used profiler for the first time I expect to have unified behavior. But it differs. I see three solutions:
- Make --ouput-dir generation of indexed dirs in case of custom value
- Make --output-dir generation to the same dir between runs for default value
- Update documentation for the case of custom value

ljacomet:
Sorry for the late reply

PR #463, opened 2023-01-10
Chrome traces: normalize file paths in build operation display names

### LINKED ISSUE #293: Chrome trace data with backslash cannot be opened on macOS

jjohannes:
Fixes #293

### DIFF
--- a/buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     id("profiler.java-library")
+    id("groovy")
 }
 
 // These projects are packaged into the Gradle profiler Jar, so let's make them reproducible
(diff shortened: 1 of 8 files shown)


asodja:
I wonder if we could add some tests, so we don't break it in future. 🤔

jjohannes:
I did initially not add a test, because there were no tests at all in the project and I didn't know where to put it.

I now added a commit that configures the subprojects to have Spock tests (feel free to adjust or move the test somewhere else).

I added a test that checks if the normalization works for all `Gradle*BuildOperationListenerInvocationHandler`s that accept `BuildOperationDescriptor`s as they are now. Hope that's sufficient for this.

asodja:
Awesome, thanks!

asodja:
Thanks for your contribution!

Issue #470, opened 2023-01-24
Android Studio sync doesn't work with Android Studio Giraffe


asodja:
Trying Canary 1 there are some issues:
1. If there are no local.properties there is Android sdk pop-up
2. It seems plugin is not started so profiler reports:
```
* ERROR
* Could not connect to Android Studio process started by the gradle-profiler.
* This might indicate that you are already running an Android Studio process in the same sandbox.
* Stop Android Studio manually in the used sandbox or use a different sandbox with --studio-sandbox-dir to isolate the process.
```

Issue #469, opened 2023-01-20
Test failures with Oracle Java 8, new Intel mac


blindpirate:
https://builds.gradle.org/buildConfiguration/GradleProfiler_MacOSJava8/60607566?hideProblemsFromDependencies=false&hideTestsFromDependencies=true

Seems to be some incompatibility on new Intel Mac (Ventura).

asodja:
So it seems we get error code 137 when starting an Android Studio from gradle-profiler. How we start Android Studio:
- we download Android Studio distribution (Gradle tasks)
- in profiler we start Android Studio as:
```
/Users/tcagent1/agent/work/7b79572c01edd19a/build/android-studio/Android Studio Preview.app/Contents/jbr/Contents/Home/bin/java -cp <classpath> com.intellij.idea.Main <test project>
```
via [ProcessBuilder](https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html). And it exits with error code 137, but I could get any message why is that.

I suspected that there is OOM issue and OS kills the process. So then tried to run just:
```
/Users/tcagent1/agent/work/7b79572c01edd19a/build/android-studio/Android Studio Preview.app/Contents/jbr/Contents/Home/bin/java -version
```
That should just print the java version. But I also got error code 137. From that, I believe memory is not an issue but might be an issue with permission. So maybe OS kills the process since the user doesn't have permission to start it.

PR #472, opened 2023-02-01
Add support for Android Studio Giraffe (2022.3)

asodja:
By using newer APIs. Good news is that all these APIs are also in other versions we support (Chipmunk and newer), so this also works with previous versions.

Fixes https://github.com/gradle/gradle-profiler/issues/469
Fixes https://github.com/gradle/gradle-profiler/issues/470

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -129,8 +129,8 @@ androidStudioTests {
     val autoDownloadAndRunInHeadless = providers.gradleProperty("autoDownloadAndRunInHeadless").orNull == "true"
     runAndroidStudioInHeadlessMode.set(autoDownloadAndRunInHeadless)
     autoDownloadAndroidStudio.set(autoDownloadAndRunInHeadless)
-    // Electric Eel (2022.1.1) Beta 1
-    testAndroidStudioVersion.set("2022.1.1.11")
+    // Giraffe (2022.3.1.2) Canary 2
+    testAndroidStudioVersion.set("2022.3.1.2")
     testAndroidSdkVersion.set("7.3.0")
     // For local development it's easier to setup Android SDK with Android Studio, since auto download needs ANDROID_SDK_ROOT to be set with
     // an accepted license in it. See https://developer.android.com/studio/intro/update.html#download-with-gradle.
--- a/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
@@ -52,8 +52,8 @@ val unpackAndroidStudio = tasks.register<Copy>("unpackAndroidStudio") {
         }
     }) {
         eachFile {
-            // Remove top folder when unzipping,
-            // that way we get rid of Android Studio.app folder that can cause issues on Mac
+            // Remove top folder when unzipping, that way we get rid of Android Studio.app folder that can cause issues on Mac
+            // where MacOS would kill the Android Studio process right after start, issue: https://github.com/gradle/gradle-profiler/issues/469
             relativePath = RelativePath(true, *relativePath.segments.drop(1).toTypedArray())
         }
     }
--- a/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
@@ -67,6 +67,7 @@ val androidStudioInstallation = objects.newInstance<AndroidStudioInstallation>()
 
 tasks.withType<Test>().configureEach {
     dependsOn(installAndroidSdk)
+    environment("ANDROID_SDK_ROOT", System.getenv("ANDROID_SDK_ROOT"))
     jvmArgumentProviders.add(
         AndroidStudioSystemProperties(
             androidStudioInstallation,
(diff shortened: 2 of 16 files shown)


build.gradle.kts
```diff
@@ -129,8 +129,8 @@ androidStudioTests {
     val autoDownloadAndRunInHeadless = providers.gradleProperty("autoDownloadAndRunInHeadless").orNull == "true"
     runAndroidStudioInHeadlessMode.set(autoDownloadAndRunInHeadless)
     autoDownloadAndroidStudio.set(autoDownloadAndRunInHeadless)
-    // Electric Eel (2022.1.1) Beta 1
-    testAndroidStudioVersion.set("2022.1.1.11")
+    // Giraffe (2022.3.1.2) Canary 2
+    testAndroidStudioVersion.set("2022.3.1.2")
```
lptr, line 133: ❓ Would it make sense to test multiple versions of Android Studio?

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
```diff
@@ -50,7 +50,13 @@ val unpackAndroidStudio = tasks.register<Copy>("unpackAndroidStudio") {
             singleFile.name.endsWith(".tar.gz") -> tarTree(singleFile)
             else -> zipTree(singleFile)
         }
-    })
+    }) {
+        eachFile {
+            // Remove top folder when unzipping,
+            // that way we get rid of Android Studio.app folder that can cause issues on Mac
```
lptr, line 56: :x: Could you please add a short description of the issues to expect here? A link to the original issue if appropriate would be good too.

src/test/groovy/org/gradle/profiler/AndroidStudioIntegrationTest.groovy
```diff
@@ -314,4 +325,14 @@ class AndroidStudioIntegrationTest extends AbstractProfilerIntegrationTest {
         ]
         new Main().run(*args)
     }
+
+    private static Optional<String> findAndroidSdkPath() {
+        if (System.getenv("ANDROID_SDK_ROOT") != null) {
+            return Optional.of(System.getenv("ANDROID_SDK_ROOT").replace("\\", "/"))
+        }
+        String userDirAndroidSdkPath = "${System.getProperty("user.home").replace("\\", "/")}/Library/Android/sdk"
```
lptr, line 333: 🤔 It would be good to have a named method to call that converts Windows to Unix paths. There might already be one on the classpath we can call instead of spelling out this replace multiple times.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
```diff
@@ -88,11 +89,13 @@ public static void waitOnPreviousGradleSyncFinish(Project project) {
...
-        if (statusBar != null) {
-            statusBar.getBackgroundProcesses().forEach(it -> waitOnProgressIndicator(it.getSecond()));
-        }
+        DumbService.getInstance(project).runReadActionInSmartMode(() -> {
+            IdeFrame frame = WindowManagerEx.getInstanceEx().findFrameFor(project);
+            StatusBarEx statusBar = frame == null ? null : (StatusBarEx) frame.getStatusBar();
+            if (statusBar != null) {
+                statusBar.getBackgroundProcesses().forEach(it -> waitOnProgressIndicator(it.getSecond()));
+            }
+        });
```
lptr, line 98: ❓ Are we losing compatibility with some older IntelliJ/Android Studio versions by switching to this API? If so, can we document how the compatibility changes?

build.gradle.kts
asodja: That would be good, but I think it needs more work, so I would not do it as part of this PR.

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java
asodja: Yes, we are compatible with older versions. We actually still compile our plugin with IntelliJ 2021.1.1 API (Android Studio Chipmunk), so with the first version we added profiler support.

build.gradle.kts
asodja: Opened https://github.com/gradle/gradle-profiler/issues/476.

src/test/groovy/org/gradle/profiler/AndroidStudioIntegrationTest.groovy
asodja: Indeed there is on the classpath. Found `org.apache.commons.io.FilenameUtils.separatorsToUnix`

buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
asodja: Added some more info

build.gradle.kts
lptr: Sounds good.

lptr (approved):
LGTM.

blindpirate:
🎉

PR #473, opened 2023-02-01
Run all tests on CI

asodja:
I noticed we run only tests in root project on CI. But we have some tests also in other projects.


.teamcity/settings/GradleProfilerTest.kt
```diff
@@ -18,7 +18,7 @@ open class GradleProfilerTest(os: Os, javaVersion: JavaVersion, arch: Arch = Arc
 
     steps {
         gradle {
-            tasks = "clean :test"
+            tasks = "clean test"
```
lptr, line 21: 💢 Gah!

src/main/java/org/gradle/profiler/CommandExec.java
```diff
@@ -248,10 +254,30 @@ public void interrupt() {
...
+            if (!JavaVersion.current().isJava9Compatible()) {
+                // ProcessHandle API is available only from JDK9
+                return;
+            }
+            try {
+                @SuppressWarnings("unchecked")
+                Stream<Object> stream = (Stream<Object>) Process.class.getMethod("descendants").invoke(process);
+                Method destroy = Class.forName("java.lang.ProcessHandle").getMethod("destroy");
+                for (Object child : stream.collect(Collectors.toList())) {
+                    destroy.invoke(child);
```
lptr, line 274: ❓ Should we check the result? Maybe print something if it wasn't successful.

src/main/java/org/gradle/profiler/CommandExec.java
```diff
@@ -248,10 +254,30 @@ public void interrupt() {
...
         }
 
+        private void destroyDescendants() {
+            if (!JavaVersion.current().isJava9Compatible()) {
+                // ProcessHandle API is available only from JDK9
+                return;
+            }
+            try {
+                @SuppressWarnings("unchecked")
+                Stream<Object> stream = (Stream<Object>) Process.class.getMethod("descendants").invoke(process);
```
lptr, line 271: ❓ Can we move this code to some helper class that is compiled against Java 9+, so we don't need to use reflection?

src/main/java/org/gradle/profiler/CommandExec.java
lptr: 🤔 Perhaps it would be also nice to log whatever we kill successfully?

src/main/java/org/gradle/profiler/CommandExec.java
```diff
@@ -248,10 +254,30 @@ public void interrupt() {
...
+                Method destroy = Class.forName("java.lang.ProcessHandle").getMethod("destroy");
+                for (Object child : stream.collect(Collectors.toList())) {
```
lptr, line 273: :x: I think you can call `forEach()` here instead, no?

subprojects/studio-plugin/src/testFixtures/groovy/org/gradle/profiler/studio/plugin/StudioPluginSpecification.groovy
```diff
@@ -19,20 +28,32 @@ class StudioPluginSpecification extends Specification {
...
+        gradleProperties << "org.gradle.jvmargs=-Xmx1024m"
+        setupWrapper(projectDir.toPath())
+    }
+
+    private def setupWrapper(Path projectDir) {
+        Files.createDirectories(projectDir.resolve("gradle/wrapper"))
+        Files.copy(Paths.get("../../gradlew"), projectDir.resolve("gradlew"))
+        Files.copy(Paths.get("../../gradlew.bat"), projectDir.resolve("gradlew.bat"))
+        Files.copy(Paths.get("../../gradle/wrapper/gradle-wrapper.jar"), projectDir.resolve("gradle/wrapper/gradle-wrapper.jar"))
+        Files.copy(Paths.get("../../gradle/wrapper/gradle-wrapper.properties"), projectDir.resolve("gradle/wrapper/gradle-wrapper.properties"))
```
lptr, line 57: 🧐 Use a loop here?

asodja: Done

src/main/java/org/gradle/profiler/CommandExec.java
asodja: The problem is, that invoke throws exception. But moved to a separate method and used forEach

src/main/java/org/gradle/profiler/CommandExec.java
asodja: I think moving this to a separate project and compiling it against Java9+ adds quite some complication. I would not do that for these 2 method calls.

src/main/java/org/gradle/profiler/CommandExec.java
asodja: Added some logging here.

src/main/java/org/gradle/profiler/CommandExec.java
lptr: Aren't we running the build with Java 11? If not then let's leave it as is.

lptr (approved):
LGTM.

src/main/java/org/gradle/profiler/CommandExec.java
asodja: Not all all platforms. On Mac and Linux we still run with Java8. We do that so we have at least some coverage with Java8, that we still support

PR #475, opened 2023-02-03
Simplify Android Studio start logic

asodja:
This simplifies Android Studio start logic a lot. It basically gets rid of manually bootstrapping Android Studio by running just running:
- on MacOS "Contents/MacOS/studio" or "Contents/MacOS/idea"
- on Windows "bin/studio.bat" or  "bin/idea.bat"
- on Linux "bin/studio.sh" or "bin/idea.sh"

With that, we get rid of manually parsing Classpaths and manually adding required JVM args (like --add-opens). Jvm Args that needs to be added at startup are provided via STUDIO_VM_OPTIONS and IDEA_VM_OPTIONS env variable.


asodja:
Lets see, how this behaves on CI...

src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
```diff
@@ -95,57 +50,37 @@ public LauncherConfigurationParser withStudioAgentParameters(int studioAgentPort
...
-        commandLine.add("-cp");
-        commandLine.add(Joiner.on(File.pathSeparator).join(classpath));
-        systemProperties.forEach((key, value) -> commandLine.add(String.format("-D%s=%s", key, value)));
-        // Support for IntelliJ with Java17 runtime
-        commandLine.addAll(JAVA17_ADD_OPENS);
+        return new LaunchConfiguration(startCommand, headlessCommand, studioInstallDir, additionalJvmArgs, studioSandbox);
+    }
+
+    private List<String> getAdditionalJvmArgs() {
+       List<String> jvmArgs = new ArrayList<>();
```
lptr, line 64: 🧐 Whitespace.

src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
```diff
@@ -95,57 +50,37 @@ public LauncherConfigurationParser withStudioAgentParameters(int studioAgentPort
...
-        commandLine.addAll(studioJvmArgs);
-        commandLine.add(mainClass);
-        if (SHOULD_RUN_HEADLESS) {
-            // Note: In headless mode ANDROID_HOME and ANDROID_SDK_ROOT have to be set otherwise
-            // Studio will fail since "missing Android SDK" modal will try to open
-            commandLine.add("headless-starter");
+            Path agentJar = GradleInstrumentation.unpackPlugin("studio-agent").toPath();
+            Path supportJar = GradleInstrumentation.unpackPlugin("instrumentation-support").toPath();
+            Path asmJar = GradleInstrumentation.unpackPlugin("asm").toPath();
+            Path protocolJar = GradleInstrumentation.unpackPlugin("client-protocol").toPath();
```
lptr, line 69: ❔ Why convert these from `File` to `Path` only to immediately convert them to `String`? IIUC, converting from `File` to `String` would produce the same results, no?

lptr (approved):
LGTM, I wonder why we haven't done startup like this from the beginning.

src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
asodja: Yeah, makes no sense. Will change that.

src/main/java/org/gradle/profiler/studio/launcher/LauncherConfigurationParser.java
asodja: Hawkeye, fixed

PR #481, opened 2023-04-21
Check ANDROID_HOME first, ANDROID_SDK_ROOT is deprecated

joshfriend:
Currently debugging an issue with Giraffe beta 1 where it seems to fail to find the android sdk directory, tries to show the missing SDK modal, and crashes because studio is running in headless mode. During this, I found out that the `ANDROID_SDK_ROOT` env var is actually deprecated and thought it would be a good first PR to get familiar with the project source. 

https://developer.android.com/tools/variables#envar

Note that I am not able to test these changes, sync/builds fail because I have an M1 mac and theres no adoptium jdk8 available.

```
Failed to query the value of task ':compileGroovy' property 'javaLauncher'.
Unable to download toolchain matching these requirements: {languageVersion=8, vendor=any, implementation=vendor-specific}
Unable to download toolchain. This might indicate that the combination (version, architecture, release/early access, ...) for the requested JDK is not available.
Could not read 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse' as it does not exist.
```

### DIFF
--- a/.teamcity/settings/extensions.kt
+++ b/.teamcity/settings/extensions.kt
@@ -29,6 +29,7 @@ fun ParametrizedWithType.androidHome(os: Os) {
         Os.linux, Os.macos -> "/opt/android/sdk"
         Os.windows -> """C:\Program Files\android\sdk"""
     }
+    param("env.ANDROID_HOME", androidHome)
     param("env.ANDROID_SDK_ROOT", androidHome)
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -132,8 +132,8 @@ androidStudioTests {
     // Giraffe (2022.3.1.2) Canary 2
     testAndroidStudioVersion.set("2022.3.1.2")
     testAndroidSdkVersion.set("7.3.0")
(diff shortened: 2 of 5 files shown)


asodja:
I guess we could use Jdk11 and compile to Java8 with release flag 🤔

asodja:
I believe we had similar issues in our intergration tests (and performance tests) with Giraffe. Solution for us was, that we added local.properties before test start, see: https://github.com/gradle/gradle-profiler/blob/66f0e36152f66c7f5c7cf61982ae0f991589b703/src/test/groovy/org/gradle/profiler/AndroidStudioIntegrationTest.groovy#L31 and 
https://github.com/gradle/gradle/blob/ca15dff9280e4d2a6aff6423516981e99d37798c/subprojects/performance/src/performanceTest/groovy/org/gradle/performance/regression/android/RealLifeAndroidStudioPerformanceTest.groovy#L46-L85

asodja (approved):
LGTM

joshfriend:
Writing out a `local.properties` file also works for me!

PR #491, opened 2023-05-09
Update profiler.android-studio-setup.gradle.kts

liutikas:
Fix buildSrc build warnings

> Task :buildSrc:compileKotlin
w: file:///ssd/ssd1/temp/gradle-profiler/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts:20:40 'toLowerCase(): String' is deprecated. Use lowercase() instead. w: file:///ssd/ssd1/temp/gradle-profiler/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts:21:50 'toLowerCase(): String' is deprecated. Use lowercase() instead.

### DIFF
--- a/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.android-studio-setup.gradle.kts
@@ -17,8 +17,8 @@ repositories {
     }
 }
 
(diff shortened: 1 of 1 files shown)


asodja:
Thanks for fixing that!

PR #494, opened 2023-05-17
Update RELEASING.md with the information about snapshot versions

asodja:


RELEASING.md
```diff
@@ -12,3 +14,15 @@ These are the steps to release Gradle profiler.
 * Update the [release draft](https://github.com/gradle/gradle-profiler/releases) prepared by [release-drafter](https://probot.github.io/apps/release-drafter/), assign the tag you created and publish the release
 * The [Publish to SDKman](https://builds.gradle.org/buildConfiguration/GradleProfiler_GradleProfilerPublishToSdkMan?branch=%3Cdefault%3E) should trigger automatically after the publishing job finishes. Though it may fail, since the artifacts are not yet on Maven Central. The build will retry the SDKMan publishing after half an hour. If it still doesn't work, you can trigger the build manually. 
 * All done!
+
+### Releasing Snapshot Versions
```
lptr, line 18: ❓ Should we talk about "alpha" versions instead?

RELEASING.md
```diff
@@ -12,3 +14,15 @@ These are the steps to release Gradle profiler.
...
+
+Releasing master snapshot versions is also supported.
+These versions are normally used for dogfooding in the [gradle/gradle](https://github.com/gradle/gradle) build.
+We use `-alpha` suffix for these versions, e.g. `0.20.0-alpha01`.
```
lptr, line 22: ❔ Can we use `-alpha-1` format instead? Would match the format we use elsewhere, like in Gradle's `8.0-milestone-1` etc.

asodja: Sure why not, we can start with that format with next version

RELEASING.md
asodja: What about pre-release versions so it matches the GitHub terminology?

RELEASING.md
```diff
@@ -12,3 +14,15 @@ These are the steps to release Gradle profiler.
...
+
+Process is the same as releasing the final versions, except that you should not publish the [release draft](https://github.com/gradle/gradle-profiler/releases).
```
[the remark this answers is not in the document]
asodja: Alpha versions are not published to the SDKMan, so they are not publicly useful unless you build them from the source or you use it as a dependency in your project but I doubt a lot of people use that.

So from that I would say that it would be extra work, that we might not need, but I am open to other suggestions.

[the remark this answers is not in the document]
asodja: One option we could start doing in the future is that we start publishing alpha versions to SDKMan and then we also publish it as a pre-release.

RELEASING.md
lptr: Let's do that.

RELEASING.md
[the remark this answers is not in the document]
lptr: OK.

lptr (approved):
LGTM.

PR #500, opened 2023-06-13
Revert reorganization back into a single project for most profiler fu…

big-guy:
…nctions


big-guy:
@asodja This is what I had in mind as a revert.  It puts all the files back without any of the detangling refactoring.

asodja (changes requested):
It seems one dependency got lost and now there are compile errors.

But other than that LGTM.

PR #507, opened 2023-07-18
Pass idea.properties to studio

6hundreds:


src/test/groovy/org/gradle/profiler/AndroidStudioIntegrationTest.groovy
```diff
@@ -330,4 +330,25 @@ class AndroidStudioIntegrationTest extends AbstractProfilerIntegrationTest {
...
+
+        when:
+        runBenchmark(scenarioFile, 1, 1)
+
+        then:
+        logFile.find("Full sync has completed in").size() == 2
+        logFile.find("and it SUCCEEDED").size() == 2
+        def ideaPropertiesFile = new File(sandboxDir, "config/idea.properties")
+        def vmOptions = ideaPropertiesFile.readLines()
+        vmOptions.contains("foo=true")
```
asodja, line 352: :nail_care:

```suggestion
        def ideaProperty = ideaPropertiesFile.readLines()
        ideaProperty.contains("foo=true")
```

src/main/java/org/gradle/profiler/studio/launcher/StudioLauncher.java
```diff
@@ -64,6 +71,24 @@ private void logLauncherConfiguration(List<String> commandLine) {
         System.out.printf("* Using command line: %s%n%n", String.join(" ", commandLine));
     }
 
+    private Map<String, String> writeIdeaProperties() {
+        if (!studioSandbox.getConfigDir().isPresent()) {
```
asodja, line 75: :x: I think we can just create this file also in the`studioSandbox.getJvmArgsDir()` so we can set `idea.properties` even if config dir is not present. `jvmArgsDir` is just some temp folder that we currently use only for `idea.vmoptions`.

:thinking: We could then maybe also merge `writeAdditionalJvmArgs` and `writeIdeaProperties` methods.

:nail_care: Maybe it would be worth also renaming `jvmArgsDir` to something else.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -95,6 +89,7 @@ class ScenarioLoader {
     private static final String TOOLING_API = "tooling-api";
     private static final String ANDROID_STUDIO_SYNC = "android-studio-sync";
     private static final String ANDROID_STUDIO_JVM_ARGS = "studio-jvm-args";
+    private static final String ANDROID_STUDIO_IDEA_PROPERTIES = "idea-properties";
```
asodja, line 92: :nail_care: Could we also update [an example](https://github.com/gradle/gradle-profiler/blob/5dd7dce5cbc9e838ab2054bf3b4f01780a020f91/README.md?plain=1#L280-L283) in README.md and add also a comment what it does.

It might help if we just add example with the configuration we will use (is it `gradle.tooling.models.parallel.fetch=true`?)

asodja (changes requested):
Looks good! I just added a comment about creating `idea.properties` in config dir. I think we could move that to the same folder as `studio.vmoptions`, so we can override these options always and not only when sandbox is present.

src/main/java/org/gradle/profiler/studio/launcher/StudioLauncher.java
```diff
@@ -64,6 +71,24 @@ private void logLauncherConfiguration(List<String> commandLine) {
...
+            System.out.println("* idea.properties can be set only in the sandbox");
+            return Collections.emptyMap();
+        }
+        try {
+            Path ideaPropertiesFile = studioSandbox.getConfigDir().get().resolve("idea.properties").toAbsolutePath();
+            Files.write(ideaPropertiesFile, ideaProperties);
+            return ImmutableMap.<String, String>builder()
+                .put("STUDIO_PROPERTIES", ideaProperties.toString())
+                .put("IDEA_PROPERTIES", ideaProperties.toString())
```
asodja, line 84: :x: I didn't notice before, but we should pass path to a file here, see this change:
```suggestion
                .put("STUDIO_PROPERTIES", ideaPropertiesFile.toString())
                .put("IDEA_PROPERTIES", ideaPropertiesFile.toString())
```

asodja (approved):
LGTM

Issue #484, opened 2023-04-24
Can't override javaHome for `GradleBuildConfigurationReader`


joshfriend:
`GradleBuildConfigurationReader.readConfiguration()` invokes a gradle build with the `help` task to generate the `GradleBuildConfiguration`. (For the project I am profiling, this takes several minutes of configuration time to perform. Is it possible to shorten this time it takes to resolve the gradle version and read the build configuration?)

https://github.com/gradle/gradle-profiler/blob/af476ee42039f44904ac8364e7bdad860e9647b7/src/main/java/org/gradle/profiler/DefaultGradleBuildConfigurationReader.java#L126-L138

I cannot override the value of javaHome for this gradle execution, so it defaults to `JAVA_HOME` which is set to jdk11 in the environment I run these tests in. We would like to enforce JDK 17 for this project by throwing an error when a developer invokes a build with jdk11 to avoid more confusing errors that may happen later in the build, but this means that the project is no longer compatible with profiler.

I'm working on updating the build environment where profiling is run, but it seems like this should be possible to override.

ov7a:
Sorry for the late reply.

The issue is in the backlog of the relevant team and is prioritized by them.

***

I don't have a quick answer for that; this issue needs to be investigated further.

PR #528, opened 2024-01-08
Run help on root project to speed up project probe

joshfriend:
Related to #484, running `help` takes a significant amount of time in our build because it is so large and it configures the whole project just to find a few basic pieces of info. Using `:help` on the root project makes this considerably faster.

### DIFF
--- a/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
+++ b/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
@@ -126,7 +126,7 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
         GradleBuildConfiguration version;
         try (ProjectConnection connection = connector.forProjectDirectory(projectDir).connect()) {
             BuildEnvironment buildEnvironment = connection.getModel(BuildEnvironment.class);
-            new ToolingApiGradleClient(connection).runTasks(ImmutableList.of("help"), ImmutableList.of("-I", initScript.getAbsolutePath()), ImmutableList.of());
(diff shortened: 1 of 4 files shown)


asodja:
I think that makes sense, there are some tests failing where we check that `help` was run, e.g.:
```
logFile.find("<tasks: [help]>").size() == 2
|       |                       |      |
|       []                      0      false
```

I replacing `<tasks: [help]>` with `<tasks: [:help]>` should fix tests.

joshfriend:
I'm having trouble running the tests locally with an ARM mac, possibly because this supports such old gradle versions. For now i just globally replaced the expected output but I doubt that is going to pass

asodja:
I think it should do a trick. Let's see. And I see StudioPluginIntegrationTest is flaky so ignore that..

We need to upgrade Gradle version, thanks for a reminder.

joshfriend:
Ok, I might have it this time 😆 
Updated the `help` scenarios in the integration tests to run `:help` as well (instead of `help`) and got some of them to pass locally (by updating the `supportedGradleVersions` array somewhat)

asodja:
I am also using arm, but it works for me. Maybe it's an issue for you, since we request toolchains for JDK8 and maybe you don't have arm version of JDK8?

asodja:
I adjusted `ProfilerIntegrationTest` since there was one more failure, this is now good to merge. Thanks for the contribution!

joshfriend:
> maybe you don't have arm version of JDK8?

Yeah that must be why, I checked and I don't have one installed, I'll get one for next time! Foojay resolver would be a nice addition to do that automatically.

Thanks for fixing the remaining issues, hope it wasn't too annoying for you!

PR #535, opened 2024-01-26
Upgrade tested Android Studio version to 2023.2.1.7

asodja:


asodja:
There is a Windows failure and I think it's related to https://youtrack.jetbrains.com/issue/IDEA-333971. Let's wait a bit and see if new AS Jellyfish version will fix it

PR #544, opened 2024-03-11
Improve initial sync handling in headless mode

asodja:
There were some problems with IntelliJ 2023.2 where we didn't detect initial import.


src/main/java/org/gradle/profiler/studio/launcher/StudioLauncherProvider.java
```diff
@@ -93,6 +93,7 @@ private List<String> getAdditionalJvmArgs() {
         if (SHOULD_RUN_HEADLESS) {
             systemProperties.put("java.awt.headless", "true");
         }
+        systemProperties.put("external.system.auto.import.disabled", "true");
```
6hundreds, line 96: Does this line is still required, if we have [this](https://github.com/gradle/gradle-profiler/pull/544/files#diff-1db42d89fc67334dae3160c490cb73dd8e3ccfec2b29c40f024de2b295638066R60)

6hundreds (approved):

PR #548, opened 2024-03-22
Prototype use of ide-starter with gradle-profiler

asodja:
There are two important parts:
1. Load ide-starter dependencies via a different classloader, so we can use gradle-profiler with Java 8 (unless doing syncs)
2. implement abstraction over ide-starter so we can use it via a nice API


subprojects/ide-provisioning-api/src/main/java/org/gradle/profiler/ide/CommandChain.java
```diff
@@ -0,0 +1,8 @@
+package org.gradle.profiler.ide;
+
+public interface CommandChain {
```
6hundreds, line 3: 🤔 This is looks like leaked to API implementation details. Not sure that it's reasonable for API client to manipulate `Command` terms for running an IDE. Client wants just to invoke `fun sync(project, ide)`, probably add some VM arguments, but that's it.

WDYT?

Issue #551, opened 2024-03-27
Fix doc info


helfese:
Fix til 'Profiling a build'
Lines 61, 66, 73, 83

asodja:
Closing, see my comment in https://github.com/gradle/gradle-profiler/pull/552#issuecomment-2023201822.

PR #552, opened 2024-03-27
Fix doc info

helfese:
Fix til 'Profiling a build'
Lines 61, 66, 73, 83

Issue #551


asodja:
Thank you for your contribution! Unfortunately I won't accept this PR.

We've noticed number of similar PRs from you also in the gradle/gradle project. While I appreciate grammar fixes I don't want to encourage such PRs and issues being constantly opened to the gradle-profiler project in the future, since they bring too little value for the effort. And our resources for the gradle-profiler project are very limited.

Thank you for your understanding.

PR #556, opened 2024-05-04
commit changes in README.md

VANSH3104:
This pull request adds documentation for the GRADLE_PROFILER_OPTS environment variable to the README.md file. The documentation explains how to use GRADLE_PROFILER_OPTS to customize JVM options for the Gradle Profiler tool, providing users with guidance on adjusting memory settings or other JVM parameters as needed. This addition aims to improve the usability and clarity of the Gradle project repository by helping users better understand how to configure the Gradle Profiler for their profiling requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,3 +404,39 @@ You can compare Gradle against Bazel, Buck, and Maven by specifying their equiva
             targets = ["//thing/res_debug"]
         }
     }
+## Configuring Profiling Options
+
+The Gradle Profiler allows you to configure various JVM options using the `GRADLE_PROFILER_OPTS` environment variable. This variable enables you to customize the behavior of the profiler to suit your specific requirements.
+
+### Usage
+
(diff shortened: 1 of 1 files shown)


ov7a:
Sorry for the late reply.

The relevant team for this area will confirm the contents.

***

Since GRADLE_PROFILER_OPTS comes from [the application plugin](https://github.com/gradle/gradle/blob/3c4f4344b82d86fa3e265264a41fc1ab120ea92b/platforms/jvm/plugins-application/src/main/java/org/gradle/jvm/application/tasks/CreateStartScripts.java#L172) it might be optimal to shrink the explanation and just point to the relevant Application plugin docs. However, it's not [documented well](https://github.com/gradle/gradle/issues/36597).

alllex:
I agree with @ov7a, it is better to refer to the application plugin and any potential options it provides.

I am also concerned that such a large section with options could be mistaken by users as a way to affect the builds being measured.

Inspired by your contribution, I created:
- https://github.com/gradle/gradle-profiler/pull/726

We'll close this PR, as it served its purpose.

@VANSH3104, feel free to follow the [suggestion](https://github.com/gradle/gradle/issues/36597) above and contribute `_OPTS` documentation to the Gradle itself. The material you have in this PR is on the right track, but it needs to be generalized.

PR #562, opened 2024-06-28
Fix issues with Android Studio Koala sync

akerimsenol:


akerimsenol:
@asodja please take a look.

CC @joshfriend

joshfriend:
(this is in reference to a bug I filed with android studio: https://issuetracker.google.com/issues/346618772)

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
```diff
@@ -89,8 +87,9 @@ private void handleSyncRequest(StudioRequest request, Project project, GradleSys
 
         // In some cases sync could happen before we trigger it,
         // for example when we open a project for the first time.
-        waitOnPreviousGradleSyncFinish(project);
-        waitOnBackgroundProcessesFinish(project);
+        if (isStartup) {
+            waitOnStartupSyncToFinish(project);
```
asodja, line 91: ❓ Will we wait for indexing that happens after the sync to finish here? 

We deliberately waited on all processes to finish with `waitOnBackgroundProcessesFinish` so also indexing was done before we triggered a next sync.

akerimsenol: Waiting for indexing shouldn't be necessary as:

1. Triggering a new sync will pause any indexing. (Though newer versions of IDEA do allow indexing while syncing. Though I think that's another discussion due to that behavior potentially also affecting the IDE time measurements of the profiler.)
2. Indexing is not considered part of the measured values.

This might not be the all around best approach, but I think the current approach is error-prone and warrants changing. We have progress indicators for things like Firebase Test Lab (or Run configurations) that has nothing to do with indexing and I've seen those "block" the profiler for no reason.

asodja: Gradle "work time" is measured via tooling api , so we actually get 3 numbers from these measurements:
- Full sync time
- Gradle work time
- IntelliJ work time: calculated as Full sync time - Gradle work time

So in that case indexing would count as  `IntelliJ work time`.

I think measuring full time (meaning additional idexing etc) that IDE needs to do project refresh is still useful, since that way we get better idea of real users experience. 

I think it would be still nice we could wait on indexing to finish, since this always a big part of sync time.

asodja: Is there any better way to wait for indexing and other work to finish right after the sync?

akerimsenol: I've asked this question on Jetbrains Slack, let's see if they have some guidance for us.

joshfriend: > I think measuring full time (meaning additional idexing etc) that IDE needs to do project refresh is still useful, since that way we get better idea of real users experience.

Yes that would be useful and I have been wondering how to measure this, however, gradle-profiler currently sometimes accidentally captures indexing which causes wild outliers in the data. Is this discussion about how to do that intentionally? If so, can that be a separate discussion so that this PR can be merged? I've tested this a few times with Koala and newer IDEs and it seems to work fine.

tinder-johnbuhanan2:
Just bumped into this :)

subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/client/GradleProfilerClient.java
asodja: @akerimsenol did you get any response from Jetbrains?

asodja:
I merged https://github.com/gradle/gradle-profiler/pull/579 that should make profiling work with Koala and Ladybug. It takes an idea from this PR to only use the wait for sync method on startup sync and not manual sync, other changes are a bit different.

With that I will close this PR.

akerimsenol:
Thanks for following this up Anže. This was inactive since I was on an extended work break.

PR #567, opened 2024-08-12
Clean up gradle-profiler build set up

liutikas:
- Reduce number of java toolchains used from 3 to 1 (just need JDK 17 now)
- Move to a shared place for setting java source and target version to 8


alllex:
After https://github.com/gradle/gradle-profiler/pull/686 has landed, it's enough to have a single JDK 17 toolchain to build Gradle Profiler 🎉 

I'll close this PR, since the desired outcome has already been achieved

PR #569, opened 2024-08-21
Query phase system property in background to avoid CC miss

mlopatkin:
This way the property doesn't become a configuration input with the sufficiently recent version of Gradle and build scan plugin.


asodja (approved):
LGTM

PR #570, opened 2024-08-30
Add delete-file and copy-file mutators

bdemers:
- **Add delete-file and copy-file mutators**
- **Add system property support for MavenScenarioInvoker**


src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -91,6 +93,8 @@ class ScenarioLoader {
     private static final String ANDROID_STUDIO_JVM_ARGS = "studio-jvm-args";
     private static final String ANDROID_STUDIO_IDEA_PROPERTIES = "idea-properties";
     private static final String JVM_ARGS = "jvm-args";
+    private static final String DELETE_FILE = "delete-file";
+    private static final String COPY_FILE = "copy-file";
```
lptr, line 97: Considering that these should be able to take directories, too, I think `delete` and `copy` would be better names, WDYT?

lptr: I've been rethinking this, and I think it's okay to include `-file`; it's good to disambiguate that this is about the file system.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -91,6 +94,8 @@ class ScenarioLoader {
     private static final String ANDROID_STUDIO_JVM_ARGS = "studio-jvm-args";
     private static final String ANDROID_STUDIO_IDEA_PROPERTIES = "idea-properties";
     private static final String JVM_ARGS = "jvm-args";
+    private static final String DELETE_FILE = "delete-file";
```
asodja, line 97: ❓ Do we plan to add a documentation in README.md in a separate PR? If not it would be good we add it now

src/main/java/org/gradle/profiler/mutations/CopyFileMutator.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final File source;
+    private final File target;
+
+    public CopyFileMutator(File source, File target) {
+        this.source = source;
+        this.target = target;
+    }
+
+    @SuppressWarnings("UnstableApiUsage")
```
asodja, line 22: ❔ Which API is unstable, maybe there is a stable replacement?

src/main/java/org/gradle/profiler/mutations/CopyFileMutator.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public void beforeScenario(ScenarioContext context) {
+        try {
+            System.out.println("Copying '" + source.getAbsolutePath() + "' to '" + target.getAbsolutePath() + "'");
+            if (!target.exists()) {
+                Files.createParentDirs(target);
+            }
+            Files.copy(source, target);
+        } catch (IOException e) {
+            throw new IllegalStateException("Failed to copy '" + source.getAbsolutePath() + "' to '" + target.getAbsolutePath() + "'", e);
```
asodja, line 32: ❔ Should we throw UncheckedIOException here, instead of IllegalStateException?

src/test/groovy/org/gradle/profiler/mutations/CopyFileMutatorTest.groovy
```diff
@@ -0,0 +1,29 @@
...
+    def "copies file from source to target"() {
+        def testDir = tmpDir.newFolder()
+
+        def expectedContents = "Copy file from source to target"
+        def source = new File(testDir, "source.txt")
+        source.text = expectedContents
+        def target = new File(testDir, "nested/target.txt")
+
+        def spec = mockConfigSpec("""{
+            copy-file {
```
asodja, line 14: ❔ Can we copy also multiple files or we can copy just one?

asodja (commented):
LGTM overall. It would be good we add some documentation for this

src/main/java/org/gradle/profiler/mutations/CopyFileMutator.java
lptr: It's Guava. I think upgrading Guava is the right solution here. Let me take a look.

src/main/java/org/gradle/profiler/mutations/CopyFileMutator.java
lptr: You're right.

src/test/groovy/org/gradle/profiler/mutations/CopyFileMutatorTest.groovy
lptr: We can copy one source (a file or a directory) to a target location.

src/main/java/org/gradle/profiler/ScenarioLoader.java
lptr: Good point.

lptr:
Build: https://builds.gradle.org/buildConfiguration/GradleProfiler_GradleProfilerTestTrigger/88291113

README.md
```diff
@@ -291,27 +291,29 @@ Values are optional and default to the values provided on the command-line or de
...
 - `clear-project-cache-before`: Deletes the contents of the `.gradle` and `buildSrc/.gradle` project cache directories before the scenario is executed (`SCENARIO`), before cleanup (`CLEANUP`) or before the build is executed (`BUILD`).
 - `clear-transform-cache-before`: Deletes the contents of the transform cache before the scenario is executed (`SCENARIO`), before cleanup (`CLEANUP`) or before the build is executed (`BUILD`).
-- `clear-jars-cache-before`: Deletes the contents of the instrumented jars cache before the scenario is executed (`SCENARIO`), before cleanup (`CLEANUP`) or before the build is executed (`BUILD`).
-- `clear-android-studio-cache-before`: Invalidates the Android Studio caches before the scenario is executed (`SCENARIO`) or before the build is executed (`BUILD`). Due to Android Studio client specifics before cleanup (`CLEANUP`) is not supported. Note: cleaning the Android Studio caches is run only when Android Studio sync (`android-studio-sync`) is used.
+- `copy-file`: Copies files from one location to another. Has to specify a `source` and a `target` path, both are resolved against the project directory.
```
asodja, line 315: ❓ Should we mention for `copy-file` and `delete-file` that these works also with directories and not just files?

Also should we mention that relative paths are resolved agains project directories, but absolute paths can also be specified?

asodja (approved):
LGTM overall

README.md
lptr: I've added some language to clarify.

PR #571, opened 2024-09-02
Improve Maven benchmarking and add profiling support

lptr:
Builds on:

- https://github.com/gradle/gradle-profiler/pull/570

### DIFF
--- a/src/main/java/org/gradle/profiler/Profiler.java
+++ b/src/main/java/org/gradle/profiler/Profiler.java
@@ -18,14 +18,24 @@ package org.gradle.profiler;
 import java.io.File;
 import java.util.function.Consumer;
 
-public class Profiler {
+public abstract class Profiler {
 
     public static final Profiler NONE = new Profiler() {
+        @Override
+        public boolean requiresGradle() {
+            return false;
+        }
+
         @Override
         public String toString() {
             return "none";
         }
     };
+
+    /**
+     * Whether this profiler supports only Gradle builds.
+     */
+    public abstract boolean requiresGradle();
 
     public void validate(ScenarioSettings settings, Consumer<String> reporter) {
     }
--- a/src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
+++ b/src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
@@ -2,22 +2,50 @@ package org.gradle.profiler.maven;
 
 import com.google.common.collect.ImmutableMap;
 import org.gradle.profiler.BuildToolCommandLineInvoker;
+import org.gradle.profiler.InstrumentingProfiler;
+import org.gradle.profiler.InstrumentingProfiler.SnapshotCapturingProfilerController;
 import org.gradle.profiler.InvocationSettings;
+import org.gradle.profiler.ScenarioSettings;
 import org.gradle.profiler.result.BuildInvocationResult;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Consumer;
 
 public class MavenScenarioInvoker extends BuildToolCommandLineInvoker<MavenScenarioDefinition, BuildInvocationResult> {
     @Override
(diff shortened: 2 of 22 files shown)


src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
```diff
@@ -1,25 +1,53 @@
...
         commandLine.addAll(scenario.getTargets());
         scenario.getSystemProperties().forEach((key, value) ->
             commandLine.add(String.format("-D%s=%s", key, value)));
 
-        doRun(scenario, settings, resultConsumer, commandLine);
+        Map<String, String> profileEnvironment;
+        SnapshotCapturingProfilerController controller;
+        // TODO This only works with Async profiler, since the only thing we call from the controller is stopSession()
+        //      Capture this in the type hierarchy somehow
+        if (settings.getProfiler() instanceof InstrumentingProfiler) {
```
asodja, line 31: 🤔 Should we throw exception if profiler is not supported?

Maybe a profiler would have something like:
`Profiler.supports(BuildTool)`

src/main/java/org/gradle/profiler/Profiler.java
```diff
@@ -18,15 +18,25 @@
...
         @Override
         public String toString() {
             return "none";
         }
     };
 
+    /**
+     * Whether this profiler supports only Gradle builds.
+     */
+    public abstract boolean requiresGradle();
```
asodja, line 38: 🤔 It's a bit unfortunate we have this `boolean requiresGradle()`, maybe we could have some more generic method.

src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
```diff
@@ -1,25 +1,53 @@
...
+            InstrumentingProfiler profiler = (InstrumentingProfiler) settings.getProfiler();
```
asodja, line 32: 🤔 Do we need this cast?

asodja: Oh I guess because of profiler.newSnapshottingController`, I wonder how we could make that nice 🤔

lptr: This is what I dislike about this, yes. I think we should re-model how profilers and build tools work together, but that's outside the scope of this PR.

src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
lptr: I tried that, but it got a bit ugly. I think the answer is to remodel the relationship between profilers and build tools on a bit deeper level, but it's a more impactful change than I think will fit in this PR.

src/main/java/org/gradle/profiler/Profiler.java
lptr: Yes, 100%. Again, let's discuss how we want to model this relationship.

lptr: Essentially, there are profilers that can run via Java agent, and profilers that can run as an external process, and profilers that support both.

And then there are build tools that support supplying an agent (Gradle, Maven), build tools that support a long running daemon (Gradle only), and build tools that support neither (Buck and Bazel might be these, at least as a start).

I think this can be modeled nicely, and will make the compatibility between tool and profiler be obvious.

src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
lptr: Let's continue on https://github.com/gradle/gradle-profiler/pull/576

src/main/java/org/gradle/profiler/Profiler.java
lptr: https://github.com/gradle/gradle-profiler/pull/576

src/main/java/org/gradle/profiler/maven/MavenScenarioInvoker.java
lptr: Let's continue on https://github.com/gradle/gradle-profiler/pull/576

PR #576, opened 2024-09-05
Improve Maven benchmarking and add profiling support, take 2

lptr:
This is reopening #571.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,9 +287,26 @@ Here is an example:
 
 Values are optional and default to the values provided on the command-line or defined in the build.
 
-### Profiling incremental builds
+### Benchmark options
 
-A scenario can define changes that should be applied to the source before each build. You can use this to benchmark or profile an incremental build. The following mutations are available:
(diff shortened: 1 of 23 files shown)


lptr:
On modeling profilers vs build tools:

There are profilers that can run via Java agent, and profilers that can run as an external process, and profilers that support both.

And then there are build tools that support supplying an agent (Gradle, Maven), build tools that support a long running daemon (Gradle only), and build tools that support neither (Buck and Bazel might be these, at least as a start).

I think this can be modeled nicely, and will make the compatibility between tool and profiler be obvious.

asodja (commented):
Overall LGTM.

I added some questions to clarify changes. The one that I am most interested in is the one in `MavenScenarioInvoker`, I wonder what happens with incompatible profilers with Maven

lptr:
FTR, we agreed to merge this as is, and revisit how profilers are configured / matched with build tools soon.

PR #578, opened 2024-09-23
Use Maven wrapper

lptr:
Use it when it is available and a tool path is not specifically configured.

### DIFF
--- a/src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
+++ b/src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
@@ -1,6 +1,7 @@
 package org.gradle.profiler;
 
 import javax.annotation.Nullable;
+import javax.annotation.OverridingMethodsMustInvokeSuper;
 import java.io.File;
 import java.io.PrintStream;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
```diff
@@ -33,9 +34,14 @@ protected void printDetail(PrintStream out) {
...
 
-    public String getExecutablePath() {
+    public String getExecutablePath(File projectDir) {
         String toolHomePath = getToolHome() == null ? System.getenv(getToolHomeEnvName()) : getToolHome().getAbsolutePath();
-        return toolHomePath == null ? getExecutableName() : toolHomePath + "/bin/" + getExecutableName();
+        return toolHomePath == null ? getExecutablePathWithoutToolHome(projectDir) : toolHomePath + "/bin/" + getExecutableName();
+    }
+
+    @OverridingMethodsMustInvokeSuper
+    protected String getExecutablePathWithoutToolHome(File projectDir) {
```
asodja, line 43: 🤔 What if we pass projectDir to `getExecutableName()` (and maybe we rename it?) so we have just one method?

src/main/java/org/gradle/profiler/maven/MavenScenarioDefinition.java
```diff
@@ -58,6 +58,19 @@ protected String getExecutableName() {
         }
     }
 
+    @Override
+    protected String getExecutablePathWithoutToolHome(File projectDir) {
+        String wrapperName = OperatingSystem.isWindows()
+            ? "mvnw.cmd"
```
asodja, line 64: ❓ Should this be `mvnw.bat`?

src/main/java/org/gradle/profiler/BuildToolCommandLineScenarioDefinition.java
lptr: We also use `getExecutableName()` when looking up the executable from the tool home.

src/main/java/org/gradle/profiler/maven/MavenScenarioDefinition.java
lptr: No, it's `mvnw.cmd`, similar to `mvn.cmd` a few lines above.

PR #579, opened 2024-10-25
Make gradle-profiler work with Android Studio Koala and Ladybug

asodja:
Continuous the work from https://github.com/gradle/gradle-profiler/pull/562


buildSrc/src/main/kotlin/tasks/ExtractAndroidStudioTask.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.gradle.api.internal.file.FileOperations
+import org.gradle.api.tasks.*
+import org.gradle.process.ExecOperations
+import org.gradle.work.DisableCachingByDefault
+import java.io.File
+import javax.inject.Inject
+
+
+@DisableCachingByDefault(because = "Not worth caching")
+abstract class ExtractAndroidStudioTask @Inject constructor(
```
asodja, line 17: 💭 This is the same implementation as we have in gradle/gradle https://github.com/gradle/gradle/blob/81c9b411450ccb9a5c52df2ef77459dad08fddb4/build-logic/integration-testing/src/main/kotlin/gradlebuild/integrationtests/ide/ExtractAndroidStudioTask.kt#L38

gradle.properties
```diff
@@ -1,4 +1,5 @@
 profiler.version=0.22.0-SNAPSHOT
+org.gradle.jvmargs=-Xmx3100m -XX:MaxMetaspaceSize=768m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
asodja, line 2: 💭 Added since otherwise I got errors on CI that GC is trashing

lptr: `3100m` seems oddly specific. How about `4g`?

src/test/groovy/org/gradle/profiler/AbstractProfilerIntegrationTest.groovy
```diff
@@ -25,8 +25,9 @@ abstract class AbstractProfilerIntegrationTest extends AbstractIntegrationTest {
         "3.3", "3.5",
         "4.0", "4.7",
         "5.2.1", "5.6.3",
-        "6.0.1", "6.1", "6.6.1",
-        "7.1.1", "7.5"
+        "6.0.1", "6.9.4",
+        "7.1.1", "7.6.4",
+        "8.10.2"
```
lptr, line 30: ❓ Does adding to this list increase the time it takes to test the profiler?

I wonder what the use of testing with Gradle 3.3 is. Profiler is for people conscious about performance; using old Gradle versions or running on old versions of Java should be unimportant for them.

For example, for Gradle I think supporting the latest two major versions should be far more than enough.

Also, if we want to add the same level of support for `8.x`, we should include `8.0.2`. Or we can drop this notion, and from each major we could include only the latest minor/patch release.

gradle.properties
asodja: We can go with 4g

src/test/groovy/org/gradle/profiler/AbstractProfilerIntegrationTest.groovy
asodja: I removed one and added one, so I think it should be similar. :)

> I wonder what the use of testing with Gradle 3.3 is. Profiler is for people conscious about performance; using old 
Gradle versions or running on old versions of Java should be unimportant for them.

Good point. I think testing older versions is only useful for version comparison tests.

I agree we could remove some versions. What if we include just latest version for every major. And for latest major we test first and last version? We can probably also get rid of 3.x, 4.x, 5.x. So maybe: 
```
6.9.4, 7.6.4, 8.0.2, 8.10.2
```
Wdyt? 

Or should we also keep maybe 5.x, 4.x?

asodja: Note: with https://github.com/gradle/gradle-profiler/pull/580 I figured that with Java 17 we will only support 7.3+, unless we add support to gradle-profiler to start Gradle under test with different JDK

asodja: Done, we now test just `6.9.4, 7.6.4, 8.0.2, 8.10.2`, let me know if we should also include some older major version

gradle.properties
asodja: Done

src/test/groovy/org/gradle/profiler/AbstractProfilerIntegrationTest.groovy
asodja: I kept also 5.6.4, since some tests check if some logic is executed if run with Gradle < 6.0. We can clean that up later, but I don't want to spent to much time cleaning it right now

lptr (approved):

kirillzh:
@asodja @lptr, any chance a new release can be published containing this fix?

asodja:
Sure, I'll find some time to do a release this week

PR #580, opened 2024-10-28
Use Java 17 for most of gradle-profiler tests

asodja:
Since we will upgrade Android Studio that supports just Java 17. Additionally in the future Gradle will support only Java 17. 

We will keep Linux running with Java 8 for now.

### DIFF
--- a/.teamcity/settings/extensions.kt
+++ b/.teamcity/settings/extensions.kt
@@ -17,7 +17,10 @@ fun BuildType.agentRequirement(os: Os, arch: Arch = Arch.AMD64) {
 fun toolchainConfiguration(os: Os) = listOf(
     "-Porg.gradle.java.installations.auto-detect=false",
     "-Porg.gradle.java.installations.auto-download=false",
-    """"-Porg.gradle.java.installations.paths=%${os.name}.java8.oracle.64bit%,%${os.name}.java11.openjdk.64bit%""""
+    """"-Porg.gradle.java.installations.paths=
+        |%${os.name}.java8.oracle.64bit%,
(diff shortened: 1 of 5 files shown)


.teamcity/settings/extensions.kt
```diff
@@ -17,7 +17,10 @@ fun BuildType.agentRequirement(os: Os, arch: Arch = Arch.AMD64) {
 fun toolchainConfiguration(os: Os) = listOf(
     "-Porg.gradle.java.installations.auto-detect=false",
     "-Porg.gradle.java.installations.auto-download=false",
-    """"-Porg.gradle.java.installations.paths=%${os.name}.java8.oracle.64bit%,%${os.name}.java11.openjdk.64bit%""""
+    """"-Porg.gradle.java.installations.paths=
+        |%${os.name}.java8.oracle.64bit%,
+        |%${os.name}.java11.openjdk.64bit%,
```
asodja, line 22: 💭 I left Java11 here, since without https://github.com/gradle/gradle-profiler/pull/579 merged we still use Java11 toolchains

cobexer (approved):
LGTM

💅 would be nice if we could reduce the number of places where we repeat Java version numbers

asodja:
> 💅 would be nice if we could reduce the number of places where we repeat Java version numbers
I removed some repetition by simplifying setting javaHome()

asodja:
It looks like this needs some more changes, since quite some number of tests break

alllex:
We can close this PR, since https://github.com/gradle/gradle-profiler/pull/686 has landed

PR #598, opened 2025-03-19
Add support specifying gradle distribution in the gradle-version argument

rmarma:
Currently the `gradle-version` argument supports:
- path to a local directory with gradle;
- gradle version that will be downloaded from the official repository.

Changes in this pull request allows you to specify a URI to gradle distribution. This can be useful when we compare different gradle versions on a machine that does not have access to external repositories.

Usage:
```shell
--gradle-version="https://<some_internal_repo>/<path>/gradle-<version_1>-bin.zip" --gradle-version="https://<some_internal_repo>/<path>/gradle-<version_2>-bin.zip"
```

### DIFF
--- a/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
+++ b/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
@@ -18,6 +18,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
+import java.net.URI;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -91,6 +92,9 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
             }
             if (versionString.matches("\\d+(\\.\\d+)+(-.+)?")) {
                 return probe(connector().useGradleVersion(versionString));
+            }
+            if (versionString.matches("^(https?|file)://\\S+")) {
+                return probe(connector().useDistribution(URI.create(versionString)));
             }
         } catch (IOException e) {
             throw new RuntimeException("Could not locate Gradle distribution for requested version '" + versionString + "'.", e);
--- a/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
+++ b/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
@@ -86,7 +86,7 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
         Logging.startOperation("Inspecting the build using Gradle version '" + versionString + "'");
         try {
             File dir = new File(versionString);
-            if (dir.isDirectory()) {
+            if (!versionString.isEmpty() && dir.isDirectory()) {
                 dir = dir.getCanonicalFile();
                 return probe(connector().useInstallation(dir));
             }
--- a/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
+++ b/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
@@ -18,6 +18,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
+import java.net.URI;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -92,6 +93,12 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
             if (versionString.matches("\\d+(\\.\\d+)+(-.+)?")) {
                 return probe(connector().useGradleVersion(versionString));
             }
+            if (versionString.matches("^(https?|file)://\\S+")) {
+                return probe(connector().useDistribution(URI.create(versionString)));
+            }
+            if (dir.isFile() && dir.getName().endsWith(".zip")) {
+                return probe(connector().useDistribution(dir.toURI()));
+            }
         } catch (IOException e) {
             throw new RuntimeException("Could not locate Gradle distribution for requested version '" + versionString + "'.", e);
         }
@@ -116,7 +123,7 @@ public class DefaultGradleBuildConfigurationReader implements GradleBuildConfigu
             Properties properties = new Properties();
             properties.load(inputStream);
             return new BuildDetails(
-                properties.getProperty("gradleHome").trim(),
+                properties.getProperty("gradleHome", "").trim(),
                 properties.getProperty("isBuildScanPluginApplied", "false").trim().equals("true"),
                 properties.getProperty("isEnterprisePluginApplied", "false").trim().equals("true"),
                 properties.getProperty("isDevelocityPluginApplied", "false").trim().equals("true")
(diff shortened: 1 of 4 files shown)


ljacomet:
Sorry for the late reply

rmarma:
https://github.com/gradle/gradle-profiler/issues/710

h0tk3y:

The relevant team for this area will confirm the design of the implementation choices.

src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
```diff
@@ -89,6 +90,12 @@ private GradleBuildConfiguration doResolveVersion(String versionString) {
             if (versionString.matches("\\d+(\\.\\d+)+(-.+)?")) {
                 return probe(connector().useGradleVersion(versionString));
             }
+            try {
+                URI distributionUri = URI.create(versionString);
+                return probe(connector().useDistribution(distributionUri));
+            } catch (IllegalArgumentException e) {
+                // do nothing because versionString is not gradle distribution uri
```
mlopatkin: :x: `URI.create()` is very permissive - it returns a valid (relative) URI for things like `"not-a-version",` `""`, or any string without illegal URI characters. Because of that, this branch silently swallows almost every previously-rejected input and passes it to useDistribution(...), which then fails with a confusing Tooling API stack trace far from the user-facing CLI.

I'd check that the scheme is one of http/https/file. With this change the `IllegalArgumentException("Unrecognized Gradle version ...")` on line 105 stays reachable for typos and garbage input, and the catch on line 99 becomes unnecessary (`URI.create` won't throw for the absolute-URI cases we care about; anything malformed naturally falls through to the existing error).

mlopatkin (changes requested):
Apologies for the very long delay in reviewing this — thanks for your patience, and thanks again for the contribution.

The feature works on the happy path (I verified that `--gradle-version <distribution-zip-url>` downloads, probes, and runs against a small sample project), but I have a couple of concerns before we can merge:

  1. UX regression on invalid inputs. URI.create() accepts almost any string, including `"not-a-version",` `""`, plain words, relative paths, etc. After this change, any input that isn't an existing directory and doesn't match the version regex gets routed into `useDistribution(...)`, which fails deep inside the Tooling API with `GradleConnectionException: Could not 
  install Gradle distribution from '<input>'` caused by `URI is not absolute`, plus a long stack trace. Previously the same input produced a single clean line: `IllegalArgumentException: Unrecognized Gradle version '<input>'` specified. The existing fall-through error on line 105 is now effectively unreachable. See inline comment for a suggested fix.
  2. Test fixture needs updating. `CommandLineIntegrationTest.help output pins the exact help text` and currently fails on this branch — the `--gradle-version` description string in the test needs to be updated to match the new wording.
  3. No tests for the new behavior. `DefaultGradleBuildConfigurationReader.doResolveVersion` has no direct unit coverage at all. A small Spock test with a mocked GradleConnector would let us assert that `8.10.2`, `/some/dir`, `https://…zip`, and `file:///…zip` each take the right branch, and that garbage strings still surface as `Unrecognized Gradle version` rather than as Tooling API errors. That test would have caught the first problem.

rmarma:
Thanks for the review. Almost everything fixed. I'll try to finish writing a small Spock test with a mocked GradleConnector.

src/test/groovy/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReaderTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+    DaemonControl daemonControl
+
+    def setup() {
+        projectDir = tmpDir.newFolder("test-project")
+        gradleUserHome = tmpDir.newFolder("gradle-home")
+        daemonControl = new DaemonControl(gradleUserHome)
+    }
+
+    def "configuration reader throw exception when gradle version is some garbage string"() {
+        given: "Given"
```
mlopatkin: 💅  Please remove this duplicate string descriptions or make them actually descriptive. Otherwise this is just noise.

src/test/groovy/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReaderTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+        when: "When"
+        reader.readConfiguration("some garbage string")
+
+        then: "Then"
+        IllegalArgumentException exception = thrown(IllegalArgumentException.class)
+        exception.message == "Unrecognized Gradle version 'some garbage string' specified."
+    }
+
+    def "configuration reader throw exception when gradle version is empty string"() {
```
mlopatkin: 💅 These two tests can be folded into a single [parameterized test](https://spockframework.org/spock/docs/2.4/data_driven_testing.html).

🤔 I'd also add at least one test with some unsupported protocol, e.g. an FTP url.

src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
mlopatkin: 🤔 If if a user specifies a path to a distribution zip (not as a `file:` URL, just a path), maybe we should also use it as a distribution - for consistency?

mlopatkin (commented):
Thank you for looking into this! My primary concern is now resolved, though it would be great to have some positive tests. I left a few other suggestions inline.

mlopatkin:
One more thing: your commits now fail the DCO check, you need to add "Signed-off-by" trail, see https://github.com/apps/dco.

rmarma:
> Thank you for looking into this! My primary concern is now resolved, though it would be great to have some positive tests. I left a few other suggestions inline.

I want to add a positive test, but mocking `GradleConnector` isn't enough for that. The [probe](https://github.com/gradle/gradle-profiler/blob/d7a4c1acfd83597805f0ad29375478da1e287266/src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java#L129-L150) method needs refactoring, or I'm doing something wrong. I'll look into it a bit more.

---

Upd. Added tests.

src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
rmarma: Done

mlopatkin (approved):
LGTM, thank you!

mlopatkin:
It looks like some tests are failing now: https://ge.gradle.org/scans?search.names=gitCommitId&search.rootProjectNames=gradle-profiler&search.timeZoneId=Europe%2FBerlin&search.values=4ed8694580d5bd57ae32d5c723b7b6ffa285515c

@rmarma could you please take a look?

rmarma:
> It looks like some tests are failing now: https://ge.gradle.org/scans?search.names=gitCommitId&search.rootProjectNames=gradle-profiler&search.timeZoneId=Europe%2FBerlin&search.values=4ed8694580d5bd57ae32d5c723b7b6ffa285515c
> 
> @rmarma could you please take a look?

The issue is that on **CI**, `new File("")` resolves to the current working directory of the process, which exists. As a result, `isDirectory()` returns `true`, causing the code to proceed and fail later with a different error.

Simply updating the test expectation to expect this behavior is not a viable fix: it would cause the test to fail locally, where the same path resolves differently.

The proposed fix is to explicitly ignore empty path values in the code before checking the `isDirectory()`.

PR #599, opened 2025-03-26
Don't use deprecated org.gradle.util.GFileUtils

### LINKED ISSUE #568: Remove usage of org.gradle.util.GFileUtils
[PidCollector.java:16 calls org.gradle.util.GFileUtils](https://github.com/gradle/gradle-profiler/blob/1125b1ed8223d4f8f5ce57b4ac48232d6ce65b53/subprojects/build-operations/src/main/java/org/gradle/trace/pid/PidCollector.java#L12) and that method has been deprecated and will be removed in Gradle 9.0

https://ge.androidx.dev/s/nzodz7gwiblhy/deprecations?expanded=WyIyNCIsIjQiXQ&expanded-stacktrace=WyIyNiJd

ov7a:
Fixes #568

### DIFF
--- a/subprojects/build-operations/src/main/java/org/gradle/trace/pid/AbstractPidCollectorBuildService.java
+++ b/subprojects/build-operations/src/main/java/org/gradle/trace/pid/AbstractPidCollectorBuildService.java
@@ -8,9 +8,12 @@ import org.gradle.api.services.BuildServiceParameters;
 import org.gradle.build.event.BuildEventsListenerRegistry;
 import org.gradle.tooling.events.FinishEvent;
 import org.gradle.tooling.events.OperationCompletionListener;
-import org.gradle.util.GFileUtils;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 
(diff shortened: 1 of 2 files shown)


subprojects/build-operations/src/main/java/org/gradle/trace/pid/AbstractPidCollectorBuildService.java
```diff
@@ -26,7 +28,11 @@ public interface Parameters extends BuildServiceParameters {
     }
 
     protected AbstractPidCollectorBuildService() {
-        GFileUtils.writeFile(getPid().toString(), getParameters().getPidFile().get().getAsFile());
+        try (FileWriter writer = new FileWriter(getParameters().getPidFile().get().getAsFile())) {
+            writer.write(getPid().toString());
```
asodja, line 32: 💅 Can we maybe simplify it and just use `Files.write`?:
```
        try {
            Files.write(getParameters().getPidFile().get().getAsFile().toPath(), getPid().toString().getBytes(StandardCharsets.UTF_8));
        } catch (IOException e) {
            throw new UncheckedIOException(e);
        }
 ```
 ?

ov7a: I don't have a strong opinion about it, but it feels more mouthful. I've updated it

asodja (approved):
LGTM

PR #613, opened 2025-07-22
Support Develocity plugin for profiling

wolfs:
Gradle profiler should apply the Develocity plugin by default now, and support both builds with the Develocity or Gradle Enterprise plugin applied for build scan profiling.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.10-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
```
asodja, line 3: 👍 Nice

settings.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
-    id("com.gradle.enterprise").version("3.16.2")
-    id("com.gradle.enterprise.gradle-enterprise-conventions-plugin").version("0.7.5")
+    id("com.gradle.develocity").version("4.1")
```
asodja, line 2: 💯

src/main/java/org/gradle/profiler/buildscan/BuildScanProfiler.java
```diff
@@ -23,7 +28,7 @@ public static String defaultBuildScanVersion(GradleVersion gradleVersion) {
         } else if (gradleVersion.compareTo(GRADLE_6) < 0) {
             return "2.4.2";
         } else {
-            return "3.5.2";
+            return "4.1";
```
asodja, line 38: 🤔  This method is kinda weird now or was already weird when we introduced Gradle enterprise plugin, since it returns either a build scan or a develocity plugin version

Can we maybe add a comment, that versions for  Gradle <6 is a Build scan plugin version and Gradle >=6 is Develocity plugin version. 
Ideally we would change a logic a bit, but I guess adding a comment for now is already an improvement.

asodja (approved):
LGTM

src/main/java/org/gradle/profiler/gradle/DefaultGradleBuildConfigurationReader.java
```diff
@@ -50,7 +52,8 @@ private void generateInitScript() throws IOException {
             writer.println(
                 "settingsEvaluated {\n" +
                     "   def detailsFile = new File(new URI('" + buildDetails.toURI() + "'))\n" +
-                    "   detailsFile.text = \"isEnterprisePluginApplied=${it.pluginManager.hasPlugin('com.gradle.enterprise')}\\n\"\n" +
+                    "   detailsFile.text = \"isEnterprisePluginApplied=${it.pluginManager.hasPlugin('com.gradle.enterprise')}\\n\" + \n" +
```
asodja, line 55: 🤔 Why is there `\\n` before closing `"`, here `${it.pluginManager.hasPlugin('com.gradle.enterprise')}\\n\"` is that a bug from before?

Isn't that then printed out as:
```
detailsFile.text = "isEnterprisePluginApplied=${it.pluginManager.hasPlugin('com.gradle.enterprise')}
"
```
?

instead of
```
detailsFile.text = "isEnterprisePluginApplied=${it.pluginManager.hasPlugin('com.gradle.enterprise')}"
```

asodja: Oh yeah, it should be ok. We print out:
```
detailsFile.text = "isEnterprisePluginApplied=${it.pluginManager.hasPlugin('com.gradle.enterprise')}
isDevelocityPluginApplied=${it.pluginManager.hasPlugin('com.gradle.develocity')}
"
```

Nevermind

wolfs: There is a `\\n` since we are writing a property file that should have one entry per line. So in the details file I want to have
```
isEnterprisePluginApplied=true
isDevelocityPluginApplied=true
```

So in the initscript I need:
```
settingsEvaluated { +
   def detailsFile = new File(new URI('<uri>'))
   detailsFile.text = "isEnterprisePluginApplied=${it.pluginManager.hasPlugin('com.gradle.enterprise')}\n" +
       "isDevelocityPluginApplied=${it.pluginManager.hasPlugin('com.gradle.develocity')}\n"
}
```

And that is why I need to double escape some `\n` values.

asodja (approved):
LGTM

PR #615, opened 2025-08-01
Add support async-profiler 3.+

### LINKED ISSUE #606: gradle-profiler is not compatible with async-profiler 3.0+
The "home" dir structure has changed, gradle-profiler looks for a `build/` dir + some `.so` file in it and these are no longer present.

Image

Image

bric3:
Adds async-profiler 3.+ support, to fix #606.

In order to better support new release, the async-profiler distribution is modeled as an object that can be better adapted should the async-profiler distribution change.

This is still draft as `AsyncProfilerIntegrationTest` is failing on some tests.

Also, there is some remaining work that might expand beyond this PR : 

* since version 2.x async-profiler added a better html viz that could be better to use than SVG.
* in 4.1, there is now a `--all` flag to include all events.
* in 4.0, `jfrconv` allows to produce heatmaps from jfr recordings : `jfrconv --cpu -o heatmap profiler.jfr heatmap-cpu.html`, I believe this should be the default, however (1) it requires the html output, (2) to run the `jfrconv` over jfr recoring, and (3) I noticed that gradle-profiler has it's own post-processing.
* I believe the wallclock events should be identified as their own kind, not under CPU.

<details><summary>OutOfMemoryError: Java heap space during <code>AsyncProfilerIntegrationTest</code></summary>

```
java.lang.OutOfMemoryError: Java heap space
org.gradle.profiler.Main$ScenarioFailedException: java.lang.OutOfMemoryError: Java heap space
	at org.gradle.profiler.Main.run(Main.java:126)
	at org.gradle.profiler.AsyncProfilerIntegrationTest.profiles multiple events using async-profiler with tooling API and warm daemon(AsyncProfilerIntegrationTest.groovy:32)
Caused by: java.lang.OutOfMemoryError: Java heap space
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$PoolReader.read(ValueReaders.java:95)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:518)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ArrayReader.read(ValueReaders.java:135)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:518)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager$TypeEntry.readConstant(TypeManager.java:291)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager.readConstants(TypeManager.java:414)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.readConstantPoolEvent(ChunkLoaderV1.java:111)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:78)
  ...
```

</details>

### DIFF
--- a/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
+++ b/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
@@ -108,12 +108,6 @@ public class AsyncProfilerDistribution {
         public final String stringVersion;
 
         public Version(String stringVersion) {
-            if (stringVersion.equals("unknown")) {
-                this.major = -1;
-                this.minor = -1;
-                this.stringVersion = stringVersion;
-                return;
-            }
             String[] parts = stringVersion.split("\\.");
             if (parts.length < 2) {
                 throw new IllegalArgumentException("Invalid version: " + stringVersion);


asodja:
Thanks! I'll try to look over the week

>I believe the wallclock events should be identified as their own kind, not under CPU.

This might be a bug

>since version 2.x async-profiler added a better html viz that could be better to use than SVG.

Do you have example of that? If it works better then svg then imo we should use it by default. But we can also do it in a separate PR later.

>in 4.0, jfrconv allows to produce heatmaps from jfr recordings : jfrconv --cpu -o heatmap profiler.jfr heatmap-cpu.html, I believe this should be the default, however (1) it requires the html output, (2) to run the jfrconv over jfr recoring, and (3) I noticed that gradle-profiler has it's own post-processing.

It would be interesting to see comparison. Maybe it's better then our post-processing. Do you have any example at hand?

asodja:
I see here is one example:
https://htmlpreview.github.io/?https://github.com/async-profiler/async-profiler/blob/master/.assets/html/flamegraph.html

It's definitely nicer and more useful. But gradle-profiler does also some post-processing as you said, I need to check that.

bric3:
@asodja So about the failing test cases, it's mostly OOMEs coming from `AsyncProfilerIntegrationTest`, e.g. `"profiles multiple events using async-profiler with tooling API and warm daemon"`

```
java.lang.OutOfMemoryError: Java heap space
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$PoolReader.read(ValueReaders.java:95)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:518)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ArrayReader.read(ValueReaders.java:135)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:518)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager$TypeEntry.readConstant(TypeManager.java:291)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager.readConstants(TypeManager.java:414)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.readConstantPoolEvent(ChunkLoaderV1.java:111)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:78)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:47)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.runWith(Thread.java:1596)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

If I set the test max memory `maxHeapSize = "2g"` in the test task, the test passes but it's really slow. And it seems there's a lot of memory pressure (allocations that leads to GC activity) from the code that fold stacks

image
image

That said it seems to work after setting a max heap. That said the test takes a significant time, and this seems to be related to `perl`, `flamegraph.pl` being very slow.

image
Screenshot 2025-08-06 at 11 59 55

I wonder if this part would be necessary if using the new async-profiler `jfrconf` that emit the new html files. Presumably this should work for any JFR file, so the "sanitization" / filtering could emit a temporary JFR file.

bric3:
Note push forced, with signed-off-by trailer in git commits.

bric3:
@asodja I had to revise the defaults for the allocations, as they were too low, i.e. there was a sample every 10 bytes. Which caused a LOT of `ObjectAllocationInNewTLAB` that needed a LOT of processing memory taken by the `JfrToStackConverter`.

Screenshot 2025-08-31 at 19 59 53

So I think it's saner to use the [same default](https://github.com/jvm-profiling-tools/async-profiler/blob/e052d51323388360fb4194c1ab8998be2223b0de/src/arguments.h#L13-L13) as AS.

```cpp
const long DEFAULT_ALLOC_INTERVAL = 524287;  // 512 KiB
```

A different work can be done to improve the situation with `JfrToStackConverter`, and maybe a different one to look for an opportunity to use the new html format of AS.

bric3:
@asodja It seems there are some checks that don't pass,
* PR labels, unfortunately I cannot edit the labels
* and another for "dependency review", unsure what this one is.

bric3:
@asodja @mlopatkin I was thinking : in this contribution I choose to support all async-profiler version, but maybe it makes more sense to only support a specific major version of async-profiler ?

asodja:
Hey @bric3,
Thanks for your effort. We didn't forget about this PR, it's just that it requires some more focus from us to get it merged.
I think this is very valuable PR. 
 
As I checked only 
```
org.gradle.profiler.AsyncProfilerIntegrationTest.profiles wall clock time using async-profiler with tooling API and warm daemon
org.gradle.profiler.AsyncProfilerIntegrationTest.profiles heap allocation using async-profiler with tooling API and warm daemon
```
are still failing. But we could check these failures and bring the PR to the finish line, we may just need some more time.
Thank you for your patience.

bric3:
@asodja Sorry I guess I completely forgot these tests since the build was green (but actually not run). I pushed the fixes.

This was due to wall clock profiling that was not properly handled, when generating the flamegraphs. I also fixed the cli / lib depending on the async-profiler version.

The question remains, which version or version range of async-profiler the gradle-profiler should support.

Async-profiler most recent is 4.1, but is readying for a 4.2 that adds new interesting features. Also, async-profiler 3.0 works with a JDK 21, but fails on JDK 24+ (haven't checked the other in-between versions).

bric3:
@asodja I see that windows build is failing. I don't have a windows machine and I can't access Teamcity. Async-profiler is not available on windows. And I'm not sure what change could trigger that failure.

cobexer:
@bric3 Thanks for taking a look at the build results!

The Windows failure is a known issue in the test infrastructure currently where a process is not terminated fast enough and thus keeps a file locked that the test attempts to delete. It's not new and  thus will not block reviews of this PR!

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
```diff
@@ -1,16 +1,25 @@
...
 
 public class AsyncProfilerAllEventsProfilerFactory extends AsyncProfilerFactory {
     @Override
     public Profiler createFromOptions(OptionSet parsedOptions) {
+        // TODO support all event from 4.1
         AsyncProfilerConfig config = super.createConfig(parsedOptions);
-        AsyncProfilerConfig overrides = new AsyncProfilerConfig(config.getProfilerHome(), ImmutableList.of("cpu", "alloc", "lock"), AsyncProfilerConfig.Counter.SAMPLES, config.getInterval(), config.getAllocSampleSize(), config.getLockThreshold(), config.getStackDepth(), config.isIncludeSystemThreads());
+        AsyncProfilerConfig overrides = new AsyncProfilerConfig(
+            config.getDistribution(),
+            ImmutableList.of("cpu", "alloc", "lock"),
```
asodja: ❓ Can we also add "wall" here or that doesn't work with 3?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDownload.java
```diff
@@ -28,7 +29,7 @@
  * Downloads a version of async profiler and installs into ~/.gradle-profiler-dist.
  */
 public class AsyncProfilerDownload {
-    private static final String ASYNC_PROFILER_VERSION = "2.7";
+    private static final String ASYNC_PROFILER_VERSION = "4.1";
```
asodja: 💅 Let's bump this to 4.2, I see it was already in the mean time

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java
```diff
@@ -53,32 +55,84 @@ public String getName() {
...
+        List<String> events = new ArrayList<>(profilerConfig.getEvents());
+        boolean ap3plus = profilerConfig.getDistribution().getVersion().compareTo(AP_3_0) >= 0;
+
+        // v3.0+: alloc/lock/wall always become auxiliary options
+        // v2.9: if multiple events, remove alloc/lock/wall and use as auxiliary options
+        boolean useAllocOption = (ap3plus || events.size() > 1) && events.remove(AsyncProfilerConfig.EVENT_ALLOC);
+        boolean useLockOption= (ap3plus || events.size() > 1) && events.remove(AsyncProfilerConfig.EVENT_LOCK);
+        boolean useWallOption= (ap3plus || events.size() > 1) && events.remove(AsyncProfilerConfig.EVENT_WALL);
+
+        // v2.9: sonly supports a single event in -e
```
asodja: ❓ Is there any benefit of keeping 2.9 support? I would just remove it, it seems it just complicates things and I don't see benefit

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java
```diff
@@ -53,32 +55,84 @@ public String getName() {
...
+        if (!ap3plus && events.size() > 1) {
+            events.subList(1, events.size()).clear();
+        }
+
+        // Determine the interval to use based on the event type
+        int intervalToUse = profilerConfig.getInterval();
+        // If the single remaining event is alloc/lock/wall, use its specific interval with -i
+        if (events.size() == 1) {
```
asodja: ❓ Do we still need this logic if we remove 2.9 support?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        String output = new CommandExec().runAndCollectOutput(
+            Arrays.asList(executable.getAbsolutePath(), "--version")
+        );
+        Matcher matcher = Pattern.compile("Async-profiler ([\\d.]+).+").matcher(output);
+        if (matcher.find()) {
+            version = new Version(matcher.group(1).trim());
+        } else {
+            version = Version.UNKNOWN;
```
asodja: ❌  Can we throw exception here, something like: async-profiler version x.y is not supported?

I think we can remove Version.UNKNOWN too then.

build.gradle.kts
```diff
@@ -35,7 +35,7 @@ dependencies {
         because("Avoid old version of commons-compress introduced by ant-compress")
     }
     implementation("commons-io:commons-io:2.16.1")
-    implementation("org.openjdk.jmc:flightrecorder:8.0.1")
+    implementation("org.openjdk.jmc:flightrecorder:8.3.1")
```
asodja, line 38: ❓ What happens if we bump it to 9.x? Is it breaking the build?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,189 @@
...
+     */
+    public static class Version implements Comparable<Version> {
+        public static final Version UNKNOWN = new Version("unknown") {
+            @Override
+            public int compareTo(Version other) {
+                return 0;
+            }
+        };
+        public static final Version AP_3_0 = new Version("4.0");
+        public static final Version AP_4_0 = new Version("4.0");
```
asodja: 💅 Where do we need `AP_4_0` and `AP_4_1`?

asodja (changes requested):
Added some comments. If you won't have time to look at them, I can fix them later this week

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,189 @@
...
+                throw new IllegalStateException("Unsupported operating system: " + OperatingSystem.getId());
+            }
+
+            library = new File(home, "lib/libasyncProfiler" + libExtension);
+        } else {
+            // Fallback to async-profiler 2.x
+            executable = new File(home, "profiler.sh");
+            library = new File(home, "build/libasyncProfiler.so");
+        }
+        System.out.println(asprofBinary.getAbsolutePath());
```
asodja: ❓ Do we need to print this path?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,189 @@
...
+        File asprofBinary = new File(home, "bin/asprof");
+        if (asprofBinary.isFile()) {
+            executable = asprofBinary;
+            String libExtension;
+            // if async profiler 3+
+            if (OperatingSystem.isLinuxX86()) {
+                libExtension = ".so";
+            } else if (OperatingSystem.isMacOS()) {
+                libExtension = ".dylib";
+            } else {
```
asodja, line 63: ❓ Hmmm, so we don't support Windows, was that also the previous implementation?

asodja: I see we didn't support Windows already before, nevermind then. Even async-profiler doesn't support Windows

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
bric3: I used those at some point, but ultimately used a different approach. I let those, to possibly use those if needed, like handling the `all` option, that said lets remove these constant for simplicity.

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
bric3: Good catch, I think it should land here, as well.

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java
bric3: I agree, 2.9 doesn't make much sense to keep around, I originally kept it for backward compat.

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
bric3: Good catch indeed it should go away !

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
bric3: Actually, this works on Linux, but it isn't supported on macOs at this time.

As of today, on macOs async-profiler uses its wall_clock engine for cpu or wall clock event but both are not supported at this time.

So, I only added the wall-clock event if the OS is not macOs (i.e. it's Linux).

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java
bric3: Yep it can be removed

build.gradle.kts
bric3: I think there's a min JDK bump for 9.x

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,156 @@
...
+                throw new IllegalStateException("Unsupported operating system: " + OperatingSystem.getId());
+            }
+
+            library = new File(home, "lib/libasyncProfiler" + libExtension);
+        } else {
+            // Understand async-profiler 2.x, so we can at least report the version used
+            executable = new File(home, "profiler.sh");
+            library = new File(home, "build/libasyncProfiler.so");
+        }
+        System.out.println(asprofBinary.getAbsolutePath()); // TODO remove
```
asodja: 💅 Let's remove this

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,156 @@
...
+        if (!executable.isFile() // || !Files.isExecutable(executable.toPath())
+            || !library.isFile() // || !Files.isExecutable(library.toPath())
+        ) {
+            throw new IllegalStateException("Invalid async-profiler distribution at: " + home);
+        }
+
+        try {
+            executable.setExecutable(true);
```
asodja: 💅 I would personally not set permissions on behalf of users, but rather check them. 
I would only set them if distribution that we unzip doesn't have permissions.
Do we need to do this?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -0,0 +1,156 @@
...
+     * prevents to move to <code>major.minor.patch</code> or else. At this point, async-profiler had versions
+     * <em>2.9</em>, <em>3.0</em>, <em>4.0</em>, <em>4.1</em>, so this code knows about major and minor only.
+     */
+    public static class Version implements Comparable<Version> {
+        public final int major;
+        public final int minor;
+        public final String stringVersion;
+
+        public Version(String stringVersion) {
+            if (stringVersion.equals("unknown")) {
```
asodja: 💅 Do we still need this?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
bric3: 🤦 Oh yeah

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
bric3: For got about this one.

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
bric3: Unfortunately this was needed, because the java extractor do not set the executable bit (while command line extractor do it). However, your comment made me realize that I overlooked the code in `AsyncProfilerDownload` that sets permissions.

asodja: I see, I would set it just there then. I don't want to fiddle with users permissions if they pass async-profiler differently. But we can warn users in case a flag is not set

bric3: I pushed a [fix](https://github.com/gradle/gradle-profiler/pull/615/commits/70576b50a36453964c3987429db019182978f694) in the original extract code.

Note about the limitations in zip file format https://commons.apache.org/proper/commons-compress/limitations.html

bric3:
On a related note `jfrconv` is an "executable" jar  – since 4.1, before there was a real executable plus the jar lib. Since a jar is a zip file that is read from the end, there's is a prepended script that gets executed before.

I wonder if the classes there could be used to post-process the jfr files. Alternatively the jar can be acquired from the release itself : https://github.com/async-profiler/async-profiler/releases#:~:text=3%20days%20ago-,jfr%2Dconverter.jar,-sha256%3Aef58045432b56f0064db721b707bbabeee93551ab1b0a0befbbff22574b16eeb

cobexer:
Hey @bric3 could you rebase on top of the current master branch and address the DCO failure as well?

build.gradle.kts
bric3: I confirm that org.openjdk.jmc:flightrecorder:9+ uses JDK 17 (classfile version 61)

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
bric3: More on that: Actually async-profiler engine has switched some time ago from _itimer_ which has a low accuracy to _wall-clock_ based one that works for both wall clock and CPU. 

However, as-of-today, there some trick to extract the data from the JFR samples (For example with `-e wall` or `--wall` there are no `ExecutionSamples` event, but the same data looks to be there under the `WallClockSample`, with a thread state equals to `RUNNABLE`).

asodja:
Thanks for your contribution!

PR #621, opened 2025-08-20
enhancements on ReflectionUtil#invokerGetter()

commandini:
- `getMethod()` will return public method, should use `getDeclaredMethod` instead.
- storing the original accessibility of target method after reflective invocation.


commandini:
@alextu @ov7a Can you review this PR, please?

alllex:
Hey, what is the limitation you are running into that requires this change?

commandini:
> Hey, what is the limitation you are running into that requires this change?

@alllex I haven't run into a limitation about that, but shouldn't the original accessibility eventually restored after the changes via Java reflection calls?

ov7a:
> shouldn't the original accessibility eventually restored after the changes via Java reflection calls?

Unless we have evidence that it matters in some context, we will not accept the change. At this point, we don't think the new version of the code is better.

PR #625, opened 2025-09-10
Refresh CI Setup

cobexer:


mlopatkin:
Not sure what is happening with the Android Studio download, but it was there before.

PR #627, opened 2025-10-01
Add clear-dir option

alllex:
This adds a new option that can be used in a scenario definition.

```
my-scenario {
  clear-dir = [{
    schedule = BUILD
    target = "gradle-user-home"
    keep = ["caches"]
  }, {
    schedule = BUILD
    target = "gradle-user-home/caches"
    keep = ["modules-2"]
  }]
}
```

The example above allows to remove everything in the GUH, except `caches/modules-2`. This is useful for cold Android Studio Sync scenarios, where the live IDE holds onto some files between resyncs, and doesn't react well when those files are backed up, deleted, and then restored again.

### DIFF
--- a/src/main/java/org/gradle/profiler/mutations/AbstractFileChangeMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/AbstractFileChangeMutator.java
@@ -9,10 +9,16 @@ import java.io.File;
 
 public abstract class AbstractFileChangeMutator implements BuildMutator {
     protected final File sourceFile;
+    private final String changeDescription;
     private String originalText;
 
-    protected AbstractFileChangeMutator(File sourceFile) {
+    protected AbstractFileChangeMutator(File sourceFile, String changeDescription) {
         this.sourceFile = sourceFile;
+        this.changeDescription = changeDescription;
+    }
+
+    protected AbstractFileChangeMutator(File sourceFile) {
+        this(sourceFile, "change");
     }
 
     @Override
@@ -22,6 +28,7 @@ public abstract class AbstractFileChangeMutator implements BuildMutator {
 
     @Override
     public void beforeBuild(BuildContext context) {
+        System.out.printf("> Applying %s to '%s'%n", changeDescription, sourceFile);
         StringBuilder modifiedText = new StringBuilder(originalText);
         applyChangeTo(context, modifiedText);
         FileSupport.writeUnchecked(sourceFile.toPath(), modifiedText.toString());
@@ -43,8 +50,7 @@ public abstract class AbstractFileChangeMutator implements BuildMutator {
     }
 
     @Override
-    public String toString()
-    {
+    public String toString() {
         return getClass().getSimpleName() + '(' + sourceFile + ')';
     }
 }
--- a/src/main/java/org/gradle/profiler/mutations/ApplyBuildScriptChangeFileMutator.java
+++ b/src/main/java/org/gradle/profiler/mutations/ApplyBuildScriptChangeFileMutator.java
@@ -6,7 +6,7 @@ import java.io.File;
 
 public class ApplyBuildScriptChangeFileMutator extends AbstractFileChangeMutator {
     public ApplyBuildScriptChangeFileMutator(File sourceFile) {
-        super(sourceFile);
+        super(sourceFile, "build script change");
     }
 
     @Override


asodja:
What is the difference compared to:
```
- `delete-file`: Deletes a file or a directory. Has to specify a `target` path; when relative it is resolved against the project directory. Can take an array of operations. Defaults to `SCENARIO` schedule.
```
?

alllex:
The main difference is that it preserves the original directory and allows preserving nested files as well.

I've added a description to the PR that clarifies the use case.

asodja:
Oh I see, this deletes just content, while the other deletes everything

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ClearDirectoryMutator extends AbstractFileSystemMutator {
```
asodja, line 17: ❓ Do we have any tests for this mutator?

alllex: Working on it!

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
```diff
@@ -0,0 +1,63 @@
...
+    public ClearDirectoryMutator(File target, Schedule schedule, List<String> keep) {
+        super(schedule);
+        this.target = target;
+        this.keep = keep;
+    }
+
+    @Override
+    protected void executeOnSchedule() {
+        System.out.printf("> Clearing directory: '%s'%n", target.getAbsolutePath());
+        if (!target.exists()) {
```
asodja, line 33: ❓ Should we log something if target doesn't exist?

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
```diff
@@ -0,0 +1,63 @@
...
+            return;
+        }
+
+        try {
+            try (Stream<Path> contents = Files.list(target.toPath())) {
```
asodja: ❓ What happens if target is a file, do we get an user-friendly error?

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    protected void executeOnSchedule() {
+        System.out.printf("> Clearing directory: '%s'%n", target.getAbsolutePath());
+        if (!target.exists()) {
+            return;
+        }
+
+        try {
```
asodja: 💅 Double `try`, can be collapsed I think

src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
```diff
@@ -24,11 +26,14 @@ protected void executeOnSchedule() {
                 gradleUserHome
             ));
         }
+
         try {
-            Files.list(gradleUserHome.toPath())
-                // Don't delete the wrapper dir, since this is where the Gradle distribution we are going to run is located
-                .filter(path -> !path.getFileName().toString().equals("wrapper"))
-                .forEach(path -> delete(path.toFile()));
+            try (Stream<Path> contents = Files.list(gradleUserHome.toPath())) {
```
asodja: 💅 Double try, we can probably have just one try

src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
```diff
@@ -24,11 +26,14 @@ protected void executeOnSchedule() {
...
+                contents
+                    // Don't delete the wrapper dir, since this is where the Gradle distribution we are going to run is located
+                    .filter(path -> !path.getFileName().toString().equals("wrapper"))
```
asodja: ❓ Should we add this to documentation?

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
alllex: I don't think so. This supports a case when some caching directories are not there for the first warm-up run, but are created on subsequent runs.

src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
alllex: Note, this is not a new behavior. It was never documented. I'll add it to the readme

alllex: Actually, this seems to be included already!
https://github.com/gradle/gradle-profiler/blob/11b4fc3e6a80b8001c7a1dfe3f9982420285f688/README.md?plain=1#L332

src/main/java/org/gradle/profiler/mutations/ClearGradleUserHomeMutator.java
```diff
@@ -3,35 +3,17 @@
...
-        this.gradleUserHome = gradleUserHome;
-    }
+public class ClearGradleUserHomeMutator extends ClearDirectoryMutator {
 
-    @Override
-    protected void executeOnSchedule() {
-        System.out.printf("> Cleaning Gradle user home: %s%n", gradleUserHome.getAbsolutePath());
-        if (!gradleUserHome.exists()) {
-            throw new IllegalArgumentException(String.format(
-                "Cannot delete Gradle user home directory (%s) since it does not exist",
```
alllex, line 23: 💭  Since the clear-dir mutator is lenient in the case of the missing directory, the GUH clearing mutator is now also lenient, while it previously wasn't. But I don't see a big reason, why it should blow up. The GUH can be created by the first warm-up run, if required. It may not exist before-hand, and it should be fine.

README.md
```diff
@@ -338,6 +338,7 @@ These mutators can be scheduled to execute at different points in the build benc
 
 #### File operations
 
+- `clear-dir`: Clears the contents of a directory without deleting the directory itself. Has to specify a `target` path; when relative it is resolved against the project directory. Optionally accepts a `keep` list to preserve specific files or directories within the target. Can take an array of operations. Defaults to `SCENARIO` schedule.
```
asodja: 💅 Could we may move this down just above `delete-dir` so user could see that there are two variants of file deletion? I think this could help in some cases

asodja: ❔  are `keep` files relative paths to target or how does that work?

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
```diff
@@ -0,0 +1,67 @@
...
+            return;
+        }
+
+        if (!target.isDirectory()) {
+            throw new IllegalArgumentException(String.format("Cannot clear '%s' since it is not a directory", target.getAbsolutePath()));
+        }
+
+        try (Stream<Path> contents = Files.list(target.toPath())) {
+            contents
+                .filter(path -> !keep.contains(path.getFileName().toString()))
```
asodja, line 43: ❔  Is my understanding correct we currently keep only top level files/directories and not files or directories on lower level. I think it's fine for now, just would like to understand it. Maybe we should mention that in documentation, so it's clear what is the current "keep" logic

README.md
alllex: Done

alllex: For now, it only accepts exact names of direct children of the target -- the simplest possible thing. I've updated the readme to clarify this.

src/main/java/org/gradle/profiler/mutations/ClearDirectoryMutator.java
alllex: See comment above

PR #628, opened 2025-10-01
build-ops-trace

alllex:


README.md
```diff
@@ -67,6 +67,8 @@ You can specify multiple versions and each of these is used to benchmark the bui
 
 You can also use the `--measure-config-time` option to measure some additional details about configuration time.
 
+You can use `--build-ops-trace` to produce a full Gradle build operations trace. This will generate detailed trace files in the output directory that capture all internal Gradle build operations.
```
asodja: 💅 Can we also add a sentence with a hint how user can inspect these traces?

asodja (approved):
LGTM

README.md
alllex: Done. I also added a note that this format is not stable, so people shouldn't rely on it in anyway.

PR #630, opened 2025-10-04
Scenario groups

alllex:
Introduces a new CLI option `--group <group name>` which allows to run a group of scenarios defined in the new `scenario-groups` block in the scenario files.

The grouping allows brining together relevant scenarios, so that it's easier to benchmark them together, since it's often interesting to _compare_ scenarios and not only measure absolute numbers. Groups provide a way for this to be represented explicitly in the scenario files.

A concrete case where this is useful is benchmarking a set of user scenarios with different performance features of Gradle, e.g. CC, Parallel CC, Parallel Execution, etc.

### Example

    scenario-groups {
        smoke-tests = ["assemble", "clean_build"]
        performance-suite = ["incremental_build", "full_build", "no_op"]
    }

    assemble {
        tasks = ["assemble"]
    }

    clean_build {
        cleanup-tasks = ["clean"]
        tasks = ["build"]
    }

    # ... definition of other scenarios ...

To run scenarios from a specific group:

    > gradle-profiler --benchmark --scenario-file performance.scenarios --group smoke-tests

### DIFF
--- a/src/main/java/org/gradle/profiler/CommandLineParser.java
+++ b/src/main/java/org/gradle/profiler/CommandLineParser.java
@@ -81,6 +81,10 @@ class CommandLineParser {
             "Title to show on benchmark report")
             .withOptionalArg()
             .ofType(String.class);
+        ArgumentAcceptingOptionSpec<String> groupOption = parser.accepts("group",
+            "Run scenarios from a group")
+            .withRequiredArg()
+            .ofType(String.class);
 
         OptionSet parsedOptions;
         try {
@@ -168,6 +172,7 @@ class CommandLineParser {
         }
         Format csvFormat = Format.parse(parsedOptions.valueOf(csvFormatOption));
         String benchmarkTitle = parsedOptions.valueOf(benchmarkTitleOption);
+        String scenarioGroup = parsedOptions.valueOf(groupOption);
 
         return new InvocationSettings.InvocationSettingsBuilder()
             .setProjectDir(projectDir)
@@ -193,6 +198,7 @@ class CommandLineParser {
             .setBuildOperationsTrace(buildOperationsTrace)
             .setCsvFormat(csvFormat)
             .setBenchmarkTitle(benchmarkTitle)
+            .setScenarioGroup(scenarioGroup)
             .build();
     }
 
--- a/src/main/java/org/gradle/profiler/ScenarioLoader.java
+++ b/src/main/java/org/gradle/profiler/ScenarioLoader.java
@@ -77,6 +77,7 @@ class ScenarioLoader {
     private static final String ITERATIONS = "iterations";
     private static final String BUILD_OPERATIONS_TRACE = "build-ops-trace";
     private static final String MEASURED_BUILD_OPERATIONS = "measured-build-ops";
+    private static final String SCENARIO_GROUPS = "scenario-groups";
     private static final String DEFAULT_SCENARIOS = "default-scenarios";
     private static final String APPLY_BUILD_SCRIPT_CHANGE_TO = "apply-build-script-change-to";
     private static final String APPLY_PROJECT_DEPENDENCY_CHANGE_TO = "apply-project-dependency-change-to";
@@ -169,7 +170,10 @@ class ScenarioLoader {
         .add(CLEAR_ANDROID_STUDIO_CACHE_BEFORE)
         .build();
 
-    private static final Set<String> RESERVED_TOP_LEVEL_KEYS = ImmutableSet.of(DEFAULT_SCENARIOS);
+    private static final Set<String> RESERVED_TOP_LEVEL_KEYS = ImmutableSet.of(
+        DEFAULT_SCENARIOS,
+        SCENARIO_GROUPS
+    );
 
     private static final List<String> BAZEL_KEYS = Arrays.asList(TARGETS, TOOL_HOME);
     private static final List<String> BUCK_KEYS = Arrays.asList(TARGETS, TYPE, TOOL_HOME);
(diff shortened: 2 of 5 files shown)


src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -193,6 +198,7 @@ public InvocationSettings parseSettings(String[] args) throws IOException, Setti
             .setBuildOperationsTrace(buildOperationsTrace)
             .setCsvFormat(csvFormat)
             .setBenchmarkTitle(benchmarkTitle)
+            .setScenarioGroup(scenarioGroup)
```
asodja, line 201: ❓ Why do we need scenarioGroup in `InvocationSettings`? Is it just to print it later or is there any othe usecase?

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -63,6 +77,8 @@ class ScenarioLoader {
     private static final String ITERATIONS = "iterations";
     private static final String BUILD_OPERATIONS_TRACE = "build-ops-trace";
     private static final String MEASURED_BUILD_OPERATIONS = "measured-build-ops";
+    private static final String SCENARIO_GROUPS = "scenario-groups";
+    private static final String DEFAULT_SCENARIOS = "default-scenarios";
```
asodja, line 81: ❓ What are `default-scenarios`?
Should we document that?

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -547,4 +556,70 @@ private static Class<?> loadType(Config config, String key, String description,
             throw new IllegalArgumentException("Unrecognized " + description + " '" + typeName + "' defined in scenario file " + scenarioFile);
         }
     }
+
+    private static Set<String> selectScenarios(Config config, InvocationSettings settings, File scenarioFile) {
+        SortedSet<String> availableScenarios = config.root().keySet().stream()
+            .filter(key -> !RESERVED_TOP_LEVEL_KEYS.contains(key))
+            .collect(Collectors.toCollection(TreeSet::new));
```
asodja, line 563: ❓ Is it ok if we have these scenarios sorted, or we want to keep the order from file? I would say it's more expected we keep the order that is in file

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -547,4 +556,70 @@ private static Class<?> loadType(Config config, String key, String description,
...
+        Config scenarioGroups = config.getConfig(SCENARIO_GROUPS);
+        Set<String> availableGroups = scenarioGroups.root().keySet();
+
+        if (!availableGroups.contains(groupName)) {
+            throw new IllegalArgumentException(String.format(
+                "Unknown scenario group '%s' requested. Available groups are: %s", groupName, String.join(", ", availableGroups)
+            ));
+        }
+
+        if (!settings.getTargets().isEmpty()) {
```
asodja: ❓ Should we do this validation `Cannot specify both --group and individual scenario names` before this if:
```
if (settings.getScenarioGroup() != null) {
...       
 } else if (!settings.getTargets().isEmpty()) {
 ...
}
```

It seems it doesn't belong in `selectScenariosFromGroup` method.

src/main/java/org/gradle/profiler/ScenarioLoader.java
asodja: Oh, I see they are documented... My bad

src/main/java/org/gradle/profiler/ScenarioLoader.java
asodja: Does that play any role at all?

src/main/java/org/gradle/profiler/ScenarioLoader.java
alllex: Yes, it was a magic constant in the method body before

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: Primarily because we want to have a `--group` cli option, and all cli-options seem to be stored in the invocation settings after being parsed. Regardless of that, this option should live as a sibling of the `targets` that we get, because they refer to the same information in different ways, and `targets` is part of the invocation settings.

It's also nice to be able to print it in the summary of the invocation, so it's immediately clear whether the right thing was invoked. I personally ran the wrong benchmarks multiple times, without immediately realizing it. If we don't add the group to the invocation settings, then we can't print it in the beginning.

src/main/java/org/gradle/profiler/ScenarioLoader.java
alllex: I mostly tried to preserve existing behavior.

Currently on master, we sort the scenarios in two cases.

1. When user supplied no targets and there are no default scenarios, we would run all scenarios from a file in alphabetical order:
https://github.com/gradle/gradle-profiler/blob/8cc83f9d61d253758956e9a2834532c0b5398afa/src/main/java/org/gradle/profiler/ScenarioLoader.java#L235

2. When a scenario is required that doesn't exist, we print the error and list available scenarios alphabetically:
https://github.com/gradle/gradle-profiler/blob/8cc83f9d61d253758956e9a2834532c0b5398afa/src/main/java/org/gradle/profiler/ScenarioLoader.java#L228

In this PR, we do exactly the same. If you follow the usages of the `availableScenarios` variable, it's only used as actual selected scenarios in the last `if`, so the (1) case from before, preserving behavior as it as on master. For all other cases, the `availableScenarios` are only used for error reporting, when a requested scenario does not exist.

The PR offers some improvement over the status quo. Namely, we now validate that all scenarios are available for the `default-scenarios` case. Previously, this would have failed at "runtime" when we would have tried to actually run the scenarios.

---

To summarize, if we stop sorting the scenarios now, then we'd introduce a behavior change for the case of the user requesting no scenarios (resulting in Profiler running all available scenarios). There would be minor behavior change for the list of scenarios in error messages, but I don't feel like they are as important.

My recommendation would be not to change the behavior in this PR. We can discuss the pros and cons of this change separately.

src/main/java/org/gradle/profiler/ScenarioLoader.java
alllex: This is a good idea 👍

alllex: Updated

PR #631, opened 2025-10-07
Log applied changes

alllex:
This adds logging for more mutation operations that happen as part of the benchmark.

Example: (the change is the `>` line)
```
* Running warm-up build #1
> Applying ABI change to '/path/to/src/main/kotlin/com/logic/utils.kt'
```

---

Some mutations log what they do, some don't. Unfortunately, we are currently inconsistent with that behavior.

I think we should generally go towards "all mutations log what they do", because most mutations can even be scheduled at different phases of the benchmark (scenario vs build), and even within those phases (pre-build cleaning vs build), and their order also matters.

### DIFF
--- a/src/test/groovy/org/gradle/profiler/mutations/ApplyAbiChangeToKotlinSourceFileMutatorTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/mutations/ApplyAbiChangeToKotlinSourceFileMutatorTest.groovy
@@ -12,7 +12,7 @@ class ApplyAbiChangeToKotlinSourceFileMutatorTest extends AbstractMutatorTest {
         mutator.beforeBuild(buildContext)
 
         then:
-        sourceFile.text == "class Thing { fun existingMethod() { }}fun _m_276d92f3_16ac_4064_9a18_5f1dfd67992f_testScenario_3c4925d7_MEASURE_7() {}"
+        sourceFile.text == "class Thing { fun existingMethod() { }}\n\nfun _m_276d92f3_16ac_4064_9a18_5f1dfd67992f_testScenario_3c4925d7_MEASURE_7() {}"
     }
 
     def "reverts changes when nothing has been applied"() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/gradle/profiler/mutations/ApplyAbiChangeToKotlinSourceFileMutator.java
```diff
@@ -7,13 +7,14 @@
...
-        super(sourceFile);
+        super(sourceFile, "ABI change");
     }
 
     @Override
     protected void applyChangeTo(BuildContext context, StringBuilder text) {
-        text.append("fun _m")
-                .append(context.getUniqueBuildId())
-                .append("() {}");
+        text.append("\n\n")
```
6hundreds, line 15: ❓ What is the purpose of `\n\n\` sprinkled around?

6hundreds:
LGTM in general. CI failure seems legit

src/main/java/org/gradle/profiler/mutations/ApplyAbiChangeToKotlinSourceFileMutator.java
alllex: It's to make the updated file not too weird. Without the newlines it appends to the last line, resulting in code like:
```
fun myActualFunction() {

}fun _m_uniqueIdentifier823457924() {}
```

alllex:
Thanks for catching the CI error. The Windows tests are flaky, so I didn't look at them, since the tests for other platforms passed. The CI failure points to a problem with a test, rather than an actual bug. I'll fix it in the follow up.

src/main/java/org/gradle/profiler/mutations/ApplyAbiChangeToKotlinSourceFileMutator.java
6hundreds: And that's was the case and before, right?

alllex: No everywhere, and not consistently, if I understood you correctly. As you can see here, the old code appended the function directly, instead of inserting newlines. With these changes, I tried to make sure we add some whitespace consistently, when applying a change.

6hundreds: I see

PR #633, opened 2025-10-07
Add `--dump-scenarios` option

alllex:
Introduces a new `--dump-scenarios` CLI option that outputs resolved scenario configurations without running them. This is useful
for:
- Verifying scenario configuration and HOCON merging behavior
- Debugging scenario definitions before executing benchmarks
- Understanding how default values and overrides are applied

Each scenario is formatted as a standalone HOCON block that can be copied into a new scenario file as-is.

Example usage:
```
gradle-profiler --benchmark --scenario-file test-scenarios.conf --dump-scenarios my-scenario
```

Example output:
```
# Scenario 1/1
my-scenario {
tasks=[
help
]
warm-ups=3
}
```

The option works with:
- Explicit scenario targets
- Default scenarios (via `default-scenarios`)
- Scenario groups (via `--group`)

### DIFF
--- a/src/main/java/org/gradle/profiler/CommandLineParser.java
+++ b/src/main/java/org/gradle/profiler/CommandLineParser.java
@@ -70,6 +70,7 @@ class CommandLineParser {
         OptionSpecBuilder measureConfigTimeOption = parser.accepts("measure-config-time", "Include a breakdown of configuration time in benchmark results");
         OptionSpecBuilder buildOpsTraceOption = parser.accepts("build-ops-trace", "Enable Gradle build operations trace");
         OptionSpecBuilder dryRunOption = parser.accepts("dry-run", "Verify configuration");
+        OptionSpecBuilder dumpScenariosOption = parser.accepts("dump-scenarios", "Dump resolved config for scenario(s) without running");
         OptionSpecBuilder bazelOption = parser.accepts("bazel", "Benchmark scenarios using Bazel");
         OptionSpecBuilder buckOption = parser.accepts("buck", "Benchmark scenarios using buck");
         OptionSpecBuilder mavenOption = parser.accepts("maven", "Benchmark scenarios using Maven");
@@ -113,7 +114,8 @@ class CommandLineParser {
         Profiler profiler = profilerFactory.createFromOptions(parsedOptions);
         boolean generateDiffs = !parsedOptions.has(noDifferentialFlamegraphOption) && hasProfiler && profiler.isCreatesStacksFiles();
         boolean benchmark = parsedOptions.has(benchmarkOption);
-        if (!benchmark && !hasProfiler) {
+        boolean dumpScenarios = parsedOptions.has(dumpScenariosOption);
+        if (!benchmark && !hasProfiler && !dumpScenarios) {
             return fail(parser, "Neither --profile or --benchmark specified.");
         }
 
@@ -182,6 +184,7 @@ class CommandLineParser {
             .setOutputDir(outputDir)
             .setInvoker(invoker)
             .setDryRun(dryRun)
+            .setDumpScenarios(dumpScenarios)
             .setScenarioFile(scenarioFile)
             .setVersions(gradleVersions)
             .setTargets(targetNames)
--- a/src/main/java/org/gradle/profiler/Main.java
+++ b/src/main/java/org/gradle/profiler/Main.java
@@ -51,6 +51,16 @@ public class Main {
                 return;
             }
 
+            if (settings.isDumpScenarios()) {
+                if (settings.getScenarioFile() == null) {
+                    System.err.println("--dump-scenarios requires a scenario file (--scenario-file)");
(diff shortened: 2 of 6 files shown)


src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -113,7 +114,8 @@ public InvocationSettings parseSettings(String[] args) throws IOException, Setti
         Profiler profiler = profilerFactory.createFromOptions(parsedOptions);
         boolean generateDiffs = !parsedOptions.has(noDifferentialFlamegraphOption) && hasProfiler && profiler.isCreatesStacksFiles();
         boolean benchmark = parsedOptions.has(benchmarkOption);
-        if (!benchmark && !hasProfiler) {
+        boolean dumpScenarios = parsedOptions.has(dumpScenariosOption);
+        if (!benchmark && !hasProfiler && !dumpScenarios) {
             return fail(parser, "Neither --profile or --benchmark specified.");
```
asodja, line 119: ❓  Is this fail message missing the new `dump-scenarios` option?

src/main/java/org/gradle/profiler/Main.java
```diff
@@ -51,6 +51,16 @@ public void run(String[] args) throws Exception {
                 return;
             }
 
+            if (settings.isDumpScenarios()) {
+                if (settings.getScenarioFile() == null) {
```
asodja: 💅 Can we move this logic to a method?

src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -70,6 +70,7 @@ public InvocationSettings parseSettings(String[] args) throws IOException, Setti
         OptionSpecBuilder measureConfigTimeOption = parser.accepts("measure-config-time", "Include a breakdown of configuration time in benchmark results");
         OptionSpecBuilder buildOpsTraceOption = parser.accepts("build-ops-trace", "Enable Gradle build operations trace");
         OptionSpecBuilder dryRunOption = parser.accepts("dry-run", "Verify configuration");
+        OptionSpecBuilder dumpScenariosOption = parser.accepts("dump-scenarios", "Dump resolved config for scenario(s) without running");
```
asodja, line 73: ❓ Are we missing a README entry for this option?

asodja (changes requested):

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: I purposefully left it out. One of these two options is mandatory for well-formed Gradle Profiler request. So, if the user forgets to provide one, it's important we fail and guide them. The guidance in 99% cases would be to enable either the profile or benchmark option. The dump-scenario option is a more rare cases needed for troubleshooting, so I think it's priority should be very low.

My suggestion would be to either leave it as it is now in the PR, or to preserve this condition as it is on master, such that we always require `--profile` or `--benchmark`, consitently, and then the `--dump-scenarios` is something users can provide only as an addition to those. A tiny downside is that we don't actually need neither `--profile` nor `--benchmark` to do the dump-scenario logic, but in practice this will probably never going to be an issue.

Wdyt?

alllex: After thinking about this more, I feel more included to not require `--profile` or `--benchmark` when having the `--dump-scenario` option. I imagine, the developer loop to troubleshoot a scenario will involve running the most simple command like:
```
gradle-profiler --scenario-file performance.scenarios --dump-scenarios clean_build
```

And the other two options are essentially unnecessary noise in this case.

src/main/java/org/gradle/profiler/Main.java
alllex: Done

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: For some reason, I was certain I already did that. but it seems I didn't 😅 

Now the documentation is there

README.md
```diff
@@ -432,6 +432,50 @@ They can be added to a scenario file like this:
         jvm-args = ["-Xmx2500m", "-XX:MaxMetaspaceSize=512m"]
     }
 
+### Troubleshooting scenario configuration
+
+Gradle Profiler supports defining scenarios in the `--scenario-file` using [Typesafe config](https://github.com/typesafehub/config) format.
```
asodja, line 437: 💅 Isn't format HOCON?

src/main/java/org/gradle/profiler/CommandLineParser.java
asodja: Ok, sounds reasonable if we leave it like that.And if somebody will find it confusing in the future we can think about it again.

README.md
alllex: I just coped however we described this earlier in the readme for consistency. Though, I noticed the same thing, and have a tiny follow:
https://github.com/gradle/gradle-profiler/pull/635

PR #636, opened 2025-10-08
Speed up test execution using parallel testing

cobexer:
* Provision up-to date Java 11 and 17 using our JDK provisioning plugin
* some refactoring to simplify adding more Java versions later


cobexer:
Separating the non Java 21 / 25 parts from #632 to get faster CI execution in your hands sooner

.teamcity/settings/extensions.kt
```diff
@@ -3,25 +3,27 @@ import jetbrains.buildServer.configs.kotlin.CheckoutMode
...
-fun toolchainConfiguration(os: Os) = listOf(
-    "-Porg.gradle.java.installations.auto-detect=false",
-    "-Porg.gradle.java.installations.auto-download=false",
-    """"-Porg.gradle.java.installations.paths=%${os.name}.java8.oracle.64bit%,%${os.name}.java11.openjdk.64bit%,%${os.name}.java17.openjdk.64bit%""""
-).joinToString(" ")
+fun toolchainConfiguration(os: Os, arch: Arch): String {
+    return listOf(
+        "-Porg.gradle.java.installations.auto-detect=false",
+        "-Porg.gradle.java.installations.auto-download=false",
+        """"-Porg.gradle.java.installations.paths=${listOf(
```
alllex: ❌  Let's use `-D` here, which should work just fine

Context: https://github.com/gradle/gradle/issues/34212

alllex: ❓  Is 4 (four) quotes intentional? What is it needed?

Either way, probably better to extract the `listOf()` call into a helper method.

cobexer: 4 quotes were intentional (see previous version of the code) but extracting makes the entire block better so I did that and also removed the usage of 4 quotes.

.teamcity/settings/extensions.kt
```diff
@@ -3,25 +3,29 @@ import jetbrains.buildServer.configs.kotlin.CheckoutMode
...
     }
 }
 
-fun toolchainConfiguration(os: Os) = listOf(
-    "-Porg.gradle.java.installations.auto-detect=false",
-    "-Porg.gradle.java.installations.auto-download=false",
-    """"-Porg.gradle.java.installations.paths=%${os.name}.java8.oracle.64bit%,%${os.name}.java11.openjdk.64bit%,%${os.name}.java17.openjdk.64bit%""""
-).joinToString(" ")
+fun javaInstallationsFor(os: Os, arch: Arch) = listOf(
+    if (os == Os.macos && arch == Arch.AARCH64) { JavaVersion.ZULU_JAVA_8.javaHome(os, arch) } else { JavaVersion.ORACLE_JAVA_8.javaHome(os, arch) },
```
alllex: 💅  Can be a bit shorter without the curly braces

alllex (approved):
LGTM!

.teamcity/settings/GradleProfilerTest.kt
```diff
@@ -30,6 +31,16 @@ open class GradleProfilerTest(os: Os, javaVersion: JavaVersion, arch: Arch = Arc
     }
 
     features {
+        parallelTests {
+            numberOfBatches = if (os == Os.macos && arch == Arch.AMD64) {
+                2 // only 2 Intel mac's left
```
asodja, line 36: I would just stop testing Intel macs as you suggested on Slack

cobexer: I'll move the macOS coverage to Java 11 and Apple Silicon, dropping testing on Intel Macs

.teamcity/settings/extensions.kt
cobexer: I had to specify them using `-D` and `-P` for all the tests to pick them up properly

.teamcity/settings/GradleProfilerTest.kt
cobexer: But I do that in a separate PR

PR #638, opened 2025-10-09
CI: Replace Intel macOS Java 8 tests with Apple Silicon macOS Java 11 tests

cobexer:
* Intel macOS is nearing it's EOL
* we have few Intel macOS machines left thus slowing us down here


asodja (approved):
LGTM

PR #639, opened 2025-10-13
Implement improved flamegraph viewer

jvandort:
The current flamegraph viewer used by gradle-profiler is not very good. 

Specifically, it is not dynamic in that it does not react to resizing browser windows. Therefore, it does not effectively utilize high-resolution monitors, making it annoying to analyze complex profiles. Improving the existing tool is difficult, as it is [written in perl](https://github.com/gradle/gradle-profiler/blob/eed6faea78ce9dc227b8e33f9313a289eb7709a8/src/main/resources/org/gradle/profiler/flamegraph/flamegraph.pl) and ain't nobody wanna write perl. Plus, the old implementation is non-apache licensed. This implementation is apache licensed and was written without referencing the perl implementation. 

[Demo](https://github.com/user-attachments/files/22896370/index.html)

The tool is still relatively rough but can certainly view flamegraphs. Currently, the tool can: 
- Load flamegraphs from a -stacks.txt file
- Navigate the flamegraphs, zooming into individual nodes
- Dynamically update colors of the graph
- Generates "merged" flamegraphs for the same method into a single aggregate flamegraph, improving resolution
- Switch between rendering different flamegraphs

The build logic for this tool completely manages the node/npm installation. 

Future work includes:
- UI cleanup
- Performance optimizations
- Diffing flamegraphs
- Integration into gradle-profiler outputs as a static file
- Analysis of cpu vs wall time
- Hide the color picker by default
- Add tests to validate -stacks files are parsed correctly and sub-graph merging works correctly

In order to make this actually usable with `gradle-profiler`, we need to implement a mechanism to embed stacks files in the flamegraph html bundle. Similar to how `gradle-profiler` currently executes the `flamegraph.pl` file to generate an `svg` with the flamegraph, we would need to have `gradle-profiler` generate emit a copy of the flamegraph application's `index.html`, with the `-stacks.txt` files embedded inside it. Then, the application would need to be updated to be able to load embedded flamegraphs in addition to its current mechanism of loading external flamegraphs.

Example screenshots:
image
image

### DIFF
--- a/subprojects/flamegraph/README.md
+++ b/subprojects/flamegraph/README.md
@@ -36,7 +36,7 @@ NPM commands: `./gradlew :flamegraph:npm --cmd "whatever you want"`
 
 For example, to install a new package:
 
-`./gradlew :flamegraph:npm --cmd "install some-package"`
+`./gradlew :flamegraph:npm --cmd "install --save fflate"`
 
 ### NPX
 
--- a/subprojects/flamegraph/package-lock.json
+++ b/subprojects/flamegraph/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.0.1",
       "license": "ISC",
       "dependencies": {
+        "fflate": "^0.8.2",
         "react": "^19.2.0",
         "react-dom": "^19.2.0"
       },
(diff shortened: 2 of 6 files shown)


bric3:
Did you looked at async-profiler's generated html?
This even allows to generate a heatmap and a flamegraph together.

jvandort:
This can't be merged as-is

PR #653, opened 2025-10-24
Add async-profiler-wall option for wall profiling

asodja:


src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
```diff
@@ -5,27 +5,29 @@
...
     public Profiler createFromOptions(OptionSet parsedOptions) {
         // TODO support all event from 4.1
         AsyncProfilerConfig config = super.createConfig(parsedOptions);
 
-        ImmutableList.Builder<String> allEvents = ImmutableList.<String>builder().add("cpu", "alloc", "lock");
-        // Combined cpu + wall is only supported on Linux at this time
-        if (!OperatingSystem.isMacOS()) {
-            allEvents.add("wall");
-        }
+        // Combined cpu + wall events are not supported on macOS at this time
```
jvandort: Do you have any resource we can link here describing why this is the case?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
```diff
@@ -29,8 +29,9 @@ public static List<String> allExtensions() {
     public static AsyncProfilerOutputType from(AsyncProfilerConfig config, ScenarioDefinition scenarioDefinition) {
         // TODO autoselect from scenario
         return (config.getEvents().size() > 1 || scenarioDefinition.createsMultipleProcesses())
+            // Only JFR supports multiple events
```
jvandort, line 32: Only JFR supports multiple events in a single file? 

Can we just spit out multiple `stacks.txt` files if we have multiple events?

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerWallProfilerFactory.java
```diff
@@ -0,0 +1,38 @@
...
+            config.getDistribution(),
+            allEvents,
+            AsyncProfilerConfig.Counter.SAMPLES,
+            config.getInterval(),
+            config.getAllocSampleSize(),
+            config.getLockThreshold(),
+            config.getWallInterval(),
+            config.getStackDepth(),
+            config.isIncludeSystemThreads(),
+            AsyncProfilerOutputType.JFR
```
jvandort, line 30: How come `AsyncProfilerAllEventsProfilerFactory` uses `getPreferredOutputType` instead of JFR, as it also has multiple event types? I feel like both `all` and `wall` should consistently be using the same parameter here.

jvandort (approved):
Mostly LGTM but I have some questions about the output format (JFR vs STACKS)

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
asodja: When you sample both asprofiler throws an error. 

There is also this issue:
https://github.com/async-profiler/async-profiler/issues/740

See `Notes`:
> Notes
>- The new feature has effect on Linux only. CPU profiling engine on macOS is already based on the wall clock mode - there is no additional gain comparing to the existing -e wall option.
>- Enabling wall profiling in addition to cpu results in higher overhead. It makes sense to increase wall interval when profiling both events together.

asodja: Also ta response from Gemini that makes it a bit clearer:
> While `async-profiler` technically supports profiling multiple events simultaneously (e.g., `cpu,wall,alloc`) when using the **JFR output format** (`-f out.jfr`), the combined use of `cpu` and `wall` on macOS offers no additional benefit.
>
> ***
>
> ## Reason for Limited Value on macOS
>
> The primary reason for the limited utility of combining `cpu` and `wall` on macOS is related to the underlying profiling mechanism:
> 
> 1.  **macOS CPU Profiling is Already Wall-Clock Based:** The **`cpu`** profiling engine on macOS already relies on a time-based sampling mechanism (similar to how wall-clock sampling works), as opposed to the hardware performance counters (`perf_events`) used on Linux. The `cpu` mode effectively filters this time-based data to show only threads that are actively running.
> 2.  **No Additional Gain:** Because the foundation is similar, the developers have noted that running `event=wall` in addition to `event=cpu` on macOS provides **no additional gain** compared to just using `event=wall` alone. The wall-clock profile on macOS already contains all the execution samples, and you can later filter them to approximate a CPU profile if needed.
>
>**In summary:** On macOS, you can usually stick with **`event=wall`** when you need to see both on-CPU and off-CPU time, or **`event=cpu`** when you only want to focus on execution time. Combining them offers little benefit.
>
>### Linux Contrast
>
>On Linux, the `cpu` event uses the highly precise hardware performance counters (`perf_events`), while `wall` uses a time-based interrupt. In this case, combining them (as in `event=cpu,wall`) and outputting to JFR gives you the benefit of both sampling methods, which is why joint profiling is more emphasized for Linux.

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
asodja: This is limitation of async-profiler. If you have multiple events you have to output them to jfr.

But we could get stacks (or other formats) out of `jfr` with `jfrconv` that is part of async-profiler distribution it seems via:
```
jfrconv --output collapsed --cpu profile.jfr > cpu.collapsed
jfrconv --output collapsed --alloc profile.jfr > alloc.collapsed
```
(other supported formats: html, collapsed, pprof, pb.gz, heatmap, otlp)

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerWallProfilerFactory.java
asodja: We can do that yes, to be more explicit. I left it that way, since it anyways output to jfr in the end, due to "multiple-events" logic later. But I can change that, so we are more explicit

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerOutputType.java
asodja: Note: I see we are actually already doing something similar, see this code:
https://github.com/gradle/gradle-profiler/blob/4f9f1e1f4515b91e841ee4fb74c6418e75bf716e/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerController.java#L125-L136

If output is JFR, we get out stacks and we output stack files and then we generate our graphs

asodja:
@jvandort I merged the PR so I can release this feature, if you have some more comments let me know and I can address it in a separate PR

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerAllEventsProfilerFactory.java
bric3: By the way as it happens, I had a discussion on this with Andrei on the same topic 

https://github.com/async-profiler/async-profiler/discussions/1557

On thing that stood out was this : 

> Current separation between `ExecutionSample` / `WallClockSample` is by profiling engine. When you want to distinguish events by thread state, there is a dedicated `state` field for this in both events.

So the `state` field is very important to distinguish the nature of the event! Especially if gradle-profiler has its own post processing.

PR #661, opened 2025-12-04
IDEA Sync Prototype

6hundreds:
This PR adds IDEA Sync support via IntelliJ IDE Starter.

**This increases the distribution size of the zip from 17mb to 283mb**

### Studio sync
1. Requires installed AS distribution on the machine, has no auto-provisioning mechanism of the desired version.
2. Triggers Sync via custom [studio-plugin](https://github.com/gradle/gradle-profiler/blob/109ecd4cb4623e518febc187567e6547abb1e7b2/subprojects/studio-plugin/src/main)
3. Uses [java-agent](https://github.com/gradle/gradle-profiler/blob/209ae15c14bce74ea92cc264175d4813e60d5d40/subprojects/studio-agent)(instrumenting of ToolingAPI) for Studio for dynamically set Gradle version and Gradle args.
4. Uses [java-agent](https://github.com/gradle/gradle-profiler/blob/209ae15c14bce74ea92cc264175d4813e60d5d40/subprojects/studio-agent)(instrumenting of ToolingAPI) for Studio for measuring duration of Gradle part of sync.
5. Manually (means `Instant.now() - start`) measures total sync duration.
6. Supports cleanup modes `before scenario/before build/never`
7. Supports headless mode

### IDEA sync
1. Supports auto-provisioning of desired IDEA version by pattern `"version-buildType" via [IDE Starter](https://github.com/JetBrains/intellij-community/blob/27b57a4c493abb7d22336ad7db20ea520308e209/tools/intellij.tools.ide.starter/README.md). Don't support providing a path to local IDEA installation
2. Triggers Sync via [IDE Driver API](https://github.com/JetBrains/intellij-community/blob/ded6a8aab5edb13dc17f3734aadc553673243765/tools/intellij.tools.ide.starter.driver/README.md)
3. Same as Studio, uses [java-agent](https://github.com/gradle/gradle-profiler/blob/209ae15c14bce74ea92cc264175d4813e60d5d40/subprojects/studio-agent)(instrumenting of ToolingAPI) for Studio for dynamically set Gradle version and Gradle args.
4. Same as Studio, uses [java-agent](https://github.com/gradle/gradle-profiler/blob/209ae15c14bce74ea92cc264175d4813e60d5d40/subprojects/studio-agent)(instrumenting of ToolingAPI) for Studio for measuring duration of Gradle part of sync.
5. Same as Studio, manually (means `Instant.now() - start`) measures total sync duration.
6. Supports cleanup modes `before scenario/before build`. `Never` is not supported as starter launches IDEA sync with clean caches.
7. Headless mode is not supported

## Open questions

1. There is some duality in configuration of profiler arguments - some arguments can be passed in scenario file, some via CLI, some in both ways(for instance, `warmup count`). In this PR I've organised it with the following logic - as scenario file may be included to VCS, it should contain only "stable" scenario-related arguments. I don't consider most of file paths "stable", because it's environment specific. That's why we pass it via CLI. Same about IDEA version - you may want to test same scenario with different IDEA versions. 

I'm opened to any adjustments to it, and I think we should standartize it over all profiler arguments. Shouldn't be done by this PR of course.

## Known problems
1. Headless mode isn't supported by `Driver`. There is an [issue](https://youtrack.jetbrains.com/issue/IJPL-221229/Open-project-from-argument-even-in-headless-mode) for fixing a some part of it(project opening part) on JB side. However, `Driver` relies heavily on UI-logic like `wait until progress indicators is gone`, so I doubt that complete support of headless mode is possible there. It worth to note, that our `Studio Plugin` relies on the same type of UI logic and it does work with headless.
2. `Starter` generates a lot of garbage files(well, it's a garbage for us), and this is can't be configured. A few instances: 
- `allure-results` dir generated with ton of reports, and I failed to find a way to at least reconfigure a report dir, to cleanup it
- `IDEA` writes a lot of indexing-diagnostics files altogether with other auxilary internal files, and this can't be configured. Although, we should able to cleanup it, as it's being written to a dir we control over.
3. IDEA writes a very useful FUS (Feature Usage Statistics) events file, which can be used for extracting durations of some processes. We can extract `total sync duration` and `gradle sync duration` out of it, and these values are more precise than values we're measuring manually. Furthermore, the file contains other useful data as duration of different sync phases, and much more.

    The nasty thing about it that the file being flushed to FS after some time, and not instantly as sync finish was reported. For instance, the very first sync should wait ~1minute for the worst case. Warm syncs(followed after the first one, without cleaning caches) are much faster, but still not predictable(1-20sec). With JB support we tried to fix it by execute special command `flushFusEvents` after the sync and adding [special internal properties](https://github.com/gradle/gradle-profiler/blob/dfaf916a923363498ea6885aa87128a8f4c54d12/subprojects/idea-sync/src/main/kotlin/org/gradle/profiler/idea/IdeaGradleClient.kt#L118-L119) that should force FUS events flushing each 10sec, but no luck. 
    
    However, there is an opportunity to install custom IDEA plugin, which will implement `StatisticsEventLogListener#onLogEvent` for intercepting log event. However, then it should communicate with the Profiler process somehow.
    
    Fixing this problem is very beneficial as it allows to completely get rid of usage instrumentation done by `studio-agent`, which is greatly simplifies infrastructure.
    
    Currently, we measure IDEA sync similarly as Studio implementation does.
4. There is a nasty race, when the agent failing by [waiting for incoming connection](https://github.com/gradle/gradle-profiler/blob/dfaf916a923363498ea6885aa87128a8f4c54d12/subprojects/idea-sync/src/main/kotlin/org/gradle/profiler/idea/IdeaGradleClient.kt#L123). It happens like 1 of 10 launches, I'm completely failed to understand what's happening there, Gemini said it's impossible and it's a bug in TCP/IP xD.


6hundreds:
Implemented by https://github.com/gradle/gradle-profiler/pull/740

PR #662, opened 2025-12-08
IDEA Sync Preparation I

6hundreds:
This PR is prerequisite for https://github.com/gradle/gradle-profiler/pull/661 and contains only extraction of the classes, required by the `IDEA Sync` subproject.


build.gradle.kts
```diff
@@ -43,6 +43,7 @@ dependencies {
         because("To write JSON output")
     }
     implementation(project(":client-protocol"))
+    implementation(project(":core"))
```
6hundreds: I remember the concerns about the name `core`, but I'm struggle to come with something better

alllex: I think so far we can confidently say this new subproject is about "build actions" as defined by the Gradle Profiler. The main class _is_ the `BuildAction`, which takes `GradleClient` as an argument, and then produces the `BuildActionResult`.

So I'd call this `build-action`

subprojects/build-action/build.gradle.kts
alllex, line 1: It would be good to add a short description to this subproject via `project.description` as we do in some other build scripts.

subprojects/build-action/build.gradle.kts
```diff
@@ -0,0 +1,3 @@
+plugins {
+    id("profiler.embedded-library")
```
asodja: ❌ If I remember correctly, `embedded-library` are jars that are embedded to distributions via `profilerPlugins`. I think this one is not embedded library

asodja: ❌ I believe we also need to publish this to maven central, since gradle-profiler uses is as a dependency. I think we are missing `profiler.publication` plugin

6hundreds: @asodja  I assume we need both? `embedded` and `publication`?

asodja: `profiler.embedded-library` is used just for libraries that are embedded into Gradle profiler (used as `profilerPlugins(projectName)`)

Instead of `profiler.embedded-library` I think you want to use `profiler.java-library.gradle`

PR #665, opened 2025-12-09
Update CodeQL GitHub action

alllex:
The CodeQL action analyzes the code and reports potential problems. However, out of the box it tries to detect the build setup and a correct build invocation based on heuristics, which are unreliable in the face of involved Gradle build logic.

This PR provides CodeQL with explicit build steps required to build the compiled part of the codebase, ensuring the analysis step can take advantage of the additional information.

The documentation for the CodeQL actions is here:
https://github.com/github/codeql-action/tree/main?tab=readme-ov-file

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,22 +1,11 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
   push:
-    branches: [ master, main ]
+    branches: [ master ]
   pull_request:
     # The branches below must be a subset of the branches above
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -32,39 +21,37 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'java', 'javascript' ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
-        # Learn more about CodeQL language support at https://git.io/codeql-language-support
+        language: [ 'java', 'kotlin', 'javascript' ]
```
asodja, line 24: ❓ Do we even need `javascript`?

asodja (commented):
LGTM

asodja (approved):
LGTM

.github/workflows/codeql-analysis.yml
alllex: We have some javascript, so I guess it is checked as well:
https://github.com/gradle/gradle-profiler/blob/fbe1f9b3409f9e38de7456b05294e304c19429cc/src/main/js/org/gradle/profiler/report/report.js

Overall, we have `node` involved, so we might extend the javascript usage at some point.

cobexer:
LGTM!

PR #668, opened 2025-12-09
Introduce `scenario-definition` subproject

6hundreds:
Extract `config` - parsing logic, as it's going to be used across different profiling facilitities


alllex (approved):
One file in the project looks kinda lonely, but we can add more files later.
Let's not block merging on this

PR #674, opened 2025-12-10
Drop testing of Gradle 5 and earlier

alllex:
Testing Gradle 5 and older versions of Gradle creates additional complications when running with modern JVMs and testing against modern Develocity plugin versions.

Since it's unlikely that users are interested in testing detailed performance of Gradle 5, we remove it from the test coverage. This does not necessarily mean that new versions of Gradle Profiler stop supporting Gradle 5 builds, so they should continue working in practice.

**We retain Gradle 6.x line as supported and tested.**

Finally, users always can use previous versions of Gradle Profiler to benchmarks old Gradle versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,19 @@ org.gradle.java.home=<path to a Java installation>
 
 It will be respected when starting the daemon for your build, regardless of the JVM Gradle Profiler uses.
 
+## Gradle version compatibility
+
+Gradle Profiler supports benchmarking and profiling builds with Gradle 6.0 or later.
+
+Gradle Profiler is tested against the latest patch releases of Gradle 6.x through 8.x.
+
+If you require running your build with a specific Java version for the Gradle Daemon JVM,
+which is different from the JVM used to run Gradle Profiler,
+see the [JVM requirements and option](#jvm-requirements-and-options) section for details.
+
+If you require running Gradle Profiler for builds that use Gradle 5.x, consider using previous versions of Gradle Profiler.
+The last version that officially supports Gradle 5.x is [Gradle Profiler 0.23.0](https://github.com/gradle/gradle-profiler/releases/tag/v0.23.0).
+
 ## Advanced profiling scenarios
 
 A scenario file can be provided to define more complex scenarios to benchmark or profile. Use the `--scenario-file` option to provide this. The scenario file is defined in [HOCON] format.


src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
```diff
@@ -1551,6 +1340,9 @@ class ProfilerIntegrationTest extends AbstractProfilerIntegrationTest {
 
         then:
         output.count("> Build cache size:") == 4
+
+        where:
+        gradleVersion << [minimalSupportedGradleVersion]
```
asodja, line 1345: 💅 Could be just inlined

src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationIntegrationTest.groovy
```diff
@@ -312,7 +306,6 @@ class BuildOperationInstrumentationIntegrationTest extends AbstractProfilerInteg
         latestSupportedGradleVersion | true
     }
 
-    @Unroll
```
asodja, line 315: ❓ Why is this not needed anymore?

src/test/groovy/org/gradle/profiler/ProfilerIntegrationTest.groovy
alllex: I am actually leaning towards doing the opposite of inlining in many other tests.

It reads much better, when all tests bodies are the same and use `gradleVersion`, which the `where:` clause explicitly determines which version we want to test with. It's also a PITA right now, that tests are not written that way, and use minimal vs latest version almost arbitrarily, while repeating it in many places.

src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationIntegrationTest.groovy
cobexer: AFAIR a simple Unroll is not needed as of Spock 2.1, see https://spockframework.org/spock/docs/2.1/migration_guide.html#_unroll_changes

alllex: The tests are unrolled even with the annotation, so I removed it for being unnecessary. See:
https://ge.gradle.org/s/5oyovscpe3fgo/tests/task/:test/details/org.gradle.profiler.BuildOperationInstrumentationIntegrationTest?top-execution=1

build.gradle.kts
```diff
@@ -127,6 +134,10 @@ tasks.test {
     // We now use forkEvery = 1 to run each test class in its own JVM, so we don't run into this problem anymore.
     setForkEvery(1)
     maxHeapSize = "2g"
+
+    // Used by tests to select a different Daemon JVM when Test JVM is unsupported by older Gradle versions
+    systemProperty("javaHomes.java8", launcherJavaHomeFor(8))
```
cobexer, line 139: 💅 If we intend to keep Java 8 longer term, we should add it to https://github.com/gradle/gradle-profiler/blob/master/.teamcity/jdks.yaml to make sure we use a known version of it.

You could pick Java 8 versions (and hashes) from https://github.com/gradle/native-platform/blob/master/.teamcity/jdks.yaml to save time

alllex: >You could pick Java 8 versions (and hashes) from [gradle/native-platform@master/.teamcity/jdks.yaml](https://github.com/gradle/native-platform/blob/master/.teamcity/jdks.yaml?rgh-link-date=2025-12-10T14%3A00%3A48Z) to save time

Do you mean reading that files at build time and pass values from it into the launcher spec, or something different?

cobexer: Our https://github.com/gradle/teamcity-jdk-provider-plugin/ plugin on TeamCity uses that file to dynamically provision JDKs on build agents which gives repositories the ability to have any Java version they may need available to them and frees us from the complexity of synchronizing Java version updates across repositories. 
`launcherJavaHomeFor(8)` will then still work as is.

README.md
```diff
@@ -229,6 +229,27 @@ The following command line options only apply when measuring Gradle builds:
 - `--no-studio-sandbox`: Do not use the Android Studio sandbox but use the default Android Studio folders for the Android Studio data.
 - `--no-diffs`: Do not generate differential flame graphs.
 
+## JVM requirements and options
```
asodja, line 232: ❓ Could we also mention that Gradle 6+ is tested and earlier versions are not anymore? And possibly thatuser should use some old gradle-profiler to test older versions

alllex: Added a section right after this one

PR #675, opened 2025-12-10
Introduce a `kotlin-library` project type

6hundreds:
As a preparation to `IDEA Sync`, introduced a `kotlin-library` project type and bump kotlin version


buildSrc/src/main/kotlin/profiler.kotlin-library.gradle.kts
```diff
@@ -0,0 +1,15 @@
+import org.gradle.api.tasks.testing.Test
+import org.gradle.internal.os.OperatingSystem
+import org.gradle.kotlin.dsl.withType
+
+plugins {
+    id("profiler.allprojects")
+    kotlin("jvm")
+}
+
+tasks.withType<Test>().configureEach {
```
6hundreds, line 10: Same as `profiler.java-library` has

PR #680, opened 2025-12-11
Introduce cross-version testing framework

alllex:
This borrows part of the gradle/gradle testing infrastructure to organize clean and predictable cross-version tests, when we want to test Gradle Profiler's functionality when running against builds of different Gradle versions.

This is a good reference point: https://github.com/gradle/gradle/blob/master/testing/internal-integ-testing/src/main/groovy/org/gradle/integtests/fixtures/compatibility/AbstractContextualMultiVersionTestInterceptor.java

We take only necessary parts, but also keep some of the flexibility.

TL;DR it's now possible to write integration tests concisely like so:

```groovy
class BuildOperationsTraceGradleCrossVersionTest extends AbstractGradleCrossVersionTest {

    def "can enable build operations trace via command-line flag"() {
        instrumentedBuildScript()

        // The `gradleVersion` is available inside the test automatically <-----------------------
        when:
        run(["--gradle-version", gradleVersion, "--benchmark", "--build-ops-trace", "help"])

        then:
        logFile.containsOne("Build operations trace: ")
    }   
}
```

The test infra will then take care of running the whole spec with all necessary Gradle versions.

### DIFF
--- a/src/test/groovy/org/gradle/profiler/BenchmarkIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/BenchmarkIntegrationTest.groovy
@@ -12,6 +12,8 @@ class BenchmarkIntegrationTest extends AbstractProfilerIntegrationTest {
         brokenBuild(gradleVersion1, 8)
 
         when:
+        warmups = 6
+        iterations = 10
         run([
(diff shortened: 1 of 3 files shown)


asodja (approved):
LGTM, nice feature 💯

PR #686, opened 2025-12-26
Require Java 17 to build and run Gradle Profiler

6hundreds:
This PR sets Java 17 for build and run the project, and also removes CI jobs for Java 8 and Java 11

### DIFF
--- a/buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
+++ b/buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
@@ -6,6 +6,9 @@ plugins {
 description =
     "Project packaged into the Gradle profiler Jar for the further injection to Gradle build. Uses Java 8 for compatibility with older Gradle versions."
 
+// Using sourceCompatibility/targetCompatibility instead of options.release because:
+// - :build-operations needs sun.management APIs and Java 9+ ProcessHandle (multi-version code)
+// - :studio-agent uses --add-exports which is incompatible with --release flag
 java {
     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ The following command line options only apply when measuring Gradle builds:
 
 ## JVM requirements and options
 
-Gradle Profiler requires a Java 8 or later JVM to run.
+Gradle Profiler requires a Java 17 or later JVM to run.
 
 By default, Gradle Profiler starts the daemon for your build using the same JVM that was used to run it.
 In case your build requires an earlier version of the JVM or you require a specific JVM, you can configure it explicitly in your build.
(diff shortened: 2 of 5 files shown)


subprojects/tooling-action/build.gradle.kts
alllex, line 1: Why do we need to extract this subproject?

buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
```diff
@@ -3,6 +3,15 @@ plugins {
     id("groovy")
 }
 
+description =
+    "Project packaged into the Gradle profiler Jar for the further injection to Gradle build. Uses Java 8 for compatibility with older Gradle versions"
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
```
alllex: We should avoid requiring the Java 8 toolchain, and instead use the Java 17 toolchain, but compile to Java 8

subprojects/tooling-action/build.gradle.kts
6hundreds: So we have this build action https://github.com/gradle/gradle-profiler/blob/e64750975c5ce1eac96ea618185bd629217b93cb/subprojects/tooling-action/src/main/java/org/gradle/profiler/toolingapi/FetchProjectPublications.java#L12
If it placed in the main project, then it's requires 17, what is making it impossible for Gradle 6.0 and 6.9.4 to execute it.

However, IDK if any external users uses it. As an alternative we can make `ToolingApiGradleCrossVersionTest` running with Gradle that works on 17

buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
6hundreds: Right

buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
```diff
@@ -3,6 +3,14 @@ plugins {
     id("groovy")
 }
 
+description =
+    "Project packaged into the Gradle profiler Jar for the further injection to Gradle build. Uses Java 8 for compatibility with older Gradle versions."
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
```
6hundreds, line 13: Can't use `release` flag as it too restrictive and required to be removed for https://github.com/gradle/gradle-profiler/blob/master/subprojects/build-operations/build.gradle.kts and https://github.com/gradle/gradle-profiler/blob/master/subprojects/studio-agent/build.gradle.kts

subprojects/studio-plugin/build.gradle.kts
```diff
@@ -35,12 +35,6 @@ project.configurations
     .getByName("testFixturesImplementation")
     .extendsFrom(ideaConfiguration, ideaPluginsConfiguration, intelliJDefaultDependenciesConfiguration)
 
-java {
```
6hundreds, line 38: Set by convention

PR #689, opened 2026-01-12
Use all tested gradle versions

6hundreds:
I was wrong before, when limited Gradle versions available for test.

### DIFF
--- a/src/test/groovy/org/gradle/profiler/fixtures/AbstractBaseProfilerIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/fixtures/AbstractBaseProfilerIntegrationTest.groovy
@@ -111,6 +111,7 @@ abstract class AbstractBaseProfilerIntegrationTest extends AbstractIntegrationTe
             throw new IllegalStateException("This test requires Java Home for $description to be provided in the '$systemProperty' system property")
         }
         return javaHome
+            .replace("\\", "\\\\") // escaping backslashes for Windows
     }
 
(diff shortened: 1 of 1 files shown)


src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationGradleCrossVersionTest.groovy
```diff
@@ -289,7 +289,6 @@ class BuildOperationInstrumentationGradleCrossVersionTest extends AbstractGradle
         configurationCache << [false, true]
     }
 
-    @Requires({ it.instance.gradleVersionWithAdvancedBenchmarking() })
```
6hundreds, line 292: This is set on the class level

src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationGradleCrossVersionTest.groovy
```diff
@@ -336,7 +335,7 @@ class BuildOperationInstrumentationGradleCrossVersionTest extends AbstractGradle
         """
 
         // We don't capture snapshotting time (yet) if the build cache is not enabled
-        file("gradle.properties").text = "org.gradle.caching=true"
```
6hundreds, line 339: It overwrites `gradle.properties` so our downgrading logic doesn't work

6hundreds:
Merging this as CI failures are irrelevant

PR #693, opened 2026-01-23
Migrate to Node Gradle plugin 7.1.0

alllex:
Also bump the Node version to the latest LTS -- 24.13.0

Since we didn't use any removed APIs, no build script migration is necessary.

The 7.1.0 plugin version supports even some Gradle 7.x releases, so we are good with 8.14 in the build.

---

We don't have automated tests, but I did the following manual testing:

- [x] `./gradlew npmInstall` completes successfully
- [x] `./gradlew generateHtmlReportJavaScript` produces output in `build/html-report/`
- [x] `./gradlew testHtmlReports` generates all 5 test reports in `build/test-html-report/`
- [x] `./gradlew build` passes
- [x] Open a generated HTML report in browser and verify it renders correctly
- [x] Check that browserify correctly bundles the JavaScript (verify `mann-whitney-utest` is included)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,7 +12,7 @@ plugins {
     application
     `maven-publish`
     id("profiler.publication")
-    id("com.github.node-gradle.node") version "3.4.0"
+    id("com.github.node-gradle.node") version "7.1.0"
     id("io.sdkman.vendors") version "2.0.0"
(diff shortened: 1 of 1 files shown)


alllex:
There are some CI failures, but they are all unrelated and known flakiness. I rerun some jobs and more of then succeeded.

PR #698, opened 2026-01-26
Add JsonGenerator to output a JSON file that can be directly parsed by custom tooling

### LINKED ISSUE #561: Output a standalone JSON artifact
The current outputs of the gradle profiler are a CSV file and an HTML file that has JSON embedded into it.

I think it would be a nice enhancement to output that JSON in a separate file instead, and let the html file consume it from there.

Happy to look into opening a PR for this since I understand that y'all might have higher priorities. I just want to avoid putting time into this and then hearing "no we don't want that" 😄

remcomokveld:
Closes https://github.com/gradle/gradle-profiler/issues/561

Adding a simple `JsonGenerator` which re-uses the `JsonResultWriter` that the `HtmlGenerator` already uses the generate JSON for HTML file to also output a standalone JSON file

### DIFF
--- a/src/main/java/org/gradle/profiler/report/JsonGenerator.java
+++ b/src/main/java/org/gradle/profiler/report/JsonGenerator.java
@@ -1,0 +1,24 @@
+package org.gradle.profiler.report;
+
+import org.gradle.profiler.InvocationSettings;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.IOException;
+import java.time.Instant;
+
+public class JsonGenerator extends AbstractGenerator {
+    public JsonGenerator(File outputFile) {
+        super(outputFile);
+    }
+
+    @Override
+    protected void write(InvocationSettings settings, BenchmarkResult benchmarkResult, BufferedWriter writer) throws IOException {
+        new JsonResultWriter(true).write(
+            settings.getBenchmarkTitle(),
+            Instant.now(),
+            benchmarkResult.getScenarios(),
+            writer
+        );
+    }
+}
--- a/src/test/groovy/org/gradle/profiler/JsonReportIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/JsonReportIntegrationTest.groovy
@@ -1,0 +1,27 @@
+package org.gradle.profiler
+
+import com.google.gson.Gson
+import org.gradle.profiler.fixtures.AbstractProfilerIntegrationTest
+
+class JsonReportIntegrationTest extends AbstractProfilerIntegrationTest {
+
+    def "json file is written for benchmarks"() {
+        given:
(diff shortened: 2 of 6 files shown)


h0tk3y:

The relevant team for this area will confirm the design of the implementation choices.

src/main/java/org/gradle/profiler/report/JsonGenerator.java
```diff
@@ -0,0 +1,24 @@
...
+import java.time.Instant;
+
+public class JsonGenerator extends AbstractGenerator {
+    public JsonGenerator(File outputFile) {
+        super(outputFile);
+    }
+
+    @Override
+    protected void write(InvocationSettings settings, BenchmarkResult benchmarkResult, BufferedWriter writer) throws IOException {
+        new JsonResultWriter(true).write(
```
6hundreds, line 17: As the initial intention is consuming by tooling, I'm not sure this json has to be pretty. Not a blocker tho, we can keep it like this so far

src/test/groovy/org/gradle/profiler/JsonReportIntegrationTest.groovy
```diff
@@ -0,0 +1,27 @@
...
+            "assemble"
+        ])
+
+        then:
+        checkInstrumentedBuildScriptOutputs(minimalSupportedGradleVersion, "assemble")
+
+        def file = new File(outputDir, "benchmark.json")
+        assert file.isFile()
+        def report = new Gson().fromJson(file.text, Map.class)
+        assert (((report.get("scenarios") as List).get(0) as Map).get("iterations") as List).size() == 2
```
6hundreds: 💭 Let's  assert more structure to lock the JSON schema contract

6hundreds (commented):
Thanks for the contribution!
Readme has to be updated in regard of this changes. See the section https://github.com/gradle/gradle-profiler?tab=readme-ov-file#benchmarking-a-build
LGTM in the rest

6hundreds:
@remcomokveld Hey! Do you have plans to address the feedback to get it to mergeable state?

src/test/groovy/org/gradle/profiler/JsonReportIntegrationTest.groovy
6hundreds: Spock's snapshot was used for golden-file testing

6hundreds:
@remcomokveld Hey, I went ahead and rebased this branch on the fresh master, also added a golden-file testing for resulting JSON and updated the README. This will be merged. 

Important to note, that the resulting JSON doesn't contain stats, which are computed in report.js file. See the follow up issue in this regard https://github.com/gradle/gradle-profiler/issues/836.

Thank you for the contribution to Gradle Profiler!

PR #704, opened 2026-02-03
chore: linting

commandini:
**Summary**

- added some default interface method implementations so that the implementers don't have to define empty-bodied methods to comply with the interface.
- removed unused imports.
- used existing String constants instead of string literals.
- used `Files.delete(...)` instead of `file.delete()` as the former provides more debugging information in case of a failure.
- used `Hashing.murmur3_32_fixed` instead of deprecated `Hashing.murmur3_32`


ov7a:
Considering windows tests as flaky. Overall, changes LGTM

PR #706, opened 2026-02-05
 Introduce "time to X" measurement kinds for build operations

octylFractal:
Allows specifying measurement kinds in the `--measure-build-op`/`measured-build-ops` configuration. From the README update:

The default measurement kind is `cumulative_time`. The available measurement kinds are:
- `cumulative_time`: the cumulative time spent in the build operation, which may be greater than the wall clock time if there are concurrent executions of the build operation type
- `time_to_last_inclusive`: the time from the start of the build until the maximum end time of any execution of the build operation type
- `time_to_first_exclusive`: the time from the start of the build until the minimum start time of any execution of the build operation type


subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * How build operations should be measured in the results. This determines how to go from the potentially
+ * multiple build operations collected during the build to the single value reported in the results.
+ */
+public enum BuildOperationMeasurementKind {
+    /**
+     * Sums the duration of all build operations of the given type.
+     * Useful for measuring total time spent in a given type of operation.
+     */
+    DURATION_SUM,
```
alllex: 💭 We should make it explicit that it's not wall-clock time, but rather something "serial", even if the requested build operations actually run in parallel

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationMeasurer.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param event the finished operation event
+     */
+    void update(OperationFinishEvent event);
+
+    /**
+     * Compute the final measured value.
+     *
+     * @return the final measured value
+     */
+    long computeFinalValue();
```
alllex: 💅 As long as we keep the measurements tied to the time, I'd prefer to make it explicit that those are milliseconds.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationMeasurer.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Implements the correct calculation based on what {@link BuildOperationMeasurementKind} is requested.
+ * Must be thread-safe.
+ */
+interface BuildOperationMeasurer {
+    static BuildOperationMeasurer createForKind(ObjectFactory objectFactory, BuildOperationMeasurementKind kind, long buildStartTime) {
+        switch (kind) {
+            case DURATION_SUM:
+                return objectFactory.newInstance(DurationSumBuildOperationMeasurer.class);
```
alllex: 🤔 Why do these have to be created via an object factory?

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -81,29 +88,33 @@ public BuildOperationTrace measureLocalBuildCache(File outFile) {
...
 
     public BuildOperationTrace measureConfigurationTime(File outFile) {
-        Provider<TimeToFirstTaskRecordingListener> listenerProvider = sharedServices.registerIfAbsent("time-to-first-task", TimeToFirstTaskRecordingListener.class, spec -> {
-            spec.getParameters().getOutputFile().set(outFile);
+        Provider<BuildOperationDurationRecordingListener> listenerProvider = sharedServices.registerIfAbsent("time-to-first-task", BuildOperationDurationRecordingListener.class, spec -> {
+            spec.getParameters().getCapturedBuildOperations().add(
+                new InternalBuildOpMeasurementRequest(
+                    outFile.toPath(),
+                    RunRootBuildWorkBuildOperationType.class.getName(),
+                    BuildOperationMeasurementKind.MAX_END_TIME
```
alllex: ❌ IIUC, this changes the semantics. Previously, the calculation was based on the **start** of the build operation: 
https://github.com/gradle/gradle-profiler/blob/c72ffe9e8d95c55dda5a70e61670cda673f9f48d/subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java#L214

Now we are including all of execution time into the configuration time measurement.

So we probably want to have "Max Start Time".

---

See that `Run tasks` is after we've build the task graph and includes running tasks:
image

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/internal/InternalBuildOpMeasurementRequest.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Represents a request to measure a specific build operation during a Gradle build.
+ */
+public final class InternalBuildOpMeasurementRequest implements Serializable {
```
alllex, line 16: ❓ Why does it have to implement Serializable contract?

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Takes the maximum end timestamp of the build operation(s) of the given type,
+     * and subtracts the start timestamp of the entire build. Useful for
+     * "time to X" measurements.
+     */
+    MAX_END_TIME,
```
alllex: 🤔 I am not sure about the naming. "End time" for me is a point in time, while what we are computing is still a duration. Having a "duration sum" next to it makes it a bit more confusing. The javadoc explains the difference, but what will end up being in the scenario files is the enum symbol. 

How about these options for all the enum's values:
- `total_serial_duration` (while a mouthful, the meaning is in the name, rather than only in the doc; would play well with total_wallclock_duration if we decide to introduce it; I think this also works without "total" prefix, because of an extra qualifier)
- `time_to_last_completed`
- `time_to_first_started` ("min start time")
- `time_to_last_started` (this one for the "max start time" case)
- `time_after_last_completed` (naturally allows measuring "until the end of the build" things, like "sync time after configuration completes")
- `time_after_last_completed` (to keep things symmetrical)

or derivations thereof. I like that in this case the "aggregation" logic is always in the beginning of the symbol. Also, all options immediately suggest support for more than one operation.

src/main/java/org/gradle/profiler/GeneratedInitScript.java
```diff
@@ -45,6 +48,11 @@ private void maybeGenerateInitScript() {
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
+        try {
+            System.err.println(Strings.join(Files.readAllLines(initScript.toPath()), "\n"));
```
alllex: ❓ Is this for debugging?

src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -1,29 +1,63 @@
...
+            String[] parts = value.split(":");
+            if (parts.length != 2) {
+                throw new IllegalArgumentException("Invalid build operation measurement format: " + value);
+            }
+            String buildOperationType = parts[0];
+            BuildOperationMeasurementKind measurementKind;
+            try {
+                measurementKind = BuildOperationMeasurementKind.valueOf(parts[1].toUpperCase(Locale.ROOT));
+            } catch (IllegalArgumentException e) {
+                throw new IllegalArgumentException("Invalid measurement kind: " + parts[1]);
```
alllex: 💅 Since it's user input, it's better to have a parsing method, instead of catching exceptions.

src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -127,7 +167,13 @@ public InvocationSettings parseSettings(String[] args) throws IOException, Setti
         boolean measureLocalBuildCache = parsedOptions.has(measureLocalBuildCacheOption);
         boolean measureConfig = parsedOptions.has(measureConfigTimeOption);
         boolean buildOperationsTrace = parsedOptions.has(buildOpsTraceOption);
-        List<String> benchmarkedBuildOperations = parsedOptions.valuesOf(benchmarkBuildOperation);
+
+        ImmutableSet.Builder<BuildOperationMeasurement> buildOperationMeasurementsBuilder = ImmutableSet.builder();
+        for (String buildOp : parsedOptions.valuesOf(measuredBuildOps)) {
+            buildOperationMeasurementsBuilder.add(new BuildOperationMeasurement(buildOp, BuildOperationMeasurementKind.DURATION_SUM));
```
alllex: 💡 It would be nice, if we'd optionally support `<type>:<measurement>` syntax here as well. Then users can get the right measurement directly from the command line.

src/main/java/org/gradle/profiler/GeneratedInitScript.java
octylFractal: Yes, the removal didn't get in the commit :/

src/main/java/org/gradle/profiler/CommandLineParser.java
octylFractal: I'm not sure what you mean, do you want a switch over each value like this?
```java
switch (parts[1].toUpperCase(Locale.ROOT)) {
  case "DURATION_SUM": BuildOperationMeasurementKind.DURATION_SUM;
  case "MAX_END_TIME": BuildOperationMeasurementKind.MAX_END_TIME;
  default:
    throw new IllegalArgumentException("Invalid measurement kind: " + parts[1]);
}
```

src/main/java/org/gradle/profiler/CommandLineParser.java
octylFractal: I'm confused what you mean with the use case. `--build-op-measurements` lets you get "the right measurement directly from the command line"? I think I'm fine supporting the format in both options and just making them aliases, but I want to be sure I'm not missing something.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/internal/InternalBuildOpMeasurementRequest.java
octylFractal: This is a parameter to a build service (`BuildOperationDurationRecordingListener`), so it must be able to go through the isolation mechanism there. If it's not `Serializable`, then this happens:
```
BuildOperationInstrumentationGradleCrossVersionTest > can benchmark configuration time(configuration-cache: #configurationCache) > can benchmark configuration time(configuration-cache: false) Gradle 9.3.1 FAILED
    org.gradle.profiler.Main$ScenarioFailedException at BuildOperationInstrumentationGradleCrossVersionTest.groovy:147
        Caused by: org.gradle.tooling.BuildException at BuildOperationInstrumentationGradleCrossVersionTest.groovy:147
            Caused by: org.gradle.internal.exceptions.LocationAwareException at DefaultExceptionAnalyser.java:97
                Caused by: org.gradle.internal.snapshot.impl.IsolationException at DefaultIsolatableFactory.java:51
                    Caused by: org.gradle.internal.snapshot.ValueSnapshottingException at AbstractValueProcessor.java:223
                        Caused by: java.io.NotSerializableException at AbstractValueProcessor.java:21
```

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationMeasurer.java
octylFractal: It doesn't need to any more actually, since we pass the `buildStartTime` differently. I was doing the same code path on both sides for consistency.

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationMeasurer.java
octylFractal: I went for using `Duration` here instead, so it's not just a documentation thing.

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: Something like this, but moved closer to the enum (or into the enum) class, and returning null, instead of catch-rethrow dance. This can help with reuse too.

It would also be nice, if we tell which values are possible in this error.

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: I'd like to be able to do `--measure-build-op MyBuildOp:max_end_time`.

I see now that I might have missed the `--build-op-measurements`. Do you think we really need two separate CLI options?

It's a common pattern for many Profiler options to allow multiple occurrences, that are collected all together. So it might be simpler to keep a single CLI arg.

src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -58,10 +92,16 @@ public InvocationSettings parseSettings(String[] args) throws IOException, Setti
...
         OptionSpecBuilder benchmarkOption = parser.accepts("benchmark", "Collect benchmark metrics");
-        ArgumentAcceptingOptionSpec<String> benchmarkBuildOperation = parser.accepts(
+        ArgumentAcceptingOptionSpec<String> measuredBuildOps = parser.accepts(
             "measure-build-op",
             "Collect benchmark metrics for build operation"
         ).withRequiredArg();
+        ArgumentAcceptingOptionSpec<BuildOperationMeasurement> buildOpMeasurements = parser.accepts(
+            "build-op-measurements",
+            "Collect specific measurements for a given build operation (format: <build operation>:<measurement kind>, where <measurement kind> is one of " +
+                Stream.of(BuildOperationMeasurementKind.values()).map(BuildOperationMeasurementKind::name).collect(Collectors.joining(", ")) +
```
alllex: 💅  Having all-uppercase words in a scenario file is okay, but I find it a bit intimidating, when a tool hints that I should type them in the console too. I'd make the printing lower case, maybe.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public static Set<String> getValidValues() {
+        return Stream.of(values())
+            .map(v -> v.name().toLowerCase(Locale.ROOT))
+            .sorted()
+            .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    @Override
+    public String toString() {
```
alllex: 💭 If we use for something other than debugging, it might be better to keep this in an explicit `toDisplayName()` method.

subprojects/build-operations/src/test/groovy/org/gradle/trace/buildops/BuildOperationMeasurerTest.groovy
```diff
@@ -0,0 +1,14 @@
+package org.gradle.trace.buildops
+
+import org.gradle.profiler.buildops.BuildOperationMeasurementKind
+import spock.lang.Specification
+
+class BuildOperationMeasurerTest extends Specification {
```
alllex, line 6: 💭 Since we can't easily test measurer logic in the integration tests, it would be good to have it covered in the unit tests.

subprojects/build-operations-measuring/src/test/groovy/org/gradle/profiler/buildops/BuildOperationMeasurementKindTest.groovy
```diff
@@ -0,0 +1,42 @@
+package org.gradle.profiler.buildops
+
+import spock.lang.Specification
+
+import java.util.stream.Stream
+
+class BuildOperationMeasurementKindTest extends Specification {
```
alllex, line 7: 👍

src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationGradleCrossVersionTest.groovy
```diff
@@ -240,6 +251,177 @@ class BuildOperationInstrumentationGradleCrossVersionTest extends AbstractGradle
...
+        def firstWarmup = lines.get(4)
+        def configureBuildOp = firstWarmup =~ /warm-up build #1,$SAMPLE,($SAMPLE)/
+        configureBuildOp.matches()
+        Double.valueOf(configureBuildOp[0][1]) > 1
+
+        lines.get(9).matches("warm-up build #6,$SAMPLE,$SAMPLE")
+        lines.get(10).matches("measured build #1,$SAMPLE,$SAMPLE")
+
+        where:
+        [via, commandLine, scenarioConfiguration, configurationCache] << [
```
alllex, line 304: 💭 We should upgrade to the latest Spock at some point. It supports multiple `where:` clauses for cross-products, IIRC

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
```diff
@@ -0,0 +1,84 @@
...
+     * @return the corresponding BuildOperationMeasurementKind
+     * @throws IllegalArgumentException if the value is not a valid {@link BuildOperationMeasurementKind}
+     */
+    public static BuildOperationMeasurementKind fromString(String value) {
+        switch (value.toLowerCase(Locale.ROOT)) {
+            case "duration_sum":
+                return DURATION_SUM;
+            case "time_to_last_completed":
+                return TIME_TO_LAST_COMPLETED;
+            case "time_to_first_started":
```
alllex: 💭 What if we make it `time_to_first` (build operation)?

I feel like the semantics does not change at all, but less to type. We can keep the enum name to be more explicit.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @param value the string value to convert
+     * @return the corresponding BuildOperationMeasurementKind
+     * @throws IllegalArgumentException if the value is not a valid {@link BuildOperationMeasurementKind}
+     */
+    public static BuildOperationMeasurementKind fromString(String value) {
+        switch (value.toLowerCase(Locale.ROOT)) {
+            case "duration_sum":
+                return DURATION_SUM;
+            case "time_to_last_completed":
```
alllex: 💭 Here is my last alternative proposal, for you to take or discard.

How about `time_to_last_inclusive`? The `inclusive` suffix is more clearly about taking into account the duration of the build operation in question. I think it also play nicely with having the shorter forms representing the exclusive variant -- `time_to_first` and (someday) `time_to_last`.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
```diff
@@ -0,0 +1,84 @@
...
+     * Convert a string value to the corresponding BuildOperationMeasurementKind.
+     * The string values are case-insensitive and must be contained in the set returned by {@link #getValidValues()}.
+     *
+     * @param value the string value to convert
+     * @return the corresponding BuildOperationMeasurementKind
+     * @throws IllegalArgumentException if the value is not a valid {@link BuildOperationMeasurementKind}
+     */
+    public static BuildOperationMeasurementKind fromString(String value) {
+        switch (value.toLowerCase(Locale.ROOT)) {
+            case "duration_sum":
```
alllex: 💭 I'm still on the fence with this one. We are planning to introduce the wall clock version. Then it would be good to have two things that are commonly recognizable as a pair of common.

Popular counterparts to the wall clock time the profiling contexts seem to be "total time", "cumulative time" or "total time". There is also "CPU time", but in our case it might be more confusing to bring in this terminology.

Having said that, it's still a suggestion. I'm not planning to die on this hill.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
octylFractal: I'd prefer not to detach these keywords from the enum name, it makes validating them in tests more difficult and increases differences between the different API surfaces. Plus, we may introduce a "time to last started" or "time to first completed" in the future.

subprojects/build-operations/src/test/groovy/org/gradle/trace/buildops/BuildOperationMeasurerTest.groovy
octylFractal: I'll do that, but I'd like to introduce http://jqwik.net/ alongside it to get better coverage so I'll do it in a separate PR.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
octylFractal: I think that makes sense.

subprojects/build-operations-measuring/src/main/java/org/gradle/profiler/buildops/BuildOperationMeasurementKind.java
octylFractal: I guess "cumulative time" is basically the same as "duration sum", I can use that.

src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationGradleCrossVersionTest.groovy
```diff
@@ -240,6 +251,177 @@ class BuildOperationInstrumentationGradleCrossVersionTest extends AbstractGradle
         ].combinations().collect { it[0] + it[1] }
     }
 
+    @Requires({ it.instance.gradleVersionWithExperimentalConfigurationCache() })
+    def "can benchmark time to end of last configure build op via #via(configuration-cache: #configurationCache)"() {
```
alllex, line 255: Something I missed is that the CC enablement doesn't add much to the test itself. I think we should split this in at least two tests. Have only one test that does CC vs no-CC, and other tests that check all the variants of measurements.

src/test/groovy/org/gradle/profiler/BuildOperationInstrumentationGradleCrossVersionTest.groovy
```diff
@@ -240,6 +251,177 @@ class BuildOperationInstrumentationGradleCrossVersionTest extends AbstractGradle
...
+        lines.get(10).matches("measured build #1,$SAMPLE,$SAMPLE,$SAMPLE")
+
+        where:
+        [via, commandLine, scenarioConfiguration, configurationCache] << [
+            [
+                [
+                    'command line',
+                    [
+                        "--measure-build-op", "org.gradle.initialization.ConfigureBuildBuildOperationType:cumulative_time",
+                        "--measure-build-op", "org.gradle.initialization.ConfigureBuildBuildOperationType:time_to_last_inclusive"
```
alllex, line 393: We are missing a test for the time to first. It's implicitly covered by the configuration time test, but it would be nicer to be explicit here.

I'd also add a check that we actually measure all possible measurement kinds (3 at this time), so that if we add a new one the test fails.

This test should also be simplified to not involve CC.

PR #707, opened 2026-02-05
Create scenario-driven heap profiling

hegyibalint:
This PR adds a new memory profiler able to dump the heap of a Gradle execution at specific execution points.
The current set of scenarios are:
 - `config-end`: When we leave configuration and enter execution
 - `build-end`: When the build is _just_ finished.


hegyibalint:
I was a bit too fast pressing the button for making this "Ready for Review", I can see that we have a bunch of failing tests.
@asodja please ignore until this is RFR

hegyibalint:
I will close this for now. I think JFRs and other tricks might provide us a better tooling to do the same.

PR #708, opened 2026-02-05
Produce Perfetto trace for build operations

alllex:
Incorporates the https://github.com/gradle/gradle-trace-converter tool into Gradle Profiler to automatically produce Perfetto traces, whenever the `--build-ops-trace` is requested.

Next to the `<scenario-name>-log.txt` build operations log file, a `<scenario-name>.perfetto.proto` is created. The conversion happens after the full scenario has been executed, and before the next one, and thus doesn't affect any measurements.

Note that same as the build operations log, the Perfetto trace is only available for the last iteration of the benchmark.

---

The build operation trace has limited use on its own, while Perfetto provides a [full-featured UI](https://ui.perfetto.dev/) to view and analyze the traces in detail. Having this functionality built in makes it much easier to investigate performance issues or improvement opportunities.

Since the build operation trace can be produced for any Gradle invocation, this means that the Perfetto traces are available for all of them as well, including Android Studio sync.

### Standalone converter app

In addition to the automatic conversion in the profiler, a standalone `gtc` (gradle-trace-converter) command-line tool is provided as a separate subproject (`:gradle-trace-converter-app`). It can convert any Gradle build operation trace to Perfetto format independently of the profiler — useful for traces captured manually from command-line builds or IDE syncs. See its [README](subprojects/gradle-trace-converter-app/README.md) for installation and usage instructions.

### Implementation

This PR borrows the code almost directly from https://github.com/gradle/gradle-trace-converter with little adjustments, removing not currently required functionality to simplify integration.

The conversion logic lives in the `:perfetto-trace` library subproject, shared by both the profiler and the standalone app. Convention plugins (`profiler.java-base`, `profiler.kotlin-base`, etc.) were introduced to standardize build configuration across subprojects.

Importantly, all the proto definitions are copied as-is without changes. It's possible that many of them are not strictly required (even in the original implementation), but I opted for getting a working solution first. We can trim it later.


src/main/java/org/gradle/profiler/gradle/GradleScenarioInvoker.java
```diff
@@ -164,6 +180,34 @@ public void run(GradleScenarioDefinition scenario, InvocationSettings settings,
             gradleClient.close();
             daemonControl.stop(buildConfiguration);
         }
+
+        if (scenario.isBuildOperationsTrace()) {
+            convertBuildOperationsToPerfetto(scenario);
+        }
```
alllex, line 186: This is the main change

subprojects/perfetto-trace/src/main/kotlin/org/gradle/profiler/perfetto/ChromeTraceWriter.kt
```diff
@@ -0,0 +1,268 @@
...
+typealias CtThreadId = Int
+
+/**
+ * Chrome-trace track ID.
+ *
+ * Structurally, a thread lane within a [process][CtProcessId].
+ */
+typealias CtTrackUuid = Long
+
+internal class ChromeTraceWriter(val outputFile: File) : BuildOperationVisitor, AutoCloseable {
```
alllex, line 39: Copied from https://github.com/gradle/gradle-trace-converter/blob/cebfefbc131543831a910e60d73b1f66da9a5129/app/src/main/kotlin/org/gradle/tools/trace/app/TraceToChromeTraceConverter.kt

subprojects/perfetto-trace/src/main/proto/perfetto/common/BUILD.gn
alllex, line 1: All proto files are copied verbatim from https://github.com/gradle/gradle-trace-converter/tree/cebfefbc131543831a910e60d73b1f66da9a5129/app/src/main/proto/

src/main/java/org/gradle/profiler/gradle/GradleScenarioInvoker.java
reinsch82: This means, there is no way to disable perfetto trace generation when `build-ops-trace` is `true`, right?
should we be able to disable it?

reinsch82 (approved):
LGTM

src/main/java/org/gradle/profiler/gradle/GradleScenarioInvoker.java
alllex: For now, Perfetto traces is the only way we analyze the full build operation trace produced by the Profiler. At least I'm not aware of a case, when this would be undesirable. But we can always introduce a disablement flag later, if the need arises.

PR #717, opened 2026-02-09
Upgrade auto-downloaded Async Profiler to 4.3

alllex:
Includes bug fixes and introduces new types of profiling we might make use of in the future

https://github.com/async-profiler/async-profiler/releases/tag/v4.3

### DIFF
--- a/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerCompatibility.java
+++ b/src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerCompatibility.java
@@ -3,7 +3,7 @@ package org.gradle.profiler.asyncprofiler;
 public class AsyncProfilerCompatibility {
 
     // https://github.com/async-profiler/async-profiler/releases/
-    public static final AsyncProfilerVersion AUTO_DOWNLOAD_VERSION = new AsyncProfilerVersion("4.2");
+    public static final AsyncProfilerVersion AUTO_DOWNLOAD_VERSION = new AsyncProfilerVersion("4.3");
 
     public static final AsyncProfilerVersion MINIMUM_SUPPORTED_VERSION = new AsyncProfilerVersion("3.0");
 


src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerCompatibility.java
```diff
@@ -0,0 +1,10 @@
+package org.gradle.profiler.asyncprofiler;
+
+public class AsyncProfilerCompatibility {
+
+    // https://github.com/async-profiler/async-profiler/releases/
+    public static final AsyncProfilerVersion AUTO_DOWNLOAD_VERSION = new AsyncProfilerVersion("4.3");
```
alllex, line 6: ℹ️  This is the main change: `4.2 => 4.3`

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -1,86 +1,58 @@
...
-        File asprofBinary = new File(home, "bin/asprof");
-        if (asprofBinary.isFile()) {
-            executable = asprofBinary;
-            String libExtension;
-            // if async profiler 3+
-            if (OperatingSystem.isLinuxX86()) {
-                libExtension = ".so";
-            } else if (OperatingSystem.isMacOS()) {
-                libExtension = ".dylib";
-            } else {
```
alllex, line 61: ℹ️ This was previously not handling Linux Aarch64 properly, which is now handled by consolidated `AsyncProfilerPlatform` selection

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerDistribution.java
```diff
@@ -91,42 +63,44 @@ public File getLibrary() {
...
-                throw new IllegalArgumentException("Invalid version: " + stringVersion);
-            }
-            try {
-                major = Integer.parseInt(parts[0]);
-                minor = Integer.parseInt(parts[1]);
-            } catch (NumberFormatException e) {
-                throw new IllegalArgumentException("Invalid version: " + stringVersion, e);
-            }
-            this.stringVersion = stringVersion;
+    public static AsyncProfilerDistribution of(AsyncProfilerPlatform platform, File home, String sourceDisplayName) {
```
alllex, line 73: ℹ️ This method is what the constructor was before, moved, no changes in the logic

src/main/java/org/gradle/profiler/asyncprofiler/AsyncProfilerVersion.java
```diff
@@ -0,0 +1,52 @@
+package org.gradle.profiler.asyncprofiler;
+
+/**
+ * Async Profiler Version.
+ * <p>
+ * The versions are usually in the form <code>major.minor</code>, like {@code 3.0} or {@code 4.2}.
+ * Occasionally, there are patch releases like {@code 4.2.1}.
+ */
+public class AsyncProfilerVersion implements Comparable<AsyncProfilerVersion> {
```
alllex, line 9: ℹ️  Added support for patch versions, since `4.2.1` exists

PR #719, opened 2026-02-12
Assert help output

alllex:
Improve the output of the `--help` command and add a test demostraiting what it looks like

### DIFF
--- a/src/main/java/org/gradle/profiler/CommandLineParser.java
+++ b/src/main/java/org/gradle/profiler/CommandLineParser.java
@@ -133,7 +133,7 @@ class CommandLineParser {
             return null;
         }
 
-        File projectDir = toAbsoluteFileOrNull(parsedOptions.valueOf(projectOption), () -> new File("."));
+        File projectDir = toAbsoluteFileOrElse(parsedOptions.valueOf(projectOption), () -> new File("."));
         boolean hasProfiler = parsedOptions.has(profilerOption);
         ProfilerFactory profilerFactory = ProfilerFactory.NONE;
         if (hasProfiler) {
@@ -148,7 +148,7 @@ class CommandLineParser {
             return fail(parser, "Neither --profile or --benchmark specified.");
         }
 
-        File outputDir = toAbsoluteFileOrNull(parsedOptions.valueOf(outputDirOption), this::findOutputDir);
+        File outputDir = toAbsoluteFileOrElse(parsedOptions.valueOf(outputDirOption), this::findOutputDir);
         File gradleUserHome = toAbsoluteFileOrNull(parsedOptions.valueOf(gradleUserHomeOption));
         Integer warmups = parsedOptions.valueOf(warmupsOption);
         Integer iterations = parsedOptions.valueOf(iterationsOption);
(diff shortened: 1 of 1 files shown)


src/main/java/org/gradle/profiler/jprofiler/JProfilerProfilerFactory.java
```diff
@@ -20,7 +20,7 @@ public class JProfilerProfilerFactory extends ProfilerFactory {
     @Override
     public void addOptions(OptionParser parser) {
         homeDir = parser.accepts("jprofiler-home", "JProfiler installation directory").availableIf("profile")
-            .withOptionalArg().ofType(String.class).defaultsTo(JProfiler.getDefaultHomeDir());
+            .withRequiredArg().ofType(String.class);
```
octylFractal, line 23: Why remove the default here? Does it cause weird output?

src/main/java/org/gradle/profiler/CommandLineParser.java
```diff
@@ -266,4 +266,9 @@ private void showVersion() {
     private File toAbsoluteFileOrNull(@Nullable File file) {
         return file == null ? null : file.getAbsoluteFile();
     }
+
+    private File toAbsoluteFileOrNull(@Nullable File file, Supplier<File> orElse) {
```
octylFractal: I think using the same name here is a little confusing. What about `toAbsoluteFileOrElse`?

src/main/java/org/gradle/profiler/jprofiler/JProfilerProfilerFactory.java
alllex: I removed the defaults in a couple places, because they are effectively computing the actual value, and then printing it. For instance, here it would call `getDefaultHome()` which returns a different path depending on the OS. While I guess it might be helpful, I don't like that `--help` output contains absolute paths.

I don't particularly like my solution, but I didn't find anything better quickly. Ideally, I'd preserve the locality of the parameter definition with its default, if there is a way to customize what gets printed in help for that default.

I'm also not really a fan of all default being computed eagerly. Though, it does not seem to create problems here.

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: Sounds good 👍

src/main/java/org/gradle/profiler/jprofiler/JProfilerProfilerFactory.java
octylFractal: I do find it helpful if the program can tell me exactly where it thinks its looking, rather than needing to try and interpret the value. But it's not a big deal.

Is it possible to add an explanation of the default here like with the other args?

alllex: It doesn't seem like the library we use provides a good way of doing it. For many options, we already render the possible values and defaults by hand in the description.

We should probably find/implement a generic solution for this and apply it to all parameters

src/main/java/org/gradle/profiler/CommandLineParser.java
alllex: Renamed

PR #722, opened 2026-02-12
Add wall clock time build op measurerer

octylFractal:
Based on #721 to avoid conflicts.

Adds Guava to `build-operations` in order to use `RangeMap`, which is kind of expensive size-wise. It might be worth shading it into the `build-operations` plugin.

### DIFF
--- a/subprojects/build-operations/src/test/groovy/org/gradle/trace/buildops/BuildOperationMeasurerTest.groovy
+++ b/subprojects/build-operations/src/test/groovy/org/gradle/trace/buildops/BuildOperationMeasurerTest.groovy
@@ -1,14 +1,210 @@
 package org.gradle.trace.buildops
 
+import org.gradle.internal.operations.OperationFinishEvent
 import org.gradle.profiler.buildops.BuildOperationMeasurementKind
 import spock.lang.Specification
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/gradle/profiler/buildops/BuildOperationInstrumentation.java
```diff
@@ -54,9 +54,12 @@ public boolean requiresInitScript() {
...
     }
 
-    private static Path createBuildOperationTempFile(String op) {
+    private static Path createBuildOperationTempFile(BuildOperationMeasurement op) {
         try {
-            Path tempFile = Files.createTempFile("gradle-profiler", "build-ops-" + op);
+            Path tempFile = Files.createTempFile(
+                "gradle-profiler",
+                "build-ops-" + op.getBuildOperationType() + "_" + op.getMeasurementKind().name()
+            );
```
octylFractal, line 62: This is a bug fix for measuring multiple ops to work at all. I think the test didn't catch it because it's a race condition, but the new tests I added worked more reliably to find this problem.

alllex:
I added some basic unit tests to sanity check the measures. It was spotless!

PR #724, opened 2026-02-13
Check description of published projects

alllex:
We require each published project to have a description:
https://github.com/gradle/gradle-profiler/blob/27fae0ff35f81b240693e4d17b9e01c8351408fd/buildSrc/src/main/kotlin/profiler.publication.gradle.kts#L19

However, this is currently only checked when the publication is attempted, which leads to situations, when new published subprojects are created without an appropriate description.

This PR introduces `checkDescription` task, which validates the presence of the `project.description` for the eligible projects.

As the Tests Trigger on TeamCity runs actual tests across operating systems and JVM versions, it doesn't seem appropriate to run this check there. To address that, this PR introduces a `sanityCheck` lifecycle task, which is then configured to depend on `checkDescription` (only on it for now). The `sanityCheck` has no relation to `check` or `test`, so it has to be expected separately. Thus, this PR introduces a new CI job `Sanity Check` that runs in all the same scenarios that Tests run, but only on a single platform/JVM, since the checks are agnostic to that.

### DIFF
--- a/.teamcity/settings/GradleProfilerSanityCheck.kt
+++ b/.teamcity/settings/GradleProfilerSanityCheck.kt
@@ -20,7 +20,6 @@ object GradleProfilerSanityCheck : BuildType({
             tasks = "clean sanityCheck"
             buildFile = ""
             gradleParams = "-s --build-cache ${toolchainConfiguration(os, arch)}"
-            param("org.jfrog.artifactory.selectedDeployableServer.defaultModuleVersionConfiguration", "GLOBAL")
         }
     }
 
--- a/.teamcity/settings/GradleProfilerPublishing.kt
+++ b/.teamcity/settings/GradleProfilerPublishing.kt
@@ -37,7 +37,6 @@ object GradleProfilerPublishing : BuildType({
         gradle {
             tasks = "clean createBuildReceipt publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag %additional.gradle.parameters%"
             gradleParams = toolchainConfiguration(os, arch) + " -Dgradle.cache.remote.push=true"
-            param("org.jfrog.artifactory.selectedDeployableServer.defaultModuleVersionConfiguration", "GLOBAL")
             buildFile = ""
         }
     }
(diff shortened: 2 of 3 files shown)


buildSrc/src/main/kotlin/profiler.embedded-library.gradle.kts
```diff
@@ -1,11 +1,10 @@
+// Project packaged into the Gradle profiler Jar for the further injection to Gradle build.
+// Uses Java 8 for compatibility with older Gradle versions.
 plugins {
     id("profiler.java-library")
     id("groovy")
 }
 
-description =
```
alllex, line 6: 💭  We don't want a generic description to be accidentally used for publishing.

.teamcity/settings/GradleProfilerSanityCheck.kt
```diff
@@ -0,0 +1,40 @@
...
+    params {
+        javaHome(os, arch, JavaVersion.OPENJDK_17)
+    }
+
+    steps {
+        gradle {
+            tasks = "clean sanityCheck"
+            buildFile = ""
+            gradleParams = "-s --build-cache ${toolchainConfiguration(os, arch)}"
+            param("org.jfrog.artifactory.selectedDeployableServer.defaultModuleVersionConfiguration", "GLOBAL")
```
cobexer: ❓ What is this supposed to do and where should it appear? I don't see it on TeamCity

cobexer (changes requested):

.teamcity/settings/GradleProfilerSanityCheck.kt
alllex: I don't know exactly, and I hoped that you'd know.

It appears in other jobs, including the "Profiler Test" job:
https://github.com/search?q=repo%3Agradle%2Fgradle-profiler%20%22org.jfrog.artifactory.selectedDeployableServer.defaultModuleVersionConfiguration%22&type=code

I can imagine it's useful for the publication job. Maybe it was copy-pasted into the Test job?

cobexer: From what I could dig up it is supposed to configure the TeamCity Artifactory Plug-in

https://github.com/search?q=repo%3Ajfrog%2Fteamcity-artifactory-plugin%20defaultModuleVersionConfiguration&type=code

That is not installed, please remove the parameter (in all places)

cobexer: FTR: I created a PR to clean it up in g/g https://github.com/gradle/gradle/pull/36791

alllex: Removed. PTAL!

PR #727, opened 2026-02-21
Add IntelliJ IDEA support and remove Android Studio dependency from IDE plugin

### LINKED ISSUE #546: Add support for profiling Gradle syncs with IntelliJ 2023.3+
We support IntelliJ sync profiling (it's a bit undocumented but it works).

With IntelliJ 2023.3 Android plugin is not embedded anymore in IntelliJ distribution. Since we use Android plugin APIs to do a sync that means that we don't support IntelliJ 2023.3 at the moment. So we have to use a different API for IntelliJ. 

That should not be the case for Android Studio, since Android plugin is embedded there.   

I believe the problem is using this API
```
ProjectSystemUtil.getSyncManager(project).syncProject()
```
[AndroidStudioSystemHelper.java#L61](https://github.com/gradle/gradle-profiler/blob/72aa9f458a21bce1553e2dca185fc82688d2bc21/subprojects/studio-plugin/src/main/java/org/gradle/profiler/studio/plugin/system/AndroidStudioSystemHelper.java#L61), that is only available in AS.

Instead, we could use for IntelliJ something like:
```
ImportSpec importSpec = new ImportSpecBuilder(project, GradleConstants.SYSTEM_ID)
    .use(ProgressExecutionMode.NO_PROGRESS_SYNC) // Or maybe ProgressExecutionMode.MODAL_SYNC
    .callback(new ExternalProjectRefreshCallback() {
        // Implement callback onSuccess/onFailure methods
    })
    .build();
ExternalSystemUtil.refreshProject(project.getBasePath(), importSpec);
```

joshfriend:
Replaces the Android-Studio-specific `studio-plugin` with a new `intellij-plugin` that works with any IntelliJ-based IDE using the standard `ExternalSystemUtil` API for Gradle sync . The plugin now targets IntelliJ 2025.1+ using the IntelliJ Platform Gradle Plugin 2.x, and all user-facing identifiers (`android-studio-sync`, `--studio-install-dir`, etc.) are renamed to IDE-neutral equivalents with backward-compatible deprecated aliases.

Depends on #728

Closes #546.


ov7a:

The relevant team for this area will confirm the implementation choices.

PR #755, opened 2026-03-30
Fix flaky `DifferentialFlameGraphIntegrationTest`

6hundreds:
This test was flaky: https://ge.gradle.org/scans/tests?search.relativeStartTime=P28D&search.rootProjectNames=gradle-profiler&search.timeZoneId=Europe%2FMoscow&tests.container=org.gradle.profiler.DifferentialFlameGraphIntegrationTest

The current theory is that because of the test build has no meaningful work, sometimes it finishes too soon for JFR to writes CPU events. We add some dummy work to the build. Mean execution time for tests in this class was ~15sec:
https://ge.gradle.org/scans/tests?search.rootProjectNames=gradle-profiler&search.timeZoneId=Europe%2FMoscow&tests.container=org.gradle.profiler.DifferentialFlameGraphIntegrationTest&tests.sortField=MEAN_DURATION

CI run for this branch shows it increased to 19-25sec, it's not representative 100%, since it's the only build we've seen. However, I'd expect some increasing in mean execution time for this in favor of making it stable.

Checking stacks file for emptiness is made for `JFRControl` parity with `AsyncProfilerController` to fail loudly if there're no stack files

### DIFF
--- a/src/test/groovy/org/gradle/profiler/DifferentialFlameGraphIntegrationTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/DifferentialFlameGraphIntegrationTest.groovy
@@ -16,15 +16,16 @@ class DifferentialFlameGraphIntegrationTest extends AbstractProfilerIntegrationT
         // Append CPU-intensive work so profilers (JFR, async-profiler) reliably capture samples.
         // JFR samples at 10ms intervals, so each task needs to burn at least ~50ms of CPU.
         buildFile << """
-tasks.configureEach {
-    doLast {
-        def hash = 0
-        for (int i = 0; i < 10_000_000; i++) {
-            hash = hash * 31 + i
-        }
-    }
-}
-"""
+            tasks.configureEach {
+                doLast {
+                    def hash = 0
+                    for (int i = 0; i < 10_000_000; i++) {
+                        hash = hash * 31 + i
+                    }
+                    println(hash)
+                }
+            }
+        """
     }
 
     def "generates differential flame graphs with #profiler"() {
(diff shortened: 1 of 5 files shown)


src/test/groovy/org/gradle/profiler/DifferentialFlameGraphIntegrationTest.groovy
```diff
@@ -9,6 +9,24 @@ import spock.lang.Requires
...
+        super.instrumentedBuildScript()
+        // Append CPU-intensive work so profilers (JFR, async-profiler) reliably capture samples.
+        // JFR samples at 10ms intervals, so each task needs to burn at least ~50ms of CPU.
+        buildFile << """
+tasks.configureEach {
+    doLast {
+        def hash = 0
+        for (int i = 0; i < 10_000_000; i++) {
+            hash = hash * 31 + i
+        }
```
alllex: As JMH teaches us, JIT is smart enough to usually optimize this loop out, because the value is not consumed anywhere. We should print the last computed number to the console to ensure this doesn't happen.

src/test/groovy/org/gradle/profiler/DifferentialFlameGraphIntegrationTest.groovy
```diff
@@ -9,6 +9,24 @@ import spock.lang.Requires
...
 @Requires({ it.instance.isCurrentJvmSupportsMultipleGradleVersions() })
 class DifferentialFlameGraphIntegrationTest extends AbstractProfilerIntegrationTest implements FlameGraphFixture {
+
+    @Override
+    def instrumentedBuildScript() {
+        super.instrumentedBuildScript()
+        // Append CPU-intensive work so profilers (JFR, async-profiler) reliably capture samples.
+        // JFR samples at 10ms intervals, so each task needs to burn at least ~50ms of CPU.
+        buildFile << """
+tasks.configureEach {
```
alllex: 💅 Can we indent this?

alllex (approved):
LGTM with some suggestions.

It would be great to get rid of the flakiness!

Issue #327, opened 2021-03-18
Make warmup parameter more customizable


pvegh:
I'd like to be able to set warmup to be 0. Currently the minimum is 1.

It should be my decision because:
* maybe I want to test on current temperatures
* maybe it's e.g. a Laptop and warming it up even more defies the purpose of my testing

Also warmup for the 2nd and later measured builds seem unnecessary to me since we're already warmed up.
Consecutive warmups make the machine uselessly hot and slow if it's e.g. a laptop. Maybe it's NOT what I want to be measuring.

IgnatBeresnev:
Our plugin is expericning OOM issues and the only build I can reliably reproduce it on takes 3 hours to build. I'd really like to skip the warmup part and just profile the heap with async profiler (also not even sure if warmup makes sense for heap profiling)

mr1chter:
I want to add that the most simple case of profiling a build will always warm up not only the daemon, but also the project, giving UP-TO-DATE tasks everywhere. But there is no simple commandline option for cleanup-tasks, so the "Advanced profiling scenarios" have to be used even in the most simplest cases.

rdehuyss:
I also tried setting warmup to `0` as our build also takes quite some time. The profiler failed immediately without any reason:

```
➜  jobrunr-pro git:(feature/v8/pro/carbon-aware-jobs) ✗ gradle-profiler --profile async-profiler-heap core:test --warmups 0 --iterations 1

* Writing results to /Users/rdehuyss/Projects/JobRunr/jobrunr-pro/profile-out-5

* Settings
Project dir: /Users/rdehuyss/Projects/JobRunr/jobrunr-pro
Output dir: /Users/rdehuyss/Projects/JobRunr/jobrunr-pro/profile-out-5
Profiler: async profiler
Benchmark: false
Versions: []
Gradle User Home: /Users/rdehuyss/Projects/JobRunr/jobrunr-pro/gradle-user-home
Targets: [core:test]
Warm-ups: 0
Builds: 1

* Inspecting the build using its default Gradle version

* Stopping daemons

There were some problems with the profiler configuration:
- Scenario using Gradle 8.5: You can not skip warm-ups when profiling or benchmarking a Gradle build. Use --no-daemon or --cold-daemon if you want to profile or benchmark JVM startup

java.lang.IllegalArgumentException: There were some problems with the profiler configuration. Please see the log output for details.
[...]

PR #762, opened 2026-04-07
Introduce --single-shot

6hundreds:
This PR introduces `--single-shot` mode, which implies `warmups 0, iterations 1, cold-daemon`, `--no-daemon` can be used as well.

The main driver for this change is GProfiler usage in `gradle/gradle` for running `smoke-ide-test` test suite - we need to be able to run sync exactly once, unless some special scenarios like incremental sync test with IP come into play. 

We also have a public issue in this regard https://github.com/gradle/gradle-profiler/issues/327.


jvandort:
I'm not sure I totally agree with this here

If smoke-ide-test only needs to run sync once, why does it use gradle-profiler at all? Gradle profiler should be used for profiling and benchmarking

asodja (approved):
LGTM

It makes sense to me. I see usecase, especially with big builds and heap memory measurements

6hundreds:
@jvandort 
> If smoke-ide-test only needs to run sync once, why does it use gradle-profiler at all?

We don't have another developed options to trigger sync. We have `gradle-ide-starter` but the approach taken there is flaky enough. Using GProfiler is a sane option to trigger sync without re-implementing IDE plugin.

jvandort:
> We have gradle-ide-starter but the approach taken there is flaky enough. Using GProfiler is a sane option to trigger sync without re-implementing IDE plugin

Sounds like we should update `gradle-ide-starter` to use the working implementation that gradle profiler uses, and then gradle profiler should delegate to `gradle-ide-starter`

6hundreds:
> Sounds like we should update gradle-ide-starter to use the working implementation that gradle profiler uses, and then gradle profiler should delegate to gradle-ide-starter

Well, it might be an option, but I see how using GProfiler might be beneficial in future for `smoke-ide-test`. For instance, for better/more comprehensive test coverage we may configure profiler to produce build operation trace of the sync, and then make assumptions on it.

PR #763, opened 2026-04-09
Support separate scenario option for benchmarking IntelliJ sync

6hundreds:
As now Profiler supports running IDEA and AS, we replacing all `studio` prefixes to `ide`-neutral ones, including in public APIs.

For preserving backward compatibility, old `studio-` handles are still operating as before, but emitting deprecation warnings now with mentioning to use `ide-` prefixed alternatives.

We hard change "public" API being used by `gradle/gradle` in testing setup, where GProfiler getting invoked programatically(attached as a library), without any deprecation cycle. Those API should be migrated on `gradle/gradle` side altogether with adopting the new version of the profiler. We don't expect(read "we don't care" ) any similar usages in wild, as the stability of that code was never promised.

It's better to review the PR on per-commit basis. 

This PR doesn't rename subprojects/packages. This is to be done as a follow up.

# UPDATE. 
We decided to go with separate scenario keys for different IDEs, as well as corresponding flags for each IDE. 

  ## Scenario blocks:                                                                                                                                                                                    
  - `intellij-idea-sync {}` — for IntelliJ IDEA sync
  - `android-studio-sync {}` — for Android Studio sync                                                                                                                                                                               
  - Nested keys inside the block are now strictly validated (`ide-jvm-args`, `studio-jvm-args` (deprecated), `idea-properties`). Nested keyset is same for both IDEs
  
  ## CLI flags:                                                                                                                                                                                    
  - `--idea-install-dir`, `--idea-sandbox-dir`, `--no-idea-sandbox`                                                                                                                                                                      
  - `--studio-install-dir`, `--studio-sandbox-dir`, `--no-studio-sandbox`                                                                                                                                                               
  - Default sandbox dirs: output-dir/idea-sandbox / output-dir/studio-sandbox when the matching install dir is provided.

## IDE executable verification
 GProfiler verifies that the supplied install dir actually contains the expected IDE's starter executable (idea/studio). Mismatch produces a clear error.
 
 ## Deprecations
- `studio-jvm-args` → use `ide-jvm-args`   
- `clear-android-studio-cache-before` → use `clear-ide-cache-before`


src/main/java/org/gradle/profiler/studio/invoker/IdeBuildInvocationResult.java
```diff
@@ -0,0 +1,56 @@
...
+        );
+    }
+
+    @Override
+    public IdeBuildActionResult getActionResult() {
+        return (IdeBuildActionResult) super.getActionResult();
+    }
+
+    public static final Sample<IdeBuildInvocationResult> GRADLE_TOTAL_EXECUTION_TIME
+        = new SingleInvocationDurationSample<IdeBuildInvocationResult>("Gradle total execution time") {
```
alllex: 💅 Here and below, it's unconventional to have the `=` assignement operator on another line

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -130,11 +144,13 @@ class ScenarioLoader {
             MAVEN,
             TOOLING_API,
             TOOL_HOME,
+            IDE_SYNC,
```
alllex: ❓ Should we add `IDE_JVM_ARGS` and `ANDROID_STUDIO_JVM_ARGS` to the list?

src/test/groovy/org/gradle/profiler/AbstractIdeSyncIntegrationTest.groovy
```diff
@@ -51,7 +51,7 @@ abstract class AbstractIdeSyncIntegrationTest extends AbstractProfilerIntegratio
         given:
         def scenarioFile = file("performance.scenarios") << """
             $scenarioName {
-                android-studio-sync {
+                ide-sync {
```
alllex: ❓ How does GProfiler understand whether to run IDEA or Studio?
Even if there is a rule, I don't think it's clear to the scenario file readers.

I think we should have two independent keys: `android-studio-sync` (as before) and `intellij-idea-sync` (new one). For now, they might expect exact same configuration, but in the future we might add flags that are specific to each IDE.

src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -535,20 +561,21 @@ public static GradleBuildInvoker invoker(Config config, GradleBuildInvoker defau
...
-                return GradleBuildInvoker.AndroidStudioCleanCacheBeforeScenario;
-            case BUILD:
-                return GradleBuildInvoker.AndroidStudioCleanCacheBeforeBuild;
-            case CLEANUP:
-            default:
-                throw new IllegalArgumentException(String.format("Unsupported cleanup schedule for '%s': '%s'", CLEAR_ANDROID_STUDIO_CACHE_BEFORE, schedule));
+        String clearCacheKey = resolveKeyWithDeprecatedFallback(config, CLEAR_IDE_CACHE_BEFORE, CLEAR_ANDROID_STUDIO_CACHE_BEFORE);
+        Schedule schedule = ConfigUtil.enumValue(config, clearCacheKey, Schedule.class, null);
+        if (schedule == null) {
+            return GradleBuildInvoker.AndroidStudio;
```
alllex: ❓ Why does this `GradleBuildInvoker` stay Android-specific, while the method is renamed to be generic?

src/test/groovy/org/gradle/profiler/AbstractIdeSyncIntegrationTest.groovy
asodja: Could we also update README.md? We use `android-studio-sync`. There are also some studio flags that needs to be updated.  

Also we should update min supported IDE version (maybe we need  a new paragraph just for IDE sync with IDE version info)

src/main/java/org/gradle/profiler/ScenarioLoader.java
6hundreds: It's an artifact after `git rebase` manipulation, to put related changes to the corresponding commits. Should be `Ide`

src/test/groovy/org/gradle/profiler/AbstractIdeSyncIntegrationTest.groovy
6hundreds: Actually, it doesn't. It depends on the file provided by `ide-path`.

src/main/java/org/gradle/profiler/ScenarioLoader.java
6hundreds: Interesting, ANDROID_STUDIO_JVM_ARGS wasn't on the list before

6hundreds: Ok, those are top-level keys, that get verified later. We don't validate nested keys. It shouldn't belong to this list, instead they should be added to a separate one, and be validated in the same manner. I don't think this is a scope of this PR.

README.md
```diff
@@ -285,6 +283,60 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+
+- `ide-jvm-args`: JVM arguments for the IDE process. Defaults to `["-Xms256m", "-Xmx4096m"]`.
+- `idea-properties`: Additional entries to write to the IDE's `idea.properties` file. Useful for setting registry values.
+
+To invalidate IDE caches between runs, use the scenario-level [`clear-ide-cache-before`](#cache-cleanup) key.
+
+For Android Studio, swap `--idea-install-dir` for `--studio-install-dir` and use the matching scenario block:
+
+    ideaSync {
+        intellij-idea-sync {
```
alllex: 💭 The naming of the block is inconsistent with the naming of CLI-options. It's also the only place, where the Profiler user would have to write "intellij".  I think consistency is important, so I'd align the naming. 

Between adding or dropping "intellij" for CLI options, I think the shorter variant is a clear winner. Thus, I think it's better to have the `idea-sync` as the key in the scenario file.

The same argument applies to the `android-studio-sync` key.

README.md
```diff
@@ -285,6 +283,60 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+            ide-jvm-args = ["-Xms256m", "-Xmx4096m"]
+        }
+    }
+
+    studioSync {
```
alllex: 💭 When the top-level name in the scenario file is generic, I personally have to do a couple of mental cycles to remember whether it's the one that has semantics only provides the name. It's the latter.

Can we spare the readers this mental exercise and make the user-defined scenario name obvious?

Something as simple as `myStudioSync` would go a long way.

README.md
```diff
@@ -285,6 +283,60 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+    }
+
+For Android Studio projects, make sure a `local.properties` file exists with `sdk.dir` pointing to your Android SDK.
+
+### Deprecated options
+
+The following options are still accepted but deprecated. A warning will be printed when they are used:
+
+| Deprecated | Replacement |
+|---|---|
```
alllex: 💅 Formatting

README.md
```diff
@@ -351,18 +403,8 @@ Here is an example:
...
-        # Note2: you need to have local.properties file in your project with sdk.dir set
-        android-studio-sync {
-            # Override default Android Studio jvm args
-            # studio-jvm-args = ["-Xms256m", "-Xmx4096m"]
-            # Pass an IDEA properties to Android Studio. This can be used to set a registry values as well
-            # idea-properties = ["gradle.tooling.models.parallel.fetch=true"]
-        }
-    }
+    # For IDE-sync scenarios (`intellij-idea-sync` / `android-studio-sync`),
+    # see [IDE sync support](#ide-sync-support).
```
alllex: 🤔 I don't think the markdown link syntax works here, and because of this it's a bit confusing

README.md
```diff
@@ -285,6 +283,60 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+
+- **IntelliJ IDEA** 2025.3 or newer
+- **Android Studio** Panda (2025.3) or newer
+
+Both IDEs are based on the IntelliJ platform. The profiler installs a lightweight plugin into the IDE that controls the sync lifecycle and communicates results back to the profiler.
+
+### Command-line options
+
+- `--idea-install-dir`: The IntelliJ IDEA installation directory. Required when measuring an `intellij-idea-sync` scenario. On macOS, it is the app directory, e.g. `~/Applications/IntelliJ IDEA.app`.
+- `--idea-sandbox-dir`: The IntelliJ IDEA sandbox directory. It's recommended to use it since it isolates the IDE process from your other IDE processes. By default, this is set to `<output-dir>/idea-sandbox`. If you want the IDE to keep old data (e.g. indexes), set and reuse your own folder.
```
alllex: 💅 Having a sentence per line is a bit easier to read and maintain in the source form

README.md
```diff
@@ -285,6 +283,60 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+- `--no-idea-sandbox`: Do not use the IntelliJ IDEA sandbox; instead use the default IDE folders for the IDE data.
+- `--studio-install-dir`: The Android Studio installation directory. Required when measuring an `android-studio-sync` scenario. On macOS, it is the app directory, e.g. `~/Applications/Android Studio.app`.
+- `--studio-sandbox-dir`: The Android Studio sandbox directory. By default, this is set to `<output-dir>/studio-sandbox`.
+- `--no-studio-sandbox`: Do not use the Android Studio sandbox; instead use the default IDE folders for the IDE data.
+
+### Scenario block keys
```
alllex:
```suggestion
### Scenario file blocks
```

README.md
```diff
@@ -285,6 +283,60 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+        }
+    }
+
+    studioSync {
+        android-studio-sync {
+            ide-jvm-args = ["-Xms256m", "-Xmx4096m"]
+        }
+    }
+
+For Android Studio projects, make sure a `local.properties` file exists with `sdk.dir` pointing to your Android SDK.
```
alllex: 💅 Can we turn this into a callout?

src/main/java/org/gradle/profiler/studio/IdeType.java
```diff
@@ -0,0 +1,79 @@
...
+    );
+
+    private final String displayName;
+    private final String vmOptionsEnvVar;
+    private final String propertiesEnvVar;
+    private final List<String> macOsStarterPaths;
+    private final List<String> windowsStarterPaths;
+    private final List<String> linuxStarterPaths;
+
+    IdeType(String displayName, String vmOptionsEnvVar, String propertiesEnvVar, List<String> macOsStarterPaths, List<String> windowsStarterPaths, List<String> linuxStarterPaths) {
```
alllex: 💅 Split into lines

src/main/java/org/gradle/profiler/studio/IdeType.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.common.collect.ImmutableList;
+import org.gradle.profiler.OperatingSystem;
+
+import java.util.List;
+
+/**
+ * Identifies which IntelliJ-based IDE is being profiled.
+ * Used to validate the install directory and locate the right starter executable.
+ */
+public enum IdeType {
```
alllex, line 12: 👍 Nice idea to introduce such an abstraction

alllex (approved):
LGTM!

Though I strongly suggest we make the scenario file block keys consistent in naming with the CLI options, so users only need to learn it once.

I think we should also update the PR title to reflect the design we are going with. Maybe make it shorter too 😄

src/main/java/org/gradle/profiler/DeprecationWarnings.java
```diff
@@ -0,0 +1,41 @@
+package org.gradle.profiler;
+
+import com.typesafe.config.Config;
+
+import java.util.List;
+
+public class DeprecationWarnings {
```
asodja, line 7: ❓ Do we need deprecations? In gradle-profiler we introduce a lot of breaking changes, so I think we could do some more and just not change keys

asodja (approved):
LGTM

README.md
```diff
@@ -285,6 +283,69 @@ see the [JVM requirements and option](#jvm-requirements-and-options) section for
...
+For Android Studio, swap `--idea-install-dir` for `--studio-install-dir` and use the matching scenario block:
+
+    ideaSync {
+        idea-sync {
+            ide-jvm-args = ["-Xms256m", "-Xmx4096m"]
+        }
+    }
+
+    studioSync {
+        android-studio-sync {
```
alllex, line 332: Now that the IDEA options and scenario block have consistent naming, we should do the same for `android-studio-sync -> studio-sync`.

PR #764, opened 2026-04-14
Rename `android-studio-sync` to `studio-sync`

6hundreds:
For consistency with AS control flags, prefixed with `studio-`


src/main/java/org/gradle/profiler/ScenarioLoader.java
```diff
@@ -415,13 +422,6 @@ private static IdeType getIdeSyncType(Config scenario) {
         return null;
     }
 
-    private static String getIdeSyncKey(IdeType ideType) {
-        return switch (ideType) {
-            case INTELLIJ_IDEA -> IDEA_SYNC;
-            case ANDROID_STUDIO -> ANDROID_STUDIO_SYNC;
-        };
-    }
```
6hundreds, line 423: Was inlined

alllex (approved):
👍

PR #765, opened 2026-04-14
Get rid of `studio` prefix in packages and subproject names

6hundreds:


subprojects/studio-plugin/src/main/kotlin/org/gradle/profiler/studio/plugin/GradleProfilerStartupActivity.kt
```diff
@@ -47,12 +47,12 @@ class GradleProfilerStartupActivity : ProjectActivity {
     }
 
     private fun logModifiedRegistryEntries() {
-        val studioPropertiesPath = System.getenv("STUDIO_PROPERTIES")
-        if (studioPropertiesPath == null || !File(studioPropertiesPath).exists()) {
```
6hundreds, line 51: It was a minor bug, added a test for this behavior above

src/main/java/org/gradle/profiler/ide/launcher/IdeLauncherProvider.java
```diff
@@ -15,7 +15,7 @@
 
 public class IdeLauncherProvider {
 
-    private static final boolean SHOULD_RUN_HEADLESS = Boolean.getBoolean("studio.tests.headless");
+    private static final boolean SHOULD_RUN_HEADLESS = Boolean.getBoolean("ide.tests.headless");
```
alllex, line 18: 💭 We would need to watch out for this rename, when upgrading Profiler in g/g

alllex (approved):

src/main/java/org/gradle/profiler/ide/launcher/IdeLauncherProvider.java
6hundreds: 100%

PR #766, opened 2026-04-15
Disable CC for publication

6hundreds:
A publication with CC attempt:
https://builds.gradle.org/buildConfiguration/GradleProfiler_GradleProfilerPublishing/111495039

GProfiler uses Nexus plugin, which is CC incompatible so far.
Also our own build logic wasn't CC compatible.

### DIFF
--- a/.teamcity/src/main/kotlin/GradleProfilerPublishing.kt
+++ b/.teamcity/src/main/kotlin/GradleProfilerPublishing.kt
@@ -33,7 +33,9 @@ object GradleProfilerPublishing : BuildType({
 
     steps {
         gradle {
-            tasks = "clean createBuildReceipt publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag %additional.gradle.parameters%"
+            // No CC since https://github.com/gradle-nexus/publish-plugin/issues/221, which is
+            // waiting for https://github.com/gradle/gradle/issues/22779
+            tasks = "clean createBuildReceipt publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag %additional.gradle.parameters% --no-configuration-cache"
             gradleParams = toolchainConfiguration(os, arch) + " -Dgradle.cache.remote.push=true"
             buildFile = ""
         }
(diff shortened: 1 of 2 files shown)


.teamcity/src/main/kotlin/GradleProfilerPublishing.kt
```diff
@@ -33,7 +33,9 @@ object GradleProfilerPublishing : BuildType({
 
     steps {
         gradle {
-            tasks = "clean createBuildReceipt publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag %additional.gradle.parameters%"
+            // No CC since https://github.com/gradle-nexus/publish-plugin/issues/221, which is
+            // waiting for https://github.com/gradle/gradle/issues/22779
+            tasks = "clean createBuildReceipt publishToSonatype closeAndReleaseSonatypeStagingRepository gitPushTag %additional.gradle.parameters% --no-configuration-cache"
```
alllex: 💅  We should move the flag at the beginning of the string. I think it's customary to put top-level build options before tasks. But also, we want the `%parameters%` to always go last allowing to override anything

buildSrc/src/main/kotlin/profiler.versioning-root-project.gradle.kts
```diff
@@ -21,8 +21,10 @@ val createdBuildReceipt = layout.buildDirectory.file(buildReceiptName)
 tasks.register("createBuildReceipt") {
     outputs.file(createdBuildReceipt).withPropertyName("buildReceipt")
     inputs.property("version", versionInfo.version)
+    val version = versionInfo.version
+    val buildReceipt = createdBuildReceipt
```
alllex, line 25: 💅  It's already a variable, why do we need to re-assign it?

alllex (approved):

buildSrc/src/main/kotlin/profiler.versioning-root-project.gradle.kts
6hundreds: It's a variable at the Gradle script object. We made it local here, to avoid capturing Gradle script at execution time

PR #793, opened 2026-06-02
Add JFR to Perfetto trace converter

hegyibalint:
> [!NOTE]
> Stacked on #808, which handles the Perfetto proto restructuring (single-file fetch + subproject split). Review and merge that one first; this PR's diff is just the JFR converter.

Introduce a converter that turns JFR recordings into Perfetto traces.

Adds the `perfetto-converter-jfr` subproject: a neutral core (`Converter`, `ConverterSession`, `ThreadIdentity`), JFR input helpers (`JfrRecordFields` and the `Jfr*` processors), and Perfetto output helpers (`PerfettoTrackEventBuilder`, `PerfettoTrackRegistry`, `PerfettoTraceEmitter`, `PerfettoCallstackInterner`, `PerfettoTraceWriter`). Processors cover Gradle build operations, CPU usage, GC, heap summaries, execution samples, thread lifetime and thread state.

The `gradle-trace-converter-app` CLI now dispatches to the JFR converter for `.jfr` inputs and to the build-operation converter otherwise.

PR #808, opened 2026-06-29
Fetch Perfetto proto as a single checksummed file and split proto subproject

hegyibalint:
Split off from #793 to keep that PR focused on the JFR converter itself.

This PR changes only how the Perfetto proto is managed:

- **Stop vendoring ~285 `.proto` files.** Instead, fetch a single `perfetto_trace.proto` from a pinned upstream Perfetto release and verify it by checksum at build time, via a reusable `FetchPerfettoProtoTask` in `buildSrc` (configurable by `version` and `sha256`, output dir as a convention).
- **Split the old `perfetto-trace` subproject in two:**
  - `perfetto-trace-proto`: the generated Perfetto protobuf classes only, sourced from the fetched single-file proto.
  - `perfetto-converter-buildops`: the build-operation to Perfetto converter, depending on `:perfetto-trace-proto`.
- The fetch task injects `java_multiple_files = true`, so generated classes are top-level (e.g. `perfetto.protos.TracePacket`); `ChromeTraceWriter` imports are adapted accordingly and its `tid` widened to match the generated long field.

#793 (the JFR converter) is stacked on top of this branch.


buildSrc/src/main/kotlin/tasks/FetchPerfettoProtoTask.kt
```diff
@@ -0,0 +1,61 @@
...
+        if (actualSha != expectedSha) {
+            throw GradleException(
+                "Checksum mismatch for $url\n  expected: $expectedSha\n  actual:   $actualSha"
+            )
+        }
+
+        // Upstream does not set java_multiple_files; the converter relies on the resulting top-level
+        // generated classes (e.g. perfetto.protos.TracePacket), so we inject it after verification.
+        val patched = String(bytes, Charsets.UTF_8)
+            .replaceFirst("package perfetto.protos;", "package perfetto.protos;\noption java_multiple_files = true;")
```
alllex, line 64: 🤖💅 This `replaceFirst` silently no-ops if the pattern `package perfetto.protos;` isn't found — e.g. if a future upstream release reformats or moves that line. In that case we'd write an *unpatched* proto (no `java_multiple_files`), codegen would emit the old `*OuterClass`-nested classes, and the failure would surface much later as a confusing compile error in `perfetto-converter-buildops` rather than here.

Could we fail loudly at fetch time instead? e.g. assert the replacement actually took effect:

```kotlin
val patched = String(bytes, Charsets.UTF_8)
    .replaceFirst("package perfetto.protos;", "package perfetto.protos;\noption java_multiple_files = true;")
if (!patched.contains("option java_multiple_files = true;")) {
    throw GradleException(
        "Could not inject 'java_multiple_files' into $url — the 'package perfetto.protos;' " +
        "declaration may have changed upstream. Update FetchPerfettoProtoTask accordingly."
    )
}
```

That way a bad upstream shape points straight at the cause.

buildSrc/src/main/kotlin/tasks/FetchPerfettoProtoTask.kt
```diff
@@ -0,0 +1,61 @@
...
+    abstract val outputDir: DirectoryProperty
+
+    @TaskAction
+    fun fetch() {
+        val version = version.get()
+        val expectedSha = sha256.get()
+        val url =
+            "https://raw.githubusercontent.com/google/perfetto/$version/protos/perfetto/trace/perfetto_trace.proto"
+
+        val bytes = URI.create(url).toURL().openStream().use { it.readBytes() }
```
alllex: 🤖💅 Minor/optional: `openStream()` here doesn't check the HTTP status or handle a cross-host redirect. On an HTTP error (or a redirect to an error page), we'd read whatever body comes back and then fail on the checksum comparison — producing a "Checksum mismatch" message that misleadingly points at the wrong problem when the real issue was that the download failed.

Not a correctness/integrity concern (the checksum still guards the bytes we actually use), so feel free to leave it. But if you're touching this file for the comment above, opening via `HttpURLConnection` and checking `responseCode == 200` before reading would give a clearer "download failed" error separate from a genuine checksum mismatch.

alllex (approved):
LGTM!

buildSrc/src/main/kotlin/tasks/FetchPerfettoProtoTask.kt
hegyibalint: Really good feedback, really good points. I'm using `HttpURLConnection` instead then.

buildSrc/src/main/kotlin/tasks/FetchPerfettoProtoTask.kt
hegyibalint: Implemented verbatim, really good point!

PR #809, opened 2026-06-30
Auto-detect headless environment for IDE sync

joshfriend:
Run the IDE in headless mode when `GraphicsEnvironment.isHeadless()` is true, so CI without a display no longer needs `-Dide.tests.headless=true`. The explicit property still forces headless on machines that have a display.

### DIFF
--- a/src/main/java/org/gradle/profiler/ide/launcher/IdeLauncherProvider.java
+++ b/src/main/java/org/gradle/profiler/ide/launcher/IdeLauncherProvider.java
@@ -5,6 +5,7 @@ import org.gradle.profiler.instrument.GradleInstrumentation;
 import org.gradle.profiler.ide.IdeType;
 import org.gradle.profiler.ide.tools.IdeSandboxCreator.IdeSandbox;
 
+import java.awt.GraphicsEnvironment;
 import java.io.File;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


src/main/java/org/gradle/profiler/ide/launcher/IdeLauncherProvider.java
```diff
@@ -15,7 +16,7 @@
 
 public class IdeLauncherProvider {
 
-    private static final boolean SHOULD_RUN_HEADLESS = Boolean.getBoolean("ide.tests.headless");
+    private static final boolean SHOULD_RUN_HEADLESS = Boolean.getBoolean("ide.tests.headless") || GraphicsEnvironment.isHeadless();
```
alllex, line 19: @6hundreds , can we manually validate that this still runs with UI on desktop?

alllex (approved):

src/main/java/org/gradle/profiler/ide/launcher/IdeLauncherProvider.java
6hundreds: It does

6hundreds (approved):

6hundreds:
@joshfriend Thanks for the contribution

PR #812, opened 2026-07-08
Wait for SDKMAN download URL before releasing, with a clear error

blindpirate:
## Problem

The `Publish to SDKMan` build [failed publishing 0.25.0](https://builds.gradle.org/buildConfiguration/GradleProfiler_GradleProfilerPublishToSdkMan/114878118) with an opaque error:

```
> Task :sdkReleaseVersion FAILED
Releasing gradleprofiler 0.25.0 for UNIVERSAL...
Response: 400: Bad Request...
> Response: 400 Bad Request
```

### Root cause

The SDKMAN vendor API (`POST https://vendors.sdkman.io/versions`) **downloads the release URL itself** and validates it. Its [`validateUrl`](https://github.com/sdkman/vendor-release/blob/master/src/main/scala/io/sdkman/vendor/release/routes/Validation.scala) returns `400 Bad Request` with body `"URL cannot be resolved: <url>"` when it can't fetch the artifact. For our payload (`platform=UNIVERSAL`, no checksums, plain version `0.25.0`) that URL check is the *only* validation that can 400 — and it did, because the `gradle-profiler-0.25.0.zip` had only just been published to Maven Central (06:28 UTC) and hadn't propagated when the release ran (06:29 UTC).

Two things made this painful:
1. The release fires before the artifact is reliably resolvable from Maven Central.
2. The `io.sdkman.vendors` plugin **hides the real message** — it drops the response body (`"URL cannot be resolved"`) and only surfaces `Response: 400 Bad Request`.

Note: re-running the SDKMAN publish for 0.25.0 does **not** need a version bump — SDKMAN only stores metadata pointing at the existing Central artifact; it never re-uploads. 0.25.0 is not registered on SDKMAN yet, so there's no conflict.

## Fix

- New `AwaitUrlTask` (buildSrc) polls the download URL with HEAD requests (10 min timeout, 15 s interval).
- `:sdkReleaseVersion` now `dependsOn` it, so the release only runs once the URL resolves.
- If the URL never resolves, the task **fails with a message naming the real cause** instead of a confusing 400:

  ```
  URL is not accessible after 10 minutes: <url>
  Last check failed with: HTTP 404.
  This is the real reason a SDKMAN release fails here: its API downloads this URL and
  rejects the release with 'HTTP 400 Bad Request (URL cannot be resolved)'. The artifact
  has most likely not propagated to Maven Central yet — re-run once it is available.
  ```

## Testing

- Configuration-cache compatible (verified `Configuration cache entry stored`, no CC problems).
- Task graph verified: `:awaitSdkmanArtifact` runs before `:sdkReleaseVersion`.
- Functionally exercised both paths:
  - Reachable URL (0.24.0 zip) → `URL is available (attempt 1)`, BUILD SUCCESSFUL.
  - Unreachable URL (nonexistent 9.9.9 zip) → fails with the clear message above.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@ import io.sdkman.vendors.tasks.SdkDefaultVersion
 import io.sdkman.vendors.tasks.SdkReleaseVersion
 import io.sdkman.vendors.tasks.SdkmanVendorBaseTask
 import java.util.Locale
+import tasks.AwaitUrlTask
 
 plugins {
     id("profiler.java-library")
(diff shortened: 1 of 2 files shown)


blindpirate:
As discussed, we'll move the distribution from maven central to GitHub releases so we can avoid the delay.

PR #813, opened 2026-07-08
fix: align wide-format CSV rows by phase and iteration so mixed warm-up scenarios stay parseable

### LINKED ISSUE #317: CSV is wrong when scenarios have different number of warm ups
Running scenarios with different number of warm-ups results in csv with wrong header and with missing commas, making it impossible to parse

Given these scenarios

```
clean_build {
    title = "Clean Build"
    tasks = ["assembleDebug"]
    cleanup-tasks = ["clean"]
    gradle-args = ["--rerun-tasks"]
    run-using = tooling-api
    daemon = cold
}

touch_sdk {
    title = "SDK"
    tasks = ["assembleDebug"]
    run-using = tooling-api
    daemon = warm
    apply-abi-change-to = [file_path]

    warm-ups = 2
}

touch_webservices {
    title = "WebServices"
    tasks = ["assembleDebug"]
    run-using = tooling-api
    daemon = warm
    apply-abi-change-to = [file_path]

    warm-ups = 2
}
```

And this is the command I use 

```
gradle-profiler --benchmark --measure-config-time --measure-gc --project-dir . --scenario-file performance.scenarios --iterations 2
```

This is the benchmark.csv file content

```csv
scenario,Clean Build,Clean Build,Clean Build,SDK,SDK,SDK,WebServices,WebServices,WebServices

value,execution,garbage collection time,task start,execution,garbage collection time,task start,execution,garbage collection time,task start

warm-up build #1,245182,12190,20707,129608,5800,21219,127184,6353,22276

measured build #1,253400,12265,23149,36249,1557,7786,33270,445,6929

measured build #2,258763,11649,21661,29886,478,6119,30580,712,6181

measured build #2,,29194,845,5686,29346,1534,5655
```

On the one hand, the label `measured build #1` should be `warm-up build #2` for `touch_sdk` and `touch_webservices` scenarios. Truth is, I don't know how to overcome this problem.

On the other hand, the second `measured build #2`  label has 7 values, while the other rows have 9

HTML shows correct rows & columns
image

Am I misusing the tool?

Thank you for your time

mvanhorn:
## Summary
Wide-format `benchmark.csv` now stays parseable and correctly labeled when scenarios in one run have different warm-up or iteration counts. Rows are keyed by `(phase, iteration)` instead of a raw result index, so a scenario's `measured build #1` never lands on another scenario's `warm-up build #2` row, and every row carries the same column count as the header.

## Why this matters
The reporter mixed scenarios with different warm-up counts and got a CSV where rows had 7 values under a 9-column header, which breaks any downstream parser. Eleven comments on #317 (including two Gradle members) confirm the wide format is the culprit; the HTML report was always fine because it serializes per-scenario JSON. `CsvGenerator.writeWide` iterated one shared row index across `scenario.getResults()`, so scenarios that ran out of results emitted a single comma regardless of how many sample columns they own.

## Changes
- The row axis is now built from per-scenario warm-up and measure counts (max across scenarios), emitting `WARM_UP #1..N` rows then `MEASURE #1..M` rows labeled via `phase.displayBuildNumber(iteration)`.
- Each scenario looks up its result by build context match; a scenario with no result for a row pads exactly `scenario.getSamples().size()` empty cells. This replaces the first-scenario `break` label hack and the single-comma early return in `writeWideRow`.
- `writeLong`, the JSON/HTML paths, and `CsvGenerator`'s public API are untouched, so no call sites change.

## Testing
New `CsvGeneratorTest.groovy` covers mixed warm-up counts (labels align per phase), uneven iteration counts (all rows match header width), and the unchanged single-scenario output.

Fixes #317

### DIFF
--- a/src/test/groovy/org/gradle/profiler/report/CsvGeneratorTest.groovy
+++ b/src/test/groovy/org/gradle/profiler/report/CsvGeneratorTest.groovy
@@ -1,0 +1,267 @@
+package org.gradle.profiler.report
+
+import org.gradle.profiler.BuildAction
+import org.gradle.profiler.BuildContext
+import org.gradle.profiler.BuildScenarioResultImpl
+import org.gradle.profiler.GradleBuildConfiguration
+import org.gradle.profiler.Phase
+import org.gradle.profiler.ScenarioContext
+import org.gradle.profiler.gradle.GradleBuildInvoker
+import org.gradle.profiler.gradle.GradleScenarioDefinition
+import org.gradle.profiler.gradle.RunTasksAction
+import org.gradle.profiler.result.BuildActionResult
+import org.gradle.profiler.result.BuildInvocationResult
+import org.gradle.profiler.result.Sample
+import org.gradle.profiler.result.SingleInvocationDurationSample
+import org.gradle.util.GradleVersion
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.Specification
+
+import java.time.Duration
+
+class CsvGeneratorTest extends Specification {
+
+    @Rule TemporaryFolder tmpDir
+
+    int scenarioCounter
+
+    def setup() {
+        scenarioCounter = 0
+    }
+
+    def "wide format keeps existing output when scenarios have the same warmups and iterations"() {
+        def result1 = scenarioResult(
+            scenario("release", "Assemble Release", ":assemble", 2, 2),
+            [BuildInvocationResult.EXECUTION_TIME],
+            [
+                [Phase.WARM_UP, 1, 100],
+                [Phase.WARM_UP, 2, 90],
+                [Phase.MEASURE, 1, 80],
+                [Phase.MEASURE, 2, 70],
+            ]
+        )
+        def result2 = scenarioResult(
+            scenario("debug", "Assemble Debug", ":assembleDebug", 2, 2),
+            [BuildInvocationResult.EXECUTION_TIME],
+            [
+                [Phase.WARM_UP, 1, 110],
+                [Phase.WARM_UP, 2, 95],
+                [Phase.MEASURE, 1, 85],
+                [Phase.MEASURE, 2, 75],
+            ]
+        )
+
+        expect:
+        writeCsv(Format.WIDE, [result1, result2]) == """scenario,Assemble Release,Assemble Debug
+version,Gradle 6.7,Gradle 6.7
+tasks,:assemble,:assembleDebug
+value,total execution time,total execution time
+warm-up build #1,100.00,110.00
+warm-up build #2,90.00,95.00
+measured build #1,80.00,85.00
+measured build #2,70.00,75.00
+"""
+    }
+
+    def "wide format aligns measured builds when scenarios have different warmup counts"() {
+        def shortWarmup = scenarioResult(
+            scenario("short-warmup", "Short Warmup", ":short", 1, 2),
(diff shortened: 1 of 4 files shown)


ljacomet:
Thank you for your contribution!

The [DCO](https://probot.github.io/apps/dco/) check is failing, please look into that as we can't accept the PR otherwise.

ghale:

This PR has a valid DCO and tests. The relevant team for this area will confirm the implementation choices.

src/test/groovy/org/gradle/profiler/report/CsvGeneratorTest.groovy
```diff
@@ -0,0 +1,267 @@
...
+
+    private static BuildScenarioResultImpl<BuildInvocationResult> scenarioResult(
+        GradleScenarioDefinition scenario,
+        List<Sample<? super BuildInvocationResult>> samples,
+        List<List> builds
+    ) {
+        def result = new BuildScenarioResultImpl<BuildInvocationResult>(scenario, { samples })
+        def context = new TestScenarioContext(scenario.name)
+        builds.each { build ->
+            long testTime = build.size() > 3 ? build[3] as long : build[2] as long
```
6hundreds: The positional build rows are a bit opaque, particularly this fallback:

```groovy
long testTime = build.size() > 3 ? build[3] as long : build[2] as long
```

Reading `[Phase.WARM_UP, 1, 200]` at a call site, it's not obvious that the test sample value silently becomes a copy of the execution time. It also means accidentally dropping a 4th value from a row that should have one doesn't fail — the test just quietly asserts against the wrong number.

Consider named maps per row, which makes the fallback unnecessary:

```groovy
[phase: Phase.WARM_UP, iteration: 1, execution: 100, test: 120],
[phase: Phase.MEASURE, iteration: 1, execution: 80],  // no test sample
```

```groovy
builds.each { build ->
    def context = context.withBuild(build.phase, build.iteration)
    result.accept(new TestInvocationResult(context, build.execution, build.test ?: build.execution))
}
```

Or simpler: always require 4 values — the test value is ignored anyway for scenarios that don't include `TestSample`.

src/test/groovy/org/gradle/profiler/report/CsvGeneratorTest.groovy
```diff
@@ -0,0 +1,267 @@
...
+            result.accept(new TestInvocationResult(context.withBuild(build[0] as Phase, build[1] as int), build[2] as long, testTime))
+        }
+        result
+    }
+
+    static class TestSample extends SingleInvocationDurationSample<BuildInvocationResult> {
```
6hundreds: PTAL at `JsonResultWriterTest` - it contains same fixtures. Could you extract those and reuse?

6hundreds (changes requested):
Good stuff! Left a few comments, PTAL

6hundreds (approved):
Let's fix the conflicts and we're good to go

mvanhorn:
The conflicts were already resolved in the 18:45 and 18:50 pushes, so the PR has been mergeable since then. The remaining red was DCO: the second commit was missing its `Signed-off-by` trailer.

Force-pushed 4956b9f, which adds the trailer and tidies that commit message (the body had absolute local paths and stray tool output in it). The tree is byte-identical to 5163307, so no code changed. DCO is green now.

6hundreds:
@mvanhorn Thank you for your contribution to Gradle Profiler, I proceed with merge this PR

PR #819, opened 2026-07-09
Publish CLI distribution ZIP to GitHub Releases instead of Maven Central

### LINKED ISSUE #817: Improve Gradle profiler publication
Currently, Gradle Profiler publication is using Maven Central, including for the distribution of the application archive.

With upcoming Maven Central publishing limits, this is not the best use of that shared resource.

We need to reduce the size of the Maven Central publication, knowing that the following 3 files are the larger ones:
* `gradle-profiler-0.25.0.zip`               33MB
* `perfetto-trace-0.25.0-javadoc.jar` 18MB
* `perfetto-trace-0.25.0.jar`                12MB

Proposed change:
* Stop publishing the application archive to Maven Central and instead attach it to the GitHub release
  * As a side effect, this would resolve some Maven Central vs. SDKMan visibility challenges we've had recently

blindpirate:
Closes https://github.com/gradle/gradle-profiler/issues/817

## Why

The gradle-profiler **CLI distribution ZIP** (`gradle-profiler-$version.zip`) is published to Maven Central purely to give SDKMAN a stable, unauthenticated download URL. This large binary is driving a Maven Central quota problem.

This PR moves *only the distribution ZIP* to GitHub Releases and points SDKMAN at the release asset. The **library JARs** (subprojects + the main `gradle-profiler` jar/sources/javadoc) keep publishing to Maven Central unchanged — gradle/gradle consumes Gradle-Profiler-as-a-library from there.

## What changed

- **`build.gradle.kts`**
  - Removed the ZIP from the `mavenJava` Maven publication (and the now-unused `archives` artifact).
  - Added a `publishToGithubReleases` task (plain GitHub REST calls via `java.net.http.HttpClient` + `groovy.json.JsonSlurper`, no new plugins/deps). It finds the existing release-drafter **draft** for `v$version`, (idempotently) deletes any stale same-named asset, uploads the `distZip`, then flips the release to published (`draft: false`, `make_latest: true`). Guarded by `onlyIf` to skip snapshots/alphas; `dependsOn(distZip)`, `mustRunAfter(gitPushTag)`. Fails loudly on a missing draft or any non-2xx response.
  - Pointed the SDKMAN `url` at `https://github.com/gradle/gradle-profiler/releases/download/v$version/gradle-profiler-$version.zip`.
- **`.teamcity/.../GradleProfilerPublishing.kt`** — run `publishToGithubReleases` in the publishing build (after `gitPushTag`).
- **`.teamcity/.../GradleProfilerPublishToSdkMan.kt`** — dropped the 30-minute `retryBuild` that waited for Maven Central sync; the GitHub asset is public immediately.
- **`.github/release-drafter.yml`** — prebuilt-download link now points at the GitHub release asset.

## Design note

CI publishes the *existing* release-drafter draft (uploads the asset + flips draft→published), preserving the curated release notes. GitHub draft-release assets aren't reachable at the public `releases/download/...` URL, so the release must be published (non-draft) with the ZIP attached before `releaseToSdkMan` runs.

## Verification

Verified in https://builds.gradle.org/buildConfiguration/GradleProfiler_GradleProfilerPublishToSdkMan/115000529?buildTab=log

- `./gradlew publishToGithubReleases --dry-run -Pprofiler.version=test` → task graph wires `distZip` → `publishToGithubReleases`.
- `./gradlew publishMavenJavaPublicationToMavenLocal -Pprofiler.version=test` → local repo has `.jar`, `-sources.jar`, `-javadoc.jar`, `.pom`, `.module` and **no `.zip`**; pom/module contain no zip references.
- `mvn -f .teamcity/pom.xml compile` → succeeds.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@ import io.sdkman.vendors.tasks.SdkDefaultVersion
 import io.sdkman.vendors.tasks.SdkReleaseVersion
 import io.sdkman.vendors.tasks.SdkmanVendorBaseTask
 import java.util.Locale
+import services.GithubReleaseService
 
 plugins {
     id("profiler.java-library")
@@ -217,22 +218,6 @@ tasks.register("testHtmlReports") {
     dependsOn(testReports.keys)
 }
 
-val profilerDistribution = artifacts.add("archives", tasks.distZip.flatMap { it.archiveFile }) {
-    type = "zip"
-}
-
-publishing {
-    publications {
-        named<MavenPublication>("mavenJava") {
-            artifact(profilerDistribution)
-            pom {
-                // For some reason adding the zip artifact changes the packaging to "pom"
-                packaging = "jar"
-            }
-        }
-    }
-}
-
 nexusPublishing {
     packageGroup.set(project.group.toString())
     repositories {
@@ -264,12 +249,37 @@ val gitPushTag = tasks.register<Exec>("gitPushTag") {
 
 fun Project.isSnapshot() = version.toString().endsWith("-SNAPSHOT")
 
+gradle.sharedServices.registerIfAbsent("githubRelease", GithubReleaseService::class) {
+    parameters.githubToken = providers.gradleProperty("githubToken")
+}
+
+tasks.register<tasks.PublishToGithubReleasesTask>("publishToGithubReleases") {
+    mustRunAfter(gitPushTag)
+
+    val versionString = project.version.toString()
+    repository = "gradle/gradle-profiler"
+    releaseVersion = versionString
+    // Wiring the ZIP file provider also carries the task dependency on distZip.
+    // The GithubReleaseService is auto-wired via @ServiceReference on the task.
+    distributionZip = tasks.distZip.flatMap { it.archiveFile }
+
+    // Mirror the alpha/snapshot check in releaseToSdkMan: alphas/snapshots have no drafter
+    // draft and SDKMAN skips them, so they get no GitHub release ZIP.
+    onlyIf {
+        !versionString.lowercase(Locale.US).run { contains("snapshot") || contains("alpha") }
+    }
+}
+
 sdkman {
+    // Use a local val for the version: inside this block `version` refers to the extension's
+    // Property<String>, so interpolating "$version" in the URL would embed the property's
+    // toString() ("extension 'sdkman' property 'version'") instead of the value.
+    val profilerVersion = project.version.toString()
     api = "https://vendors.sdkman.io"
     candidate = "gradleprofiler"
     hashtag = "#gradleprofiler"
-    version = project.version.toString()
-    url = "https://repo1.maven.org/maven2/org/gradle/profiler/gradle-profiler/$version/gradle-profiler-$version.zip"
+    version = profilerVersion
+    url = "https://github.com/gradle/gradle-profiler/releases/download/v$profilerVersion/gradle-profiler-$profilerVersion.zip"
     consumerKey = project.findProperty("sdkmanKey") as String?
     consumerToken = project.findProperty("sdkmanToken") as String?
 }
--- a/buildSrc/src/main/kotlin/tasks/PublishToGithubReleasesTask.kt
+++ b/buildSrc/src/main/kotlin/tasks/PublishToGithubReleasesTask.kt
@@ -1,0 +1,54 @@
+package tasks
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.services.ServiceReference
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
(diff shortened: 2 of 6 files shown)


buildSrc/src/main/kotlin/tasks/PublishToGithubReleasesTask.kt
```diff
@@ -0,0 +1,114 @@
...
+    @TaskAction
+    fun publish() {
+        val repo = repository.get()
+        val version = releaseVersion.get()
+        val token = githubToken.orNull
+            ?: throw GradleException("The 'githubToken' property is required to publish to GitHub Releases.")
+        val tag = "v$version"
+        val assetName = "gradle-profiler-$version.zip"
+        val apiBase = "https://api.github.com/repos/$repo"
+        val client = HttpClient.newHttpClient()
```
6hundreds: 💅 What I would do is to extract `GithubReleaseBuildService` that will encapsulated all the http client burden within, and then let the task use it

build.gradle.kts
```diff
@@ -264,12 +248,29 @@ val gitPushTag = tasks.register<Exec>("gitPushTag") {
 
 fun Project.isSnapshot() = version.toString().endsWith("-SNAPSHOT")
 
+tasks.register<tasks.PublishToGithubReleasesTask>("publishToGithubReleases") {
+    dependsOn(tasks.distZip)
```
6hundreds: ❓ Do we need it tho? we have line 258 that establishes the dependency

build.gradle.kts
```diff
@@ -264,12 +248,29 @@ val gitPushTag = tasks.register<Exec>("gitPushTag") {
...
+    mustRunAfter(gitPushTag)
+
+    val versionString = project.version.toString()
+    repository.set("gradle/gradle-profiler")
```
6hundreds: 💅 Can we use an assignment, instead of `set`?

6hundreds: Here and below

build.gradle.kts
```diff
@@ -264,12 +248,29 @@ val gitPushTag = tasks.register<Exec>("gitPushTag") {
...
+    releaseVersion.set(versionString)
+    distributionZip.set(tasks.distZip.flatMap { it.archiveFile })
+    project.findProperty("githubToken")?.toString()?.let { githubToken.set(it) }
```
6hundreds: ❓ Shouldn't an absence of the gh token be a hard failure?

6hundreds: ❓ Also, I'm not sure, but isn't `providers.gradleProperty` is more idiomatic way to obtain a property?

6hundreds (changes requested):
Left some comments

buildSrc/src/main/kotlin/tasks/PublishToGithubReleasesTask.kt
blindpirate: Done in 91834ddb — extracted `GithubReleaseService` (a `BuildService`) that owns the `HttpClient` and all the REST calls; the task now just calls `publishReleaseAsset(...)`.

build.gradle.kts
blindpirate: Good catch — removed the explicit `dependsOn(tasks.distZip)` in 91834ddb; wiring `distributionZip` to the distZip `archiveFile` provider already carries the task dependency (confirmed via `--dry-run`).

build.gradle.kts
blindpirate: Switched all of these to `=` property assignment in 91834ddb.

build.gradle.kts
blindpirate: Now uses `providers.gradleProperty("githubToken")` in 91834ddb. A missing token is a hard failure when the task actually runs — the service throws `GradleException`; it's guarded by `onlyIf` so alphas/snapshots (which don't publish) never hit it.

build.gradle.kts
```diff
@@ -264,12 +249,34 @@ val gitPushTag = tasks.register<Exec>("gitPushTag") {
...
+tasks.register<tasks.PublishToGithubReleasesTask>("publishToGithubReleases") {
+    mustRunAfter(gitPushTag)
+
+    val versionString = project.version.toString()
+    repository = "gradle/gradle-profiler"
+    releaseVersion = versionString
+    // Wiring the ZIP file provider also carries the task dependency on distZip.
+    distributionZip = tasks.distZip.flatMap { it.archiveFile }
+    githubReleaseService = githubRelease
+    usesService(githubRelease)
```
6hundreds: 💅 If we'll add a `ServiceReference` to a service field of the task, this line will be redundant. I think it's more idiomatic authoring

6hundreds (approved):
LGTM with small comment

build.gradle.kts
blindpirate: Done in 57c8c117 — switched to `@ServiceReference("githubRelease")` on the task's service property, so Gradle auto-injects the service and registers `usesService`. Dropped the explicit assignment + `usesService(...)` lines.

PR #824, opened 2026-07-22
Make Kotlin mutations compatible with the explicit ABI mode

### LINKED ISSUE #504: Kotlin ABI change fails when `explicit-api=strict`
I have a kotlin subproject with [Explicit API](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors) enabled. When I use `apply-abi-change-to`. The compilation fails because the function that is generated doesn't have the `public` modifier. Would it be possible to add it always?

vRallev:
ABI changes in Kotlin files don't add a `public` visibility modifier. This fails to compile, if Kotlin's explicit ABI mode is enabled:
```
Visibility must be specified in explicit API mode.
```

Adding the `public` modifier by default is a safe change. If the explicit ABI mode is enabled, then the code now successfully compiles. If the explicit ABI mode is disabled, then the optional modifier doesn't produce any warning. Only the IDE would highlight it as redundant.

Fixes #504


ghale:
This issue needs a decision from the team responsible for that area. They have been informed. Response time may vary.

alllex:
Sorry for the delayed response. The change makes sense. We'll merge it after the CI comes back green

PR #828, opened 2026-07-27
Fix task start measurement under configuration cache reuse

### LINKED ISSUE #739: "task start" timer not reset between builds in 0.24.0
I'm running a scenario like this:
```
build-warm {
    title = "build-warm"
    tasks = ["app:buildInternalDebug"]
    warm-ups = 4
    iterations = 5
}
```

After upgrading the profiler version from 0.23.0 to 0.24.0, the `task start` metric seems to accumulate the total execution time (or something along these lines) and isn't reset between runs. The `total execution time` itself is not affected though.

Example results:
_0.23.0_ (median `task start` is 0.792 s, as expected)
Image
[benchmark-0.23.0.csv](https://github.com/user-attachments/files/25941380/benchmark-0.23.0.csv)
[benchmark-0.23.0.html](https://github.com/user-attachments/files/25941381/benchmark-0.23.0.html)

_0.24.0_ (median `task start` is 36.383 s)
Image
[benchmark-0.24.0.csv](https://github.com/user-attachments/files/25941406/benchmark-0.24.0.csv)
[benchmark-0.24.0.html](https://github.com/user-attachments/files/25941405/benchmark-0.24.0.html)

The only difference between these runs is the version of the profiler.

This seems to impact only scenarios that don't do any mutations. For example, a scenario like this is not impacted:
```
build-non-abi-change-incremental {
    title = "build-non-abi-change-incremental"
    tasks = ["app:buildInternalDebug"]

    apply-non-abi-change-to = "path/to/file.kt"

    warm-ups = 4
    iterations = 5
}
```

joshfriend:
The build start time captured when the build service is registered is restored from the configuration cache snapshot on reuse, so it belongs to the build that created the cache entry. This made the task start measurement grow by roughly one build duration per invocation.

Prefer the build start time carried by the operation details, which is created fresh for every invocation. Falls back to the captured value for operation types whose details do not expose a build start time.

Fixes #739

### DIFF
--- a/subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
+++ b/subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
@@ -7,14 +7,12 @@ import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.internal.GradleInternal;
 import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.ListProperty;
-import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.provider.ProviderFactory;
 import org.gradle.api.services.BuildService;
 import org.gradle.api.services.BuildServiceParameters;
 import org.gradle.api.services.BuildServiceRegistry;
 import org.gradle.execution.RunRootBuildWorkBuildOperationType;
-import org.gradle.initialization.BuildRequestMetaData;
 import org.gradle.internal.build.event.BuildEventListenerRegistryInternal;
 import org.gradle.internal.file.PathToFileResolver;
 import org.gradle.internal.operations.BuildOperationDescriptor;
@@ -40,8 +38,11 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.OptionalLong;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -97,7 +98,6 @@ public class BuildOperationTrace {
                     BuildOperationMeasurementKind.TIME_TO_FIRST_EXCLUSIVE
                 )
             );
-            spec.getParameters().getBuildStartTime().set(gradle.getServices().get(BuildRequestMetaData.class).getStartTime());
         });
         registry.onOperationCompletion(listenerProvider);
         return this;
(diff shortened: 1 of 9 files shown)


ghale:

This PR has a valid DCO and tests. The relevant team for this area will confirm the implementation choices.

alllex:
This change makes sense, and we'll proceed with the review when the time permits.

cc @octylFractal, since you worked in this area relatively recently

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -249,17 +246,34 @@ public void finished(BuildOperationDescriptor buildOperationDescriptor, Operatio
...
+         * Prefer the build start time observed from the root build work operation over the one captured when
+         * the build service was registered. The registration only runs when the init script is executed, so
+         * with configuration cache reuse the captured value is restored from the cache snapshot and belongs
+         * to the build that created the cache entry. Operation details are created freshly for every
+         * invocation, so {@link RunRootBuildWorkBuildOperationType.Details#getBuildStartTime()} always
+         * reflects the current build request.
+         */
+        private long buildStartTime() {
+            long observed = observedBuildStartTime.get();
+            return observed != Long.MIN_VALUE ? observed : getParameters().getBuildStartTime().get();
```
octylFractal: I don't think this fallback is needed, as I think we only reach it on config failure or if there was no work to perform. Can we instead skip writing the measurement files if this value is absent and necessary for the measurement (e.g. using OptionalLong to pass it down and checking in each op type).

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -212,6 +212,7 @@ interface Params extends BuildServiceParameters {
         }
 
         private final List<BuildOperationCollector> collectors;
+        private final AtomicLong observedBuildStartTime = new AtomicLong(Long.MIN_VALUE);
```
octylFractal: I think it might be better to explicitly store an `AtomicReference<OptionalLong>` here, so if we ever get back `Long.MIN_VALUE` for some reason we don't think it's an absent value.

octylFractal (changes requested):

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
```diff
@@ -276,14 +284,16 @@ private BuildOperationCollector(Class<?> detailsType, Path outputFile, BuildOper
...
                 buildOperationCount.incrementAndGet();
             }
         }
 
-        public void write() throws IOException {
-            Duration value = measurer.computeFinalValue();
-            writeToFile(outputFile, value.toMillis(), buildOperationCount.get());
+        public void write(OptionalLong buildStartTime) throws IOException {
+            Optional<Duration> value = measurer.computeFinalValue(buildStartTime);
+            if (value.isPresent()) {
```
octylFractal, line 295: This won't work as-is, as this will result in a file from a previous run being read instead. We need to truncate here.

6hundreds:
Possibly a fix for https://github.com/gradle/gradle-profiler/issues/739 ?

subprojects/build-operations/src/main/java/org/gradle/trace/buildops/BuildOperationTrace.java
octylFractal: Alternatively, we might want to truncate or delete all files on the profiler side, to ensure stale results are never used.

joshfriend: I'm now truncating the file from the daemon side and added some new tests for this case

octylFractal (approved):
LGTM, thank you!

PR #837, opened 2026-08-06
Include computed statistics in the benchmark report JSON

### LINKED ISSUE #836: Include computed statistics (mean, percentiles, stddev) in benchmark report JSON
## Context

#698 added a standalone `benchmark.json` so external tooling can parse benchmark results. The JSON contains raw per-iteration samples only; the statistics shown by the HTML report (mean, min, P25, median, P75, max, std.dev) are computed client-side in `report.js`, so JSON consumers have to reimplement them.

## Proposal

Compute the statistics in Java and include them in the report data. Per scenario, per sample, add a `stats` object over MEASURE-phase iterations:

```json
"stats": {
  "mean": 46.3, "min": 44.1, "p25": 45.0, "median": 46.0,
  "p75": 47.2, "max": 51.8, "stddev": 1.9
}
```

Since the HTML report embeds the same JSON (`JsonResultWriter`), `report.js` can consume the precomputed values — a single source of truth for both outputs. Computations should match `report.js` exactly (R-7 quantile interpolation, population standard deviation) so both reports show the same numbers.

The `stats` key is additive, so the change is backward compatible with the schema introduced in #698.

## Out of scope: confidence

An earlier version of this proposal also included the Mann-Whitney confidence vs. a baseline scenario. **We decided not to include confidence in the JSON**:

- Any fixed baseline pairing is arbitrary for a file format with no interactivity (the HTML lets the user pick any baseline; several pairing rules were considered, each with real objections).
- The formula itself is not standardized: the HTML report uses a one-sided, tie-corrected value while gradle/gradle's perf-test infrastructure uses `1 − two-sided p` without tie correction (`DataSeries.confidenceInDifference`) — same data, different numbers.
- Precedent: per-sample confidence existed in the JSON in 2020 and was removed in c3e651da when the interactive HTML confidence replaced it.

Raw iteration values are in the report, so consumers can compute any comparison with any formula they prefer. A cross-checked Java implementation of the HTML report's formula is preserved at gradle-profiler@1516f25d should this ever be revisited.

6hundreds:
Closes #836

## What

Adds computed statistics to the benchmark report JSON introduced in #698, so external tooling no longer has to reimplement them. Each sample now carries a `stats` object — `mean`, `min`, `p25`, `median`, `p75`, `max`, `stddev` — computed over the MEASURE-phase iterations:

```json
"samples": [
  {
    "name": "total execution time",
    "unit": "ms",
    "stats": { "mean": 71.25, "min": 68.0, "p25": 69.5, "median": 71.0, "p75": 72.75, "max": 75.0, "stddev": 2.586 }
  }
]
```

The HTML report's statistics table now displays these precomputed values (with a fallback to local computation for reports produced by older versions), and no longer breaks when a scenario has no measured iterations (e.g. when it failed during a warm-up).

## Confidence: deferred

Earlier revisions of this PR also emitted a Mann-Whitney `confidence` per sample. It was removed from the scope after review discussion:

- The **pairing rule** is contested (always vs. first scenario? only across build tools, mirroring the HTML default? only between scenarios running the same tasks, like the original 2020 implementation did by scenario name?).
- The **formula** differs between the HTML report (one-sided, tie-corrected) and gradle/gradle's perf-test infrastructure (`DataSeries.confidenceInDifference`: `1 − two-sided p`, no tie correction) — standardization is an open question on #836.
- Precedent: per-sample confidence existed in the JSON in 2020 and was removed in c3e651da when the interactive HTML confidence replaced it. The `example-*.json` fixtures carried the stale field until this PR.

Raw iteration values remain in the report, so consumers can compute any comparison they need. The cross-checked implementation (verified digit-for-digit against the JS libraries used by `report.js`) is preserved in this branch's history at 1516f25d and can be wired back in once pairing and formula are settled.

## Design notes

- `stats` computations intentionally match `report.js` (R-7 quantile interpolation, population standard deviation), so the HTML report and the JSON always show the same numbers.
- The new key is additive, so the change is backward compatible with the schema locked in #698.

## Also in this PR

- Extracted `ScenarioResultWriter` from `JsonResultWriter`.
- Migrated `JsonResultWriterTest` to Spock snapshot testing (same mechanism as `JsonReportIntegrationTest`; regenerate with `./gradlew test -PupdateSnapshots=true`).
- New integration test covering the multi-scenario report shape end-to-end.
- `example-regression.json` now carries real `stats`, so the `testHtmlReportRegression` preview exercises the precomputed path (previously only the fallback path was ever rendered).
- Fixed a pre-existing configuration cache incompatibility in the `testHtmlReport*` preview tasks (separate commit).

## Verification

- `SampleStatisticsTest` covers the statistics including single-value and empty samples.
- Rendered the HTML report headlessly with stats-injected data to confirm the table displays precomputed values, and with stats-less fixture data to confirm the fallback path.


src/main/java/org/gradle/profiler/report/JsonResultWriter.java
```diff
@@ -58,70 +63,13 @@ public Environment() {
...
-        List<Sample<? super T>> samples = scenarioResult.getSamples();
-        for (Sample<? super T> sample : samples) {
-            samplesJson.add(serializeSample(sample));
-        }
-        json.add("samples", samplesJson);
-        JsonArray iterationsJson = new JsonArray();
-        for (T result : results) {
-            iterationsJson.add(serializeIteration(result, samples));
+    private static void writeScenarioResults(List<? extends BuildScenarioResult<?>> results, Gson gson, JsonObject json) {
+        BuildScenarioResult<?> baseline = results.size() > 1 ? results.get(0) : null;
```
asodja: ❓ So according to 🤖 this chooses the first scenarion as a baseline and thus we compare unrelated scenarios, e.g.: the new integration test compares `assemble` against `help`.

The HTML report avoids this: `report.js:182` only sets a baseline when `multipleBuildToolsPresent`, so a plain multi-scenario run shows no confidence column there while the JSON now reports one.

Should we change this?

src/main/js/org/gradle/profiler/report/report.js
```diff
@@ -169,8 +169,11 @@ new Vue({
                 sample.color = `hsl(${scenarioIndex * 360 / scenarios.length}, ${100 - 80 * sampleIndex / samples.length}%, ${30 + 40 * sampleIndex / samples.length}%)`;
                 sample.thickness = sampleIndex === 0 ? 3 : 2;
                 sample.selected = sampleIndex === 0;
+                // Use the statistics computed by the profiler when present, fall back to computing them for reports produced by older versions
                 const data = measuredIterations(scenario).map(iteration => iteration.values[sample.name]);
-                OPERATIONS.forEach(operation => sample[operation.name] = operation.apply(data));
+                OPERATIONS.forEach(operation => sample[operation.name] = sample.stats && sample.stats[operation.name] !== undefined
```
asodja, line 175: 💅 🤖 If a scenario has no measured iterations, the Java side leaves `stats` out, so we end up here computing from an empty array — and `mean()` on line 18 throws on an empty array. Because this runs in `beforeCreate`, the whole report comes out blank rather than just that one row.

It's reachable: a scenario that fails during its first warm-up still gets written to the report (Main.java:164). This is pre-existing, not something the PR broke, but since we're already editing this line: `data.length ? operation.apply(data) : null`.

asodja (changes requested):

build.gradle.kts
```diff
@@ -208,13 +208,13 @@ testReports.forEach { (taskName, fileName) ->
         from("src/main/resources/org/gradle/profiler/report")
         into(layout.buildDirectory.dir("test-html-report"))
         rename("report-template.html", "test-report-${fileName}.html")
+        val scriptFile = File(tasks.processResources.get().destinationDir, "org/gradle/profiler/report/report.js")
```
asodja, line 211: 💅 🤖 None of the example*.json files have `stats` in them, so these tasks only ever render the fallback path of the report.js change. The new path — the one that now fills the stats table and the diff columns in report-template.html:223-225: isn't covered anywhere. Adding `stats` to one example file would exercise it.

src/main/java/org/gradle/profiler/report/SampleStatistics.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+            if (tieCount > 1) {
+                tieCorrection += (Math.pow(tieCount, 3) - tieCount) / ((double) n * (n - 1));
+            }
+            i += tieCount;
+        }
+
+        double stddev = Math.sqrt((double) n1 * n2 / 12 * ((n + 1) - tieCorrection));
+        double z = Math.abs((uMin - meanU) / stddev);
+        return 0.5 * (1 + Erf.erf(z / Math.sqrt(2)));
```
asodja: ❔ Is it expected that identical scenarios it returns 0.5, not 0, and it can never go below 0.5? So according "confidence": 0.5 actually means "no difference detected"

I am not too familiar with this statistics, I guess it's ok, I just want to ask

src/main/java/org/gradle/profiler/report/JsonResultWriter.java
6hundreds: Good catch on assemble vs help — that comparison is indeed meaningless. But mirroring the HTML rule isn't the fix, for two reasons, plus some history we dug up:                                                                 
                                                                                                                                                                                                                                   
1. The HTML report doesn't actually avoid such comparisons. `multipleBuildToolsPresent` only controls the _default_ baseline (report.js:180); the checkbox in the results table lets you set any scenario as baseline (report-template.html:213), including cross-task pairs like assemble vs help. So the HTML relies on human judgment with a conservative default — there's no strict rule to be consistent with.                                   
2. `Multiple build tools` would drop confidence exactly for the comparisons this JSON is mostly meant for: same build, same tasks, different setup — isolated projects on/off, vintage vs configuration cache, --parallel on/off. Those share one build tool, so the JSON would report nothing for them while version comparisons would work. That asymmetry is hard to explain to consumers.                                                                      

Having said that, maybe it's better to decline baseline comparison in JSON. HTML must do it dynamically anyways, and robots can choose any meaningful baseline they want. WDYT?

src/main/java/org/gradle/profiler/report/SampleStatistics.java
6hundreds: Added a java doc, yes, that's fine

src/main/java/org/gradle/profiler/report/JsonResultWriter.java
asodja: That makes sense to me

6hundreds: @asodja Removed it, PTAL

src/main/js/org/gradle/profiler/report/report.js
6hundreds: Fixed

build.gradle.kts
6hundreds: Updated

README.md
asodja: 💅 I guess all "The result files" don't contain confidence now. Not very important, just a nitpick

asodja (approved):
LGTM

PR #842, opened 2026-08-21
Improve Isolated Projects compatibility of the build

alllex:
Removes the versioning violations tracked by #852. The 3 violations from the Nexus publish plugin stay, so the feature cannot be enabled yet.

### Why

Isolated Projects forbids build logic in one project to read the mutable state of another project. Versioning caused 28 of the build's 31 unique violations, and none of them were necessary: the version comes from a Gradle property and an optional build receipt that every project can read for itself, so the cross-project access only shared a string.

### What

- Versioning becomes a self-contained convention: each project resolves the version from the same two inputs, instead of reading it from an extension on the root project.
- The version is a provider, resolved only where a `String` is required.

### Verification

- Unique violations go from 31 to 3. The 3 that remain name the Nexus publish plugin.
- All 15 projects resolve the same version, and a staged build receipt overrides that version in all of them. This is the behavior the root-project extension supplied before.
- A receipt that carries no version fails the build, as it did before. This is the case a provider makes easy to lose, because an unreadable file and an absent one are the same absent value.
- A local publish of all 8 publishable projects gives the expected coordinates and metadata. Versioning feeds the published coordinates, so configuration alone does not prove this.
- The release task graph resolves, up to and including the staging close and release steps.
- Configuration performance is unverified. The feature stays off while the 3 violations remain, so this change has no measurement of the speed-up that motivates the feature.

### Details

**The build receipt controls the version of a promotion build, and it fails quietly if it stops working.** The publishing job stores the receipt as an artifact. The SDKMAN job reads that artifact and announces the version it finds. The receipt is what ties the second job to the version the first job published, because a property can move on between two separate builds. A receipt that stops being read costs nothing visible: the announcement still succeeds, under the wrong version. The receipt therefore stays the first branch of the resolution, and a receipt without a version is an error rather than a fallback to the property.

### DIFF
--- a/buildSrc/src/main/kotlin/versioning.kt
+++ b/buildSrc/src/main/kotlin/versioning.kt
@@ -1,0 +1,22 @@
+import org.gradle.api.Project
+import java.nio.file.Files
+import java.util.Properties
+
+/**
+ * Resolves the profiler version without observing the state of another project.
+ *
+ * Each project computes the version from the same two immutable inputs: the build
+ * receipt that CI drops into the root directory for a promotion build, and the
+ * `profiler.version` Gradle property. Sharing the value through an extension on
+ * the root project instead would break Isolated Projects.
+ */
+fun Project.resolveProfilerVersion(): String {
+    val rootDir = isolated.rootProject.projectDirectory.asFile
+    val incomingBuildReceipt = rootDir.resolve("incoming-distributions/$buildReceiptName")
+    if (incomingBuildReceipt.isFile) {
+        val properties = Properties()
+        Files.newInputStream(incomingBuildReceipt.toPath()).use { properties.load(it) }
+        return properties.getProperty("version")
+    }
+    return providers.gradleProperty("profiler.version").get()
+}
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/versioning.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+fun Project.resolveProfilerVersion(): String {
+    val rootDir = isolated.rootProject.projectDirectory.asFile
+    val incomingBuildReceipt = rootDir.resolve("incoming-distributions/$buildReceiptName")
+    if (incomingBuildReceipt.isFile) {
+        val properties = Properties()
+        Files.newInputStream(incomingBuildReceipt.toPath()).use { properties.load(it) }
+        return properties.getProperty("version")
+    }
+    return providers.gradleProperty("profiler.version").get()
```
6hundreds: ❓ Why call `get` so early tho? As I see it can be called only at execution of `createBuildReceipt` normally and at `taskGraph.whenReady`?

alllex: We use this version for `Project.version`, which is set in `profiler.versioning.gradle.kts`.

I made this more explicit by making the function return a `Provider` and calling get at the appropriate times.

6hundreds: I was under impression that `setVersion` accepts `Provider` (it accepts `Object`). It's not

6hundreds (approved):

PR #855, opened 2026-09-20
Pass jvm-args to the build JVM when running the CLI without a daemon

### LINKED ISSUE #853: jvm-args is silently ignored when daemon = none (build JVM never receives the declared heap)
## Summary

With `daemon = none`, a scenario's `jvm-args` never reach the JVM that runs the build. They are
applied to the Gradle **launcher** process via `GRADLE_OPTS`, while `org.gradle.jvmargs` is
simultaneously blanked on the command line — so the single-use daemon that actually executes the
build is launched without the declared heap. Benchmarks silently measure whatever heap the
environment happens to supply, with no warning that the declared value had no effect.

This contradicts the documented behaviour of the option:

> `jvm-args`: Sets or overrides the jvm arguments set by `org.gradle.jvmargs` in gradle.properties.

## Environment

- gradle-profiler **0.23.0**
- Gradle **9.4.1**
- JDK Amazon Corretto 21.0.11+10-LTS, Linux x86_64, 32 cores, 61.8 GiB RAM

## Root cause

`src/main/java/org/gradle/profiler/gradle/CliGradleClient.java`, in `runTasks()`:

```java
commandLine.add("-Dorg.gradle.daemon=" + daemon);
if (!daemon) {
    commandLine.add("-Dorg.gradle.jvmargs");   // valueless: sets the property to ""
}
...
if (daemon) {
    // wraps jvm-args into -Dorg.gradle.jvmargs="..." in GRADLE_OPTS  (works correctly)
} else {
    builder.environment().put("GRADLE_OPTS", daemonJvmArgs);  // raw -Xmx -> launcher only
}
```

Two things happen together in the `!daemon` path:

1. A **valueless** `-Dorg.gradle.jvmargs` is appended to the Gradle command line, discarding any
   value from `gradle.properties`.
2. The scenario's `jvm-args` are placed in `GRADLE_OPTS` as raw JVM flags, which size the
   **launcher/client** JVM rather than being forwarded to the build JVM.

The flag was introduced in d820309e9eeb (2017-04-06, "Fix --no-daemon to always use the client
VM"), whose commit message states the premise directly:

> When a project had a gradle.properties file with org.gradle.jvmargs, our no-daemon executor
> would fork a new VM (because the agent argument would make the client VM seem incompatible).
> We now explicitly pass an empty org.gradle.jvmargs, because the GRADLE_OPTS already contain
> everything from the gradle.properties. This ensures that we always run inside the client VM.

That premise no longer holds. On current Gradle, `--no-daemon` forks a single-use daemon
regardless, which every one of our runs reports:

```
To honour the JVM settings for this build a single-use Daemon process will be forked.
```

So the build does not run inside the client VM any more, and blanking the property now removes
the heap configuration from the process that does run it.

This is not a Gradle defect — Gradle correctly honours an explicitly-empty `org.gradle.jvmargs`.
The problem is emitting a valueless `-D`.

## Evidence

**1. The emitted command line** ends with the valueless flag:

```
Running command .../gradle-9.4.1/bin/gradle --gradle-user-home /.gradle \
  -Dorg.gradle.caching.push=false --no-configuration-cache -Dscan.tag.… \
  -I /tmp/gradleProfiler…gradle :build-logic:convention:fastCleanWithFlags \
  -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs
```

**2. The JVM that receives `jvm-args` is not the one running the build.** A scenario with

```
jvm-args = ["-Xmx40G", "-Xms40G", "-XX:+UseParallelGC", ...,
            "-Xlog:gc*,gc+heap=debug:file=profile-out/gc.log:time,uptime,level"]
```

produced a GC log whose owner exited after 102 seconds having done essentially nothing:

```
Heap Min Capacity: 40G   Heap Initial Capacity: 40G   Heap Max Capacity: 40G
...
[101.610s] PSYoungGen  total 12233216K, used 3355444K   (eden 32% used)
[101.610s] ParOldGen   total 27962368K, used 0K
[101.610s] Metaspace   used 10619K, committed 10816K
```

Zero garbage collections across the entire run, an empty old generation, and 10 MB of metaspace,
i.e. almost no classes loaded. A JVM configuring and building a large Android project loads
hundreds of MB. The declared 40 GiB went to a process that did no work.

**3. Changing the declared heap has no measurable effect.** Two runs identical in every respect
except the `jvm-args` heap — `-Xmx24G` vs `-Xmx31G` — behaved identically, both reaching the same
30-minute timeout. A 7 GiB difference in the declared heap produced no difference in behaviour.

**4. The same JVM args take effect under `run-using = cli` with a daemon**, which takes the other
branch and wraps them into `-Dorg.gradle.jvmargs="..."`.

## Impact

Any `daemon = none` scenario that declares `jvm-args` benchmarks a JVM configuration it did not
ask for. In our case the build JVM fell back to the ambient environment value, so a scenario
declaring 40 GiB ran the build in a far smaller heap, and Gradle eventually stopped the daemon
with:

```
Gradle build daemon has been stopped: since the JVM garbage collector is thrashing
```

That is a confusing symptom for a heap the user believes they configured generously. Because the
failure is indirect, it is easily misattributed to the machine, the collector, or Gradle itself —
it cost us a considerable amount of time before we read the profiler source.

## Reproduction

1. A project with `org.gradle.jvmargs=-Xmx2g` in `gradle.properties`.
2. A scenario with `daemon = none` and `jvm-args = ["-Xmx8g", "-XX:+PrintFlagsFinal"]`.
3. Inspect the build JVM's `MaxHeapSize`. Expected 8 GiB; actual is the ambient value (or JVM
   ergonomics), not 8 GiB.

Note that the forked daemon's flag output is not captured in gradle-profiler's own log, so
reading `MaxHeapSize` may require a build scan or an `-Xlog:...:file=` sink that the daemon
inherits.

## Suggested fix

Either:

- Forward the scenario's `jvm-args` as `-Dorg.gradle.jvmargs="..."` in the `!daemon` path, as the
  `daemon` path already does, instead of emitting a valueless `-D`; or
- Fail fast: when `jvm-args` is non-empty and `daemon = none`, error or warn loudly that the
  option has no effect, rather than silently discarding it.

At minimum the limitation should be documented on the `jvm-args` option, since the current
wording promises behaviour that the `daemon = none` path does not provide.

adityaanikam:
Fixes #853

With `daemon = none`, `CliGradleClient` put the scenario `jvm-args` into `GRADLE_OPTS` and added a valueless `-Dorg.gradle.jvmargs`. The comment history suggests this came from the time when a `--no-daemon` build ran inside the client JVM. On current Gradle that is no longer true: `--no-daemon` forks a single-use daemon, and only `org.gradle.jvmargs` reaches its JVM. `GRADLE_OPTS` sizes just the launcher. The valueless property then makes the build JVM fall back to default settings, so `jvm-args`, `system-properties` and profiler agent arguments were silently dropped.

I checked this on Gradle 9.7.1 with a small project that prints the build JVM's input arguments and max heap. With `daemon = none` and `jvm-args = ["-Xmx1500m"]` the build JVM ignored the setting. Passing the value as `-Dorg.gradle.jvmargs` in `GRADLE_OPTS` gave a build JVM with the requested heap.

The change removes the `daemon` and `!daemon` split in `runTasks`, so both modes build `GRADLE_OPTS` the same way: the client JVM arguments plus `-Dorg.gradle.jvmargs='...'` with the quoted scenario arguments. The full list is never empty because `GradleScenarioInvoker` already merges the build's own JVM arguments from the configuration reader, so project heap settings are kept. I also updated the Javadoc of `getClientJvmArguments()`, which said the client arguments were only for `--daemon`. `GRADLE_OPTS` is now logged in both modes.

I considered a smaller change that only adds the property while keeping the old `GRADLE_OPTS` handling for the launcher. I did not choose it because profiler agents (JFR, async-profiler) would then attach to both the launcher and the build JVM, and could write to the same output file.

Testing: I added `override jvm args when not using a daemon`, a `daemon = none` twin of the existing `override jvm args` test. It asserts on the arguments of the JVM that runs the build. It uses `latestSupportedGradleVersion` on purpose. On 7.6.4 it passes even without the fix, and I did not investigate why. On 9.3.1 (and 9.7.1) it fails without the fix on the `-Xmx2G` assertion and passes with it. To run it locally I had to point the test at a pre-downloaded Gradle distribution, because my connection cannot finish the Tooling API download within its 10 second limit. I did not run the rest of the test suite locally, so I am relying on CI for that.

Since #853 was still marked `to-triage`, I am happy to change the approach if you would rather warn or fail fast when `jvm-args` is used with `daemon = none`.


ov7a:
Thank you for your interest in Gradle.

This change would require a proper design and a discussion with the team.

See the original issue, the team will discuss it:
* https://github.com/gradle/gradle-profiler/issues/853

PR #856, opened 2026-09-21
Enable IP in the build

### LINKED ISSUE #852: The build is not compatible with Isolated Projects
Isolated Projects cannot be enabled on this build: diagnostics mode reports 31 unique violations.

Versioning causes 28. Every subproject applies a plugin to the root project, then reads the version back out of an extension there, although the version comes from a Gradle property and an optional build receipt that every project can read for itself. #842 is open and removes these.

The Nexus publish plugin causes the other 3, configuring every subproject through `allprojects`. Third-party, so not fixable here: release 2.0.0 is from April 2024, and [upstream #343](https://github.com/gradle-nexus/publish-plugin/issues/343) has been open since June 2024, with a November 2025 request for a plan unanswered. Clearing them means moving to a publishing plugin that does not aggregate through `allprojects`. [`com.gradleup.nmcp`](https://github.com/GradleUp/nmcp) is a candidate, not yet evaluated, and it would also retire the transitional OSSRH staging shim the build points at.

A conditional application of the Nexus plugin was tried in #842 and reverted: it dropped the publishing tasks from the isolated code path, so a publish under the feature reported a missing task rather than a violation, leaving the build looking configured while publishing was gone.

ov7a:
There are no known violations now, and I haven't found any.


alllex (approved):

6hundreds (approved):
💯

